PR #2, opened 2023-06-06
fix main readme images background

yrtimiD:
Added white rectangle behind the image to ensure always readable text + some visual padding around. Additionally, this ensures that after future editing of the diagrams on the draw.io bg customizations are not lost.


layers.drawio.svg
stevenhartley, line 1: Some of the icons have disappeared when I look at the diff (AMQP, DDS, CloudEvent, etc))

overview.drawio.svg
stevenhartley, line 1: Same comment as previous, cloudevent icon is missing.

ues.drawio.svg
stevenhartley, line 1: Font has changed, I would prefer to keep the old font

stevenhartley (changes requested):
Please see comments.

overview.drawio.svg
yrtimiD: Fixed this. These two images were not originally embedded but fetched from the remote location making them disappear randomly. Now they are embedded and should not make surprises.

ues.drawio.svg
yrtimiD: thanks, embedded the font, now should work always
(and changed font style for the "Legend" label to be as all other, hope this is ok?)

layers.drawio.svg
yrtimiD: Fixed this. These two images were not originally embedded but fetched from the remote location making them disappear randomly. Now they are embedded and should not make surprises.

overview.drawio.svg
yrtimiD: fixed

stevenhartley (approved):
Thank you for the change!

PR #3, opened 2023-06-13
fix protobuf mime type

yrtimiD:
application/protobuf is not official yet, instead application/x-protobuf should be used


stevenhartley (commented):
Thanks for the fixes @yrtimiD !

Issue #1, opened 2023-05-16
Initial commit


stevenhartley:
Initial commit of the specifications

stevenhartley:
code committed

PR #16, opened 2023-08-31
uProtocol 1.5

stevenhartley:
Various changes and improvements to introduce uProtocol 1.5. 

#1


basics/uri.adoc
```diff
@@ -98,11 +158,96 @@ a| Method or topic Id from the protobuf file
...
-The wildcard character `*` is used in certain circumstances in lieu of an item in the URI for matching any character or sequence of characters in a search.
+wildcard are used to replace portions of the URI to perform pattern matching (i.e. subscribing to multiple topics, searching for services, etc...). 
+ 
+ * *SHOULD* support wildcards in the authority, entity, resource, and message portions of the URI
+ * Wildcards *MUST* be `*` (asterisk) for long and short form URIs
+ * Wildcard *MUST* be the max value for Micro URI (i.e. 0xFF for 8 bit item)
+
+NOTE: Not all uPlatforms and uEs support wildcards, please refer to the uPlatform and uE documentation for more information.
 
 == Example URIs
```
halimragab, line 252: Is it possible to extend the example URIs to show the short Uri and Micro Uri?

up-l1/README.adoc
```diff
@@ -279,6 +239,23 @@ Access token as defined in per link:../up-l2/permissions.adoc#_token_based_acces
 * *MAY* be of type `String` for Java, or `std::vector<uint8_t>` for C++
 
 
+=== UPayload
+
+UPayload is a container for the uP-L3 application layer data to be transmitted between uEs. The UPayload structure contains only the data, its size, and an optional hint to the serialization format. 
+
+If the data is passed by reference (ex. pointer):
+
+* `data` *MUST* contain the pointer to the payload and `size` contains the actual data size
```
sfactor33, line 248: need to add another serialization hint type of pointer to make this explicit?  streamer likely needs to know this for data heading out of the device so it can dereference the point to send the actual payload outside the device.  Has to be able to do this without knowledge of the topic.

basics/uri.adoc
stevenhartley: yes done (in next merge)

basics/uri.adoc
```diff
@@ -27,12 +28,70 @@ The following is taken from https://datatracker.ietf.org/doc/html/rfc3986#append
...
+ * *MUST* support detection of local vs remote URIs from `UAuthority`
+
+if the authority is an IP address:
+
+ * IP address *MUST* be stored in `device` member variable 
+
+=== UEntity
+`UEntity` type contains the uE id, name, and version portion of the uProtocol URI.
+ * Type name *MUST* be `UEntity`
+ * *MUST* contain `name` and `version` member variables of type `std::string`, with accessor methods to fetch each
```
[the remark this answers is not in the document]
stevenhartley: Above the spec states that the SDK author can replace variable data types like std::string with fixed likes like char[] etc...,

basics/uri.adoc
```diff
@@ -27,12 +28,70 @@ The following is taken from https://datatracker.ietf.org/doc/html/rfc3986#append
...
+
+`UUri` type is used to represent the URI. It is composed of `UAuthority`, `UEntity`, and `UResource` types.
+
+ * Type name *MUST* be `UUri`
+ * *MUST* have-a <<UAuthority>>, <<UEntity>>, and <<UResource>> types 
+ * *MUST* have methods to build `UUri` from either a string or any combination of <<UAuthority>>, <<UEntity>>, and <<UResource>> types based on the URI specifications
+ * *MUST* have a method to build the an RPC URI passing `<<UAuthority>> and <<UEntity>> where <<UResource>> is set to `"rpc.response"`.
+ * *MUST* have a method to convert to and from string
+ * *MUST* support URI with and without scheme defined (in the string)
+
```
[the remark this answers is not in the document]
stevenhartley: Yes now that I'm almost done coding all the factory methods, I'll put in notes that the SDKs must do the same.

basics/uri.adoc
```diff
@@ -27,12 +28,70 @@ The following is taken from https://datatracker.ietf.org/doc/html/rfc3986#append
...
+  - `name` and `instance` only
+  - `name` only
+
+=== id
+the UUri ID is used by some transports to map the UUri (long/short/micro format (described below) to a transport specific ID. the parameter allows for optimizations at the transport layer that use IDs in lieu of uProtocols long/short/micro uris. 
+
+ * Type name *MUST* be `id`
+ * *MUST* fit within a 64 bit (long) with accessor method to fetch it
+ * *MAY* be a 32 bit integer
+
```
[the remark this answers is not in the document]
stevenhartley: You mean the name of the accessors? hum that is a good point, right now the accessor has the name of the type so id(), will state that in the spec (good point to enable consistency)

basics/uri.adoc
```diff
@@ -140,3 +285,48 @@ NOTE: Local URIs do not contain the authority (and begin with `/`)
 
 
 NOTE: Please also refer to https://devhints.io/semver[devhints.io/semver] for more examples of version strings
+
+
+== Best Practices
+
+A UUri contain UAuthority, UEntity, and UResource. <<Uri Formats>> defined earlier meanrly define how UAuthority, UEntity, and UResource are expressed (long string names, short-form string, or encoded byte array). Some formats are human readable, wile others are not.
```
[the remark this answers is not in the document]
stevenhartley: fixed thanks

basics/uri.adoc
```diff
@@ -140,3 +285,48 @@ NOTE: Local URIs do not contain the authority (and begin with `/`)
...
+
+uProtocol URI specifications complies with of https://datatracker.ietf.org/doc/html/rfc3986[RFC3986] that builds https://datatracker.ietf.org/doc/html/rfc1034[DNS concepts and facilities] to allow representation of UAuthorty as names or  IP addresses.
```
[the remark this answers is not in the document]
stevenhartley: fixed

basics/uri.adoc
```diff
@@ -140,3 +285,48 @@ NOTE: Local URIs do not contain the authority (and begin with `/`)
...
+
+A UUri contain UAuthority, UEntity, and UResource. <<Uri Formats>> defined earlier meanrly define how UAuthority, UEntity, and UResource are expressed (long string names, short-form string, or encoded byte array). Some formats are human readable, wile others are not.
+
+uProtocol URI specifications complies with of https://datatracker.ietf.org/doc/html/rfc3986[RFC3986] that builds https://datatracker.ietf.org/doc/html/rfc1034[DNS concepts and facilities] to allow representation of UAuthorty as names or  IP addresses. 
+
+UEntity name to id, and UResource name+instance+message to id, are declared in the protos and compiled into language specific code using  https://protobuf.dev/programming-guides/proto2/#options[Options]. Application layer (uP-L3) uEs (client and servers) then have available the all the metadata needed to populate UUri with both the ids and names for UEntity and UResource as well the application layer code can also convert to/from the name-based (long Uri Format) to ID based (Short or Micro Uri Formats) locally (no dynamic lookup required). uEs are then able to send to the link:../up-l1/README.adoc[uP-L1 uTransport layer] The fully populated UUri objects.
```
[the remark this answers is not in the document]
stevenhartley: They are part of different data models (UAttributes ID and UEntity ID)...

basics/uri.adoc
```diff
@@ -140,3 +285,48 @@ NOTE: Local URIs do not contain the authority (and begin with `/`)
...
+* Local communication (within the same device) *SHOULD* use local URIs (empty authority)
+* Remote communication (between devices) *MUST* contain at least the device name
+* Intra-domain name/address resolution *MUST* be locally resolvable (i.e. does not require global recursive dns resolution)
+
+
+=== UEntity & UResource
+
+==== uTransports (uP-L1)
+
+* *MUST* declare if they shall use long, short, or micro Uri format such that passing a UUri with missing information to construct the format will result in an `INVALID_PARAMETER` returned error
```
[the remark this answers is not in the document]
stevenhartley: Transport technology needs to state what format they require so if they will use short/micro then IDs MUST be populated by the uEs but you bring up a good point, how do I enforce this, let me take a crack at it...

up-l1/README.adoc
stevenhartley: Lets talk about this with the others, it is a valid point

up-l1/README.adoc
```diff
@@ -476,17 +438,17 @@ API Signature:
 
 === Receive()
```
eranrs, line 439: How should it be implemented or used?
Is the receive waiting for the next message? for how long? 
Is it the last received message? who will maintain the last message? a listener? 
I would suggest that the receive will be the blocking solution that will have a timeout.

basics/uri.adoc
```diff
@@ -27,12 +28,70 @@ The following is taken from https://datatracker.ietf.org/doc/html/rfc3986#append
...
+`UAuthority` type is responsible for storing the uDevice and uDomain portion of the uProtocol URI.
+
+ * Type name *MUST* be `UAuthority`
+ * *MUST* contain `domain` and `device` member variables of type `std::string` with accessor methods of the same name to fetch each
+ * *MUST* be able to construct a `UAuthority` from a string
+ * *MUST* support detection of local vs remote URIs from `UAuthority`
+
+if the authority is an IP address:
+
+ * IP address *MUST* be stored in `device` member variable
```
tamarafischer, line 63: Is it important to know that the device has an IP address, as apposed to some other sting defining a device?
If so, I would suggest a static factory method called fromIP.
UAuthority.fromIpAddress(ip address)

This makes it explicit. In addition, we can now add a hint to the UAuthority that can help resolvers, resolve the UAuthority from some short value in the _Fully Qualified URI_

basics/uri.adoc
```diff
@@ -140,3 +323,50 @@ NOTE: Local URIs do not contain the authority (and begin with `/`)
...
+
+A UUri contain UAuthority, UEntity, and UResource. <<Uri Formats>> defined earlier merely define how UAuthority, UEntity, and UResource are expressed (long string names, short-form string, or encoded byte array). Some formats are human readable, while others are not.
+
+uProtocol URI specifications complies with of https://datatracker.ietf.org/doc/html/rfc3986[RFC3986] that builds https://datatracker.ietf.org/doc/html/rfc1034[DNS concepts and facilities] to allow representation of UAuthorty as names or IP addresses. 
+
+UEntity name to id, and UResource name+instance+message to id, are declared in the protos and compiled into language specific code using  https://protobuf.dev/programming-guides/proto2/#options[Options]. Application layer (uP-L3) uEs (client and servers) then have available the all the metadata needed to populate UUri with both the ids and names for UEntity and UResource as well the application layer code can also convert to/from the name-based (long Uri Format) to ID based (Short or Micro Uri Formats) locally (no dynamic lookup required). uEs are then able to send to the link:../up-l1/README.adoc[uP-L1 uTransport layer] The fully populated UUri objects.
```
tamarafischer, line 334: I think this is a typo "then have available the all the metadata needed"
I think there is a "the" too many.

basics/uri.adoc
```diff
@@ -140,3 +323,50 @@ NOTE: Local URIs do not contain the authority (and begin with `/`)
...
+
+link:../up-l2/README.adoc[Dispatchers] however do not have the id mapping metadata as they are not aware of the service definitions in the protos. As a result, dispatchers can only work with the URI information provided (names or ids). If a dispatcher is required to convert the Uri format, it must dynamically lookup (resolve through discovery) the name/ids.
```
tamarafischer, line 336: A backoffice application that has many tenants, ie users, drivers of cars, will be subscribing to the same topic, say temperature, but from different cars according to what the driver specifies.
The driver specifies a VIN - this is the car to monitor the temperature.
There are many drivers.
Events come in to the backoffice application -> if the UAuthority does not contain the vin or there is no way to understand what device the event came from, correlating to the subscription request - the backoffice application has a big problem and can't support many drivers/cars.
In addition, to not leak data between drivers, it is very important that this correlation is a pure function!

up-l1/README.adoc
tamarafischer: Why? Can you please elaborate?

eranrs: I don't understand the when would you want to use this method. is it for getting the next message? is it to pick a message from a queue? or picking the last message (local uTwin)?
That is why I am looking at the performance issues of the options. The "local uTwin" or the "queue" option will need to queue messages using the callback function that would need to run in the background. how many messages do we queue? for how long?

basics/uri.adoc
sfactor33: What would the uE do with the IP address?  The SDK needs to know it.  Can't we hide that behind the helper functions?  I can see this will need to be different on the MCU side of things where IP addresses might be hardcoded, but that is different.

basics/uri.adoc
sfactor33: If VIN doesn't reach the BO directly, then we must make sure the BO can resolve that UUri form to the long form to add the VIN to Tamara's point.  That conversion may not be trivial.  
The vehicle would be the easiest place to do that conversion, at the cost of a longer message to the BO.

basics/uri.adoc
stevenhartley: Removing this requirement, IP address is stored in InetAddress not to confuse with device and domain names

basics/uri.adoc
stevenhartley: Today we use local addresses within a device and then append authority in the streamer, I would expect we continue to do this moving forward

basics/uri.adoc
```diff
@@ -140,3 +323,50 @@ NOTE: Local URIs do not contain the authority (and begin with `/`)
...
+
+UAuthority attributes are resolved through DNS lookups either to a local cache and/or to recursive DNS servers. Not all devices are globally addressable (through a recursive DNS servers). (ex. mobile devices behind a firewall). 
+
+IMPORTANT: If a uE has any specific Uri format requirements, they *MUST* be documented in their respective APIs. Please see link:../up-l3/README.adoc#_core_platform_ues[Core Platform uEs] for more specific requirements for core services.
+
+The next sections shall provide best practices for populating and using URIs in uProtocol.
+
+=== UAuthority
+* Local communication (within the same device) *SHOULD* use local URIs (empty authority)
+* Remote communication (between devices) *MUST* contain at least the device name
```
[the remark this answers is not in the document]
stevenhartley: VIN is domain

sdk.adoc
```diff
@@ -0,0 +1,111 @@
...
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+----
+
+== Overview
+The purpose of uProtocol language specific SDKs are to provide the consistent mental model across all programming languages (Java, C++, Rust, Python, etc...) for core uP-L1 & L2 specifications, interfaces, and data models. The SDK is core to all uProtocol software as it is used to implements uEs (apps & services), transports, and platforms
```
eranrs, line 27: SDK is not only libraries.
You can state these are SDK libraries.

sdk.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+| `cloudevents`
+| link:up-l1/cloudevents.adoc[CloudEvents]
+| Specification of how uProtocol data models can be expressed using CloudEvents data model
+
+| `transport`
+| link:up-l1/README.adoc[uTransport]
+| uP-L1 uTransport interface & data model definitions used to build language specific transports implementations (i.e. zenoh, SOME/IP, MQTT, etc..) that run on a given uPlatform.
+
+| `rpcclient`
```
eranrs, line 86: Aren't these uTransport, rpcClient and RPCServer and interfaces which need to be implemented in the uLink?
The SDK libraries are at the end, the specific implementation.

eranrs (changes requested):
Need to clarify what is an SDK.
An SDK includes libraries, test suit and documentation.

sdk.adoc
stevenhartley: Rephrased and removed the term SDK like we discussed

sdk.adoc
stevenhartley: Rephased and addressed

eranrs (approved):

Issue #24, opened 2023-11-06
uP-L3 & uP-L2: Throttling publication rate in the streamer


stevenhartley:
Often there is the case where a topic is published at a high frequency within a device but we don't require (in fact prefer) to receive the published event less frequently between devices. Example situation requiring this throttling is say a service publishing current GPS location every 500ms however sending at that frequency would be too costly to the cloud so we would like to sent every 3 seconds. 

The CR shall be to cover adding to the SubscribeAttribute, an optional attribute that will be interpreted by the streamers for throttling published events.

stevenhartley:
We will handle this topic in up-core-api

PR #30, opened 2023-11-29
Introduce Publication Throttling

stevenhartley:
The following adds to uStreamer and uSubscription the ability to throttle published events when we send them between devices. Example use is cloud application subscribing to high frequency topics like GPS location.

#24

### DIFF
--- a/up-l2/dispatchers/README.adoc
+++ b/up-l2/dispatchers/README.adoc
@@ -30,13 +30,13 @@ SPDX-License-Identifier: Apache-2.0
 ----
 
 
-Like IP packets, uProtocol messages (CE) have a source attribute (originator of the message) and sink attribute (where should the message be sent to). These attributes are used to route the CEs from one uE to the next if the destination is not the receiving uE.
+Like IP packets, uProtocol messages (UMessages) have a source attribute (originator of the message) and sink attribute (where should the message be sent to). These attributes are used to route the UMessages from one uE to the next if the destination is not the receiving uE. For more information about these and other uProtocol attributes, please refer to link:../../basics/uattributes.adoc[UAttributes].
 
 NOTE: Dispatcher and router shall be used For the remainder of this document we will use the term dispatcher and router interchangeably.
 
-The header contains information for routing as well as metadata (of the data). One of the core principles of uProtocol is that the data portion of CE *MUST* be untouched by message routers, this is very similar to how most Internet standards work today. Only the source who generated the CE and the sink who will consume the CE needs to understand/analyze the payload of the CE.
+The header contains information for routing as well as metadata (of the data). One of the core principles of uProtocol is that the data portion of UMessage *MUST* be untouched by dispatchers, this is very similar to how most Internet standards work today. Only the source who generated the UMessages and the sink who will consume the UMessages needs to understand/analyze the payload of the UMessage (UPayload).
 
-To be able to forward/dispatch/route CEs through the network, we must define specific purpose built uEs to perform these tasks (ex. Ethernet switches, IP routers, etc...). Platform uEs that are responsible for event dispatching and implementing the communication layer are described in the sections below. We will elaborate on these specific uEs in the Platform uEs section below.
+To be able to forward/dispatch/route UMessages through the network, we must define specific purpose built uEs to perform these tasks (ex. Ethernet switches, IP routers, etc...). Platform uEs that are responsible for event dispatching and implementing the communication layer are described in the sections below. We will elaborate on these specific uEs in the Platform uEs section below.
 
 .Dispatchers
 [width=100%",cols="30%,70%"]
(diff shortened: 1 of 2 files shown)


up-l2/dispatchers/README.adoc
```diff
@@ -30,95 +30,88 @@ SPDX-License-Identifier: Apache-2.0
...
-|The dispatcher ran out of resources (buffer full)
+=== uStreamer Throttling of Published Messages
 
-|`*INVALID_ARGUMENT*`
-|Invalid CE header attributes not covered above (ex. any mal-formatted attributes)
+Throttling is the act of reducing the publication frequency of messages published between devices. A subscriber might want to throttle publishations to reduce network bandwidth costs (e.g. cloud app subscribing to vehicle topics like GPS location). 
 
-|`*UNKNOWN*`
-|An unknown (but not critical) error has occurred
+A subscriber uses  `max_frequency` attribute of https://github.com/eclipse-uprotocol/uprotocol-core-api/blob/main/src/main/proto/core/usubscription/v3/usubscription.proto#L121[SubscribeAttributes] message to set the desired maximum publication frequency (value in Hertz), this attribute is only considered for remote topics.
```
grandcb01, line 105: Capturing some comments made by Doug

I’d prefer not to specify the throttling in frequency.  Period or a max message rate in seconds/milliseconds would be preferred as this will match our code.  Frequency tends to imply a trend.

up-l2/dispatchers/README.adoc
```diff
@@ -30,95 +30,88 @@ SPDX-License-Identifier: Apache-2.0
...
-.RPC Error Flow
-[#rpc-error-flow]
-image::rpc_flow.png[RPC Error Handling]
\ No newline at end of file
+If messages are received at a higher frequency than `max_frequency`:
+
+* *MUST* be dropped by the uStreamer
+* Dropped messages *MUST NOT* be added to the DLT
+
+if `max_frequency` is not set or the frequency is less than the publisher's frequency:
```
cbeck47, line 115: Agree on if max_frequency is not set.  

Should it state, if max_frequency is higher than the publisher's frequency than must not throttle messages?

Assuming that is correct, my concern is that we are adding complexity into streamer logic when it is not needed in this scenario.

up-l2/dispatchers/README.adoc
halimragab: agree with the above comment. I think units of time is more clear.

> maximum publication frequency

also, I feel using the words "maximum publication" here may lead to confusion. As this is really just a filtering, and the publisher likely won't change its publication frequency

sfactor33:
I agree with the change proposal to use max_frequency as the usub attribute

tamarafischer:
The job of uSubscription is to answer the question "who is subscribed"
This is used by the dispatchers/streamers to deliver messages
Dispatchers have to handle many many events and need to be as fast as possible - minimum processing
Given an event, a dispatcher needs to figure out where it goes and send it on its way.

By definition, a uE publishes state changes.
It is the business logic of the uE to decide on the frequency of its published data.
As an optimization it can decide not to publish if no one is subscribed - but the definition is that if you have a state change, you publish it (this enables uTwin to always have the last state)

I believe we should be using backpressure - it is a pull mechanism and is decided by **clients**.

Backpressure is a mechanism of the client to tell the publisher that it is overwhelmed and to stop sending more events - in the case of distributed systems, most people put a broker such as Kafka(EventHubs, Kinesis, whatever) so that the publisher can send data and consumers can read it at their own pace, hence implementing backpressure in the reactive streams.

IMHO adding throtteling on the side of the server will not give the clients wiggle room to get better and read more.

If we have a use case that we want to control the data coming out of the car, it is the job of some other component - yes, probably the streamer that sends data out to the cloud.
That streamer will probably implement a priority queue to send out the most important information first and other logic.
IMHO, this is not something the uE should deal with. A uE publishes state changes and another uE consumes them - they are decoupled in space and time - cars and cloud are decoupled in space and time.

up-l2/dispatchers/README.adoc
```diff
@@ -30,95 +30,88 @@ SPDX-License-Identifier: Apache-2.0
...
-  ** *MUST* queue CEs not successfully acknowledged (transport level at-least-once delivery confirmation described above)
-  ** *MUST* attempt to retry transmission of the CE. Retry policy is specific to the dispatcher implementation
-  ** Dispatcher *MUST NOT* discard CEs unless either CE has expired (CE.ttl), or the egress queue is full. CEs that cannot be delivered are sent to a Dead Letter Office Topic
+* *MUST* support At-least-once delivery policy, this means that the dispatcher will make every attempt to dispatch the message to the intended receiver
+  ** *MUST* queue messages not successfully acknowledged (transport level at-least-once delivery confirmation described above)
+  ** *MUST* attempt to retry transmission of the message. Retry policy is specific to the dispatcher implementation
+  ** Dispatcher *MUST NOT* discard message unless either the message has expired (see https://github.com/eclipse-uprotocol/uprotocol-core-api/blob/main/src/main/proto/uattributes.proto[`UAttribute:ttl`] for more details), or the egress queue is full. Messages that cannot be delivered are sent to a Dead Letter Office Topic (DLT)
```
tamarafischer, line 71: Whenever we define a queue to buffer messages, we define a strategy for when the queue is full.
 Specifying the Dead Letter Queue is only one option and it can become full as well or blow up the system by hogging all the memory. 
There are many strategies that can be used and I think we should specify something here as well.
Priority Queue, Drop Oldest, Drop Newest ....

up-l2/dispatchers/README.adoc
```diff
@@ -30,95 +30,88 @@ SPDX-License-Identifier: Apache-2.0
...
-  ** DLT *MUST* include the reason for the failed delivery attempt using  error codes defined in google.rpc.Code
-  ** uEs MUST be able to subscribe to the DLT to be notified of message deliver failures
+  ** DLT *MUST* include at least the message header, *SHOULD* contain the full Message
+  ** DLT *MUST* include the reason for the failed delivery attempt using per https://github.com/eclipse-uprotocol/uprotocol-core-api/blob/main/src/main/proto/ustatus.proto[UStatus] error codes
+  ** uEs *SHOULD* be able to subscribe to the DLT to be notified of message deliver failures
 
 If the uP-L1 delivery method is push:
 
-* *SHALL* provide an API to start/stop dispatching of CEs per-topic, this is to avoid having to queue CEs on the Receiver if the Receiver is not ready to receive the CEs
+* *SHOULD* provide an API to start/stop dispatching of messages per-topic, this is to avoid having to queue messages on the Receiver if the Receiver is not ready to receive
```
tamarafischer, line 85: This is coupling the sender and receiver.
In a distributed system, uEs are separated by space and time - they are not directly connected.
This is the whole idea. Publishers send and consumers consumer.
A publisher might have more than one consumer - one is happily consuming while another is shut down and will come back up later and continue consuming from where it left off.
Does this mean that the publisher will stop because one of the consumers is having a difficult day?

up-l2/dispatchers/README.adoc
```diff
@@ -30,95 +30,88 @@ SPDX-License-Identifier: Apache-2.0
...
 
 
 == RPC Error Handling
 
-When a dispatcher is unable to dispatch an event for a given reason (queue full, etc...), the dispatcher is responsible to generate an RPC Response message and send it to the originator of the request. The status code is populated in the  link:../messages/v1/README.adoc#_response_message[`commstatus`] attribute of the response message using the <<commstatus-codes>> defined below.
+When a dispatcher is unable to dispatch an event for a given reason (e.g. queue full, etc...), the dispatcher is responsible to generate an RPC Response message and send it to the originator of the request. The status code is populated in the  link:../messages/v1/README.adoc#_response_message[`commstatus`] attribute of the response message using the error codes defined in https://github.com/eclipse-uprotocol/uprotocol-core-api/blob/main/src/main/proto/ustatus.proto[UCode].
```
tamarafischer, line 90: so, if the dispatcher could not deliver a message it sends another RPC that can't be delivered for the same reason the first event could not be delivered hence adding more and more messages that can't be delivered, bogging down the system.
IMHO dispatchers so their best to send the messages, using brokers, using caches using whatever. Either the subscribed uE gets its data or it does not. Why do we need another message saying that things failed? How does that help?
I think I am not understanding this correctly. Can we please clarify?

up-l2/dispatchers/README.adoc
```diff
@@ -30,95 +30,88 @@ SPDX-License-Identifier: Apache-2.0
...
-|`*UNAUTHENTICATED*`
-|source does not have valid authentication credentials (ex. uE's identity does not match the source attribute)
 
-|`*RESOURCE_EXHAUSTED*`
-|The dispatcher ran out of resources (buffer full)
+=== uStreamer Throttling of Published Messages
 
-|`*INVALID_ARGUMENT*`
-|Invalid CE header attributes not covered above (ex. any mal-formatted attributes)
+Throttling is the act of reducing the publication frequency of messages published between devices. A subscriber might want to throttle publishations to reduce network bandwidth costs (e.g. cloud app subscribing to vehicle topics like GPS location).
```
tamarafischer, line 103: A subscriber should simply pull that data as fast as he can handle it and maybe optimize or scale to pull more in the future.
This is called backpressure on the reactive stream.

It is the responsibility of the uE and not the uPlatform/dispatcher/uSubscription

up-l3/usubscription/v2/README.adoc
```diff
@@ -218,7 +218,9 @@ image::unsub_local.svg[Unsubscribe Local Flow]
 .Remote Unsubscribe Flow
 image::unsub_remote.svg[Unsubscribe Remote Flow]
 
-* uSubscription *MUST* change the subscriber to itself (core.usubscription) when unsubscribing to remote topics 
+* uSubscription *MUST* change the subscriber to itself (core.usubscription) when unsubscribing to remote topics
```
tamarafischer, line 221: this is expensive copying, an it is an anti-pattern to change the payload.
The system should know - or another communication message needs to be created, meaning it is another message with a **different ID**.
I believe messages should be immutable.
If this is the case, we might have to add a tracking attribute since the message id is not longer something valid we can use to track messages through the system.

Issue #21, opened 2023-11-06
uP-L1: Zenoh Transport Specific Requirements


stevenhartley:
The following will capture the requirements for mapping UUri, UAttributes, and UPayload to zenoh defined header and payload for consistencies across implementations & languages of the zenoh transport.

PR #52, opened 2024-01-17
Add Zenoh spec to uProtocol

evshary:
The PR is not ready to be merged, but I think it would be easier to discuss the spec if we open a draft PR.
I'll keep working on `uprotocol-rust-ulink-zenoh` and updating the spec here.
Feel free to add any comments on this.


stevenhartley:
PR to address #21

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,93 @@ Zero Overhead Network Protocol.
...
+| CS5 | INTERACTIVE_HIGH
+| CS6 | REAL_TIME
+|===
+
+Note that the priority in Zenoh only works for publisher.
+
+=== URI Mapping
+
+UUri should be able to be serialized it into micro form.
+The mapping from micro-from UUri to Zenoh key is `Vec<u8>` to lower hex string.
```
PLeVasseur, line 98: Would be good IMHO to give an example of what is meant by this, e.g.
`[22, 33, 128]` => `what here` to help make concrete for the reader

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,93 @@ Zero Overhead Network Protocol.
...
+If UAuthority is defined in UUri, it should be put in the following part.
+Then the rest of the UUri can be put in the last part.
+
+* UUri without UAthourity: `up/[byte representation of the UUri]`
+* UUri with UAthourity: `up/[byte representation of the UAuthority]/[byte representation of the rest of the UUri]`
+
+Note that if the UUri only contains UAthority with the empty UEntity and UResource,
+it's defined as a special form of Zenoh key `up/[byte representation of UAuthority]/**`.
+This is only used by uStreamer.
+
```
PLeVasseur, line 110: I think it'd be good to add a section to the bottom here where we explain
1. what a UUri without UAuthority and with UAuthority look like (written out in psuedocode for .proto maybe, since there will be different implementation languages)
2. what the byte array representation would look like i.e. `[22, 33, 128]`
3. and then what the final Zenoh KeyExpr string looks like, i.e. `up/actual-byte-strings/more-actual-byte-strings`.

For me at least, I think it'll help clarify. WDYT?

up-l1/zenoh.adoc
evshary: Done.

up-l1/zenoh.adoc
evshary: Done.

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+Then we can transform the micro-form UUri, which should be `u8` array, into string in lowercase hexadecimal representation.
+
+Zenoh key for uProtocol consists of three parts and they are separated by `/`.
+Zenoh key **MUST** start with `up/`.
+If UAuthority is defined in UUri, it should be put in the second part.
+Then the rest of the UUri can be put in the last part.
+
+To be more specific:
+
+* UUri without UAthourity: `up/[byte representation of the UUri]`
```
PLeVasseur, line 105: `UAthourity` => `UAuthority`

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+* UUri with UAthourity: `up/[byte representation of the UAuthority]/[byte representation of the rest of the UUri]`
```
PLeVasseur, line 106: `UAthourity` => `UAuthority`

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+
+Note that there is a special case. If the UUri only contains UAthority with empty UEntity and UResource,
```
PLeVasseur, line 108: `UAthority` => `UAuthority`

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+it's defined as a special form of Zenoh key `up/[byte representation of UAuthority]/**`.
+This is only used by uStreamer.
+
+Some examples:
+
+* UUri without UAthourity:
```
PLeVasseur, line 114: `UAthourity` => `UAuthority`

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+  }
+}
+----
+
+There is no UAuthority, so we can ignore the second part.
+Only ID in UEntity and UResource matters while serializing into link:../basics/uri.adoc#421-local-micro-uri[micro-form Uri].
+Then transform it into lowercase hexadecimal representation.
+The Zenoh key should be `up/0100162e04d20100`.
+
+* UUri with UAthourity
```
PLeVasseur, line 139: `UAthourity` => `UAuthority`

PLeVasseur:
Thanks @evshary, I think it's just about there. :slightly_smiling_face:  I think adding a link to the appropriate part of the `up-spec` detailing how Micro Form URIs work and then continuing to describe that particular example would be a good way to tie things together.

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+  entity: empty,
+  resource: empty,
+}
+----
+
+The UAuthority should be serialized into mirco form.
+It should be length + serialized UUri in lowercase hexadecimal representation, which is `06` + `0102030a0b0c`.
+There is no UEntity and UResource, so it is the special form UUri.
+The third part should be `+++**+++`.
+Therefore, the Zenoh key is `up/060102030a0b0c/+++**+++`.
```
PLeVasseur, line 157: Just making a link to refer to elsewhere

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+
+=== UPayload
+
+Although UPayload can be encoded into protobuf, it can save a lot of time without encoding.
+UPayload consists of three parts: length, format, data.
+
+* Length should only work with shared memory, and we ignore it temporarily.
+* Format can be sent with Zenoh encoding. Zenoh supports `Encoding::WithSuffix`. The type here should be `KnownEncoding::AppCustom` and the int32 format is transformed into the string.
```
stevenhartley, line 165: @evshary for AppCustom is the type string and that is why you do to_string() of the UPayloadFormat value (0, 1, 2, etc...)?

stevenhartley (changes requested):
Please see comments

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
-for more information, please visit https://projects.eclipse.org/projects/iot.zenoh or https://zenoh.io/
+For more information, please visit https://projects.eclipse.org/projects/iot.zenoh or https://zenoh.io/
 
 == Specifications
 
-_Coming Soon_
\ No newline at end of file
+=== Zenoh Config
+
+While constructing the UPClientZenoh object, the Zenoh config needs to be assigned.
```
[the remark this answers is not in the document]
evshary: Okay. I think removing this part in the initial version will not affect the interoperability.
Why we need Zenoh config is that we need a way to configure how Zenoh works.
Let's discuss it a bit more in the meeting.

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+    number: Id({01, 02, 03, 10, 11, 12})
+  },
+  entity: empty,
+  resource: empty,
+}
+----
+
+The UAuthority should be serialized into mirco form.
+It should be length + serialized UUri in lowercase hexadecimal representation, which is `06` + `0102030a0b0c`.
+There is no UEntity and UResource, so it is the special form UUri.
```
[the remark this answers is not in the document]
evshary: uStreamer needs it. When the UUri only contains UAuthority and keeps UEntity and UResource blank, it means we only want to listen to all the data from a certain UAuthority.
Maybe my description is a little misleading. Let me make some adjustments.

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+Then the rest of the UUri can be put in the last part.
+
+To be more specific:
+
+* UUri without UAthourity: `up/[byte representation of the UUri]`
+* UUri with UAthourity: `up/[byte representation of the UAuthority]/[byte representation of the rest of the UUri]`
+
+Note that there is a special case. If the UUri only contains UAthority with empty UEntity and UResource,
+it's defined as a special form of Zenoh key `up/[byte representation of UAuthority]/**`.
+This is only used by uStreamer.
```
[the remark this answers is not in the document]
evshary: Makes sense. This should not be limited to use in uStreamer.

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+  authority: UAuthority {
+    name: UAuthName,
+    number: Id({01, 02, 03, 10, 11, 12})
+  },
+  entity: empty,
+  resource: empty,
+}
+----
+
+The UAuthority should be serialized into mirco form.
```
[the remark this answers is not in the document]
evshary: Sounds good!

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+  resource: UResource {
+    name: "door",
+    instance: "front_left",
+    message: "Door",
+    id: 5678,
+  }
+}
+----
+
+There is no UAuthority, so we can ignore the second part.
```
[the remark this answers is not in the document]
evshary: The complete Zenoh key is `up/<uauthoriy>/<uentity+uresource>`
Since there is no UAuthority in the local URI, we only have `up/<uentity+uresource>`.
That's what I mean ignoring the second part, which is `<uauthority>`

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+|===
+
+=== URI Mapping
+
+UUri should be able to be serialized into link:../basics/uri.adoc#42-micro-uris[micro form].
+Then we can transform the micro-form UUri, which should be `u8` array, into string in lowercase hexadecimal representation.
+
+Zenoh key for uProtocol consists of three parts and they are separated by `/`.
+Zenoh key **MUST** start with `up/`.
+If UAuthority is defined in UUri, it should be put in the second part.
```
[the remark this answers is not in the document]
evshary: To be honest, I'm not pretty sure whether I should use "SHOULD" or "MUST" here. If other language implementations don't follow the Zenoh key transformation, we will lose the interoperability. Does that mean we should use MUST here?

up-l1/zenoh.adoc
evshary: `Encoding::WithSuffix` consists of an encoding type and an extra string.
We use the customized encoding type here and put the UPayloadFormat into the extra string.
```rust
Encoding::WithSuffix(
    KnownEncoding::AppCustom,
    payload.format.value().to_string().into(),
)
```
That's why I do `to_string()` here.

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+| UAttribute object encoded into protobuf
+|===
+
+Besides the user attachment, some features in the UAttribute can be mapped into Zenoh transport settings.
+
+==== Message Type
+
+There are 3 kinds of message types in uProtocol (publish, request, response).
+Different Zenoh operations should use the corresponding message types.
+
```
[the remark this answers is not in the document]
evshary: Different language implementations might have different API names. Maybe I should replace operations with API to make it clear...

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,141 @@ Zero Overhead Network Protocol.
...
+This can reduce the unnecessary serialization on payload, which mostly takes time.
+
+User attachment supports key-value pairs, and the following table is the key-value pairs using now.
+
+[cols="1,1"]
+|===
+| key | value
+
+| "uattr"
+| UAttribute object encoded into protobuf
```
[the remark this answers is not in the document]
evshary: > Now there are two options for sending the protobuf, as google.protobug.Any (meaning it also contains the schema) or byte[] without the scheme. If we start having to support v1 and v2 of UAttributes, we'll need to pass the scheme or version somewhere.

I'm not a protobuf expert. Do you prefer to use the `google.protobug.Any` with schema?

evshary:
@stevenhartley @PLeVasseur Thanks for your review. Maybe we can discuss the rest reviews in the weekly meeting.

up-l1/zenoh.adoc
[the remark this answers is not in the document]
stevenhartley: Given if we don't implement this, there will be IoP issues, it is a MUST.

up-l1/zenoh.adoc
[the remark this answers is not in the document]
stevenhartley: No because it is too big (it passes a string scheme). Lets pass the version as a second attachment as needed.

up-l1/zenoh.adoc
```diff
@@ -5,7 +5,7 @@
 The key words "*MUST*", "*MUST NOT*", "*REQUIRED*", "*SHALL*", "*SHALL NOT*", "*SHOULD*", "*SHOULD NOT*", "*RECOMMENDED*", "*MAY*", and "*OPTIONAL*" in this document are to be interpreted as described in https://www.rfc-editor.org/info/bcp14[IETF BCP14 (RFC2119 & RFC8174)]
 
 ----
-Copyright (c) 2023 General Motors GTO LLC
+Copyright (c) 2024 ZettaScale Technology
```
sophokles73, line 8: I would suggest to use
```
SPDX-FileCopyrightText: 2024 Contributors to the Eclipse Foundation
```
as suggested by https://www.eclipse.org/projects/handbook/#ip-copyright-headers and https://reuse.software/tutorial/#step-2 and then remove the redundant entry below the license header.

up-l1/zenoh.adoc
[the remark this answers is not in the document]
sophokles73: @stevenhartley The core-api seems to only define [UAttributes v1](https://github.com/eclipse-uprotocol/up-core-api/blob/main/uprotocol/uattributes.proto), i.e. I cannot find a definition using package name `uprotocol.v2`.

up-l1/zenoh.adoc
```diff
@@ -35,9 +35,138 @@ Zero Overhead Network Protocol.
...
+
+[cols="1,1"]
+|===
+| key | value
+
+| "uattr"
+| UAttribute object encoded into protobuf
+|===
+
+Besides the user attachment, some features in the UAttribute **SHOULD** be mapped into Zenoh transport settings.
```
sophokles73, line 58: This seems rather unspecific: *some* features (which?) *SHOULD* be mapped differently (how?) ...
Or does this refer to the following sections? If so, I guess we could simply remove this sentence as it does not provide any guidance.

stevenhartley: There is no v2 of the protocol, we have to finish defining and implementing v1 and then only afterwards can we consider breaking it to define v2 :-).

evshary: Yes, it refers to the following sections. I'll remove the sentence. Thanks!

up-l1/zenoh.adoc
evshary: Do you mean removing the line `SPDX-FileCopyrightText: 2024 Contributors to the Eclipse Foundation` and only use `Copyright (c) 2024 Contributors to the Eclipse Foundation` instead?

PR #55, opened 2024-01-19
Fix direction of UML composition relation

sophokles73:
Also fixed cardinality at UAuthority end.


stevenhartley (approved):
Thanks @sophokles73

Issue #65, opened 2024-02-06
uTransport is inconsistent with uMessage definition


sophokles73:
The [uTransport's send() method](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/README.adoc#21-send) is defined to accept three parameters. The first one being a UUri which is supposed to represent the *Destination for the UPayload*.

The [uMessage protobuf file](https://github.com/eclipse-uprotocol/up-core-api/blob/main/uprotocol/umessage.proto) defines the uMessage correspondingly:

```protobuf
message UMessage {
    // The source (address) of the message. For published events this is the producers publish topic
    // for requests it is the calling uE respose topic (who sent the request). For responses this is the
    // method URI from the uService
    UUri source = 1;

    // uProtocol mandatory and optional attributes
    UAttributes attributes = 2;

    // Optional message payload containing the data to be sent
    UPayload payload = 3;
}
```

However, the first (UUri typed) property is defined as the *source* (address) instead of the destination. The destination address of the message seems to be contained in the UAttributes' `sink` property.

stevenhartley:
@sophokles73 thanks for pointing that out, the verbiage in up-java is right but not in the spec, I tried to fix it in #66, please review, thank you!

stevenhartley:
Merged fix

tamarafischer:
@sophokles73 "The first one being a UUri which is supposed to represent the Destination for the UPayload."

The first one is the **source**, meaning who is the **owner** of the payload.
The is used to provide the routing system with a very quick way of knowing the owner - someone publishing a state change or someone requesting a state change, and then the fan-out can be optimized without deserializing the UAttributes.
We know that routers/dispatchers should never touch the UPayload.

PR #66, opened 2024-02-07
Fix uTransport send() definition to map source.

stevenhartley:
#65

### DIFF
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -67,7 +67,7 @@ NOTE: uProtocol data model used by the uTransport interface are declared in http
 `UListener` is an interface used for receiving messages by the client uE from the uTransport layer. Listeners are used for receiving messages asynchronously by the uTransport layer.  
 
   * Type name *MUST* be `UListener`
-  * *MUST* contain the method `UStatus onReceive(UMessage&)` that is called by the Transport to notify (callback) the client
+  * *MUST* contain the method `void onReceive(UMessage&)` that is called by the Transport to notify (callback) the client
 
 
 == uTransport
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -97,7 +97,7 @@ The API signature:
 
 | UUri
 
-| Destination for the UPayload
+| The source address for the message, (ex. publish topic, the return address for rpc-request, or the rpc method for rpc response)
 
 | UPayload
 a| Data and metadata to be sent
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -67,7 +67,7 @@ NOTE: uProtocol data model used by the uTransport interface are declared in http
 `UListener` is an interface used for receiving messages by the client uE from the uTransport layer. Listeners are used for receiving messages asynchronously by the uTransport layer.  
 
   * Type name *MUST* be `UListener`
-  * *MUST* contain the method `UStatus onReceive(UUri&, UPayload&, UAttributes&)` that is called by the Transport to notify (callback) the client
+  * *MUST* contain the method `UStatus onReceive(UMessage&)` that is called by the Transport to notify (callback) the client
 
 
 == uTransport
@@ -78,7 +78,7 @@ The transport layer API responsible for point-2-point communication. The interfa
 
 === Send()
 
-Publish/send UPayload and UAttributes to a UUri (topic). 
+Transmit a UMessage.  
 
 
 Communication protocols (a.k.a. uProtocol Transports) define their own Protocol Data Unit (PDU) that comprises of header and payload. Some transports header parameters map already to uProtocol UAttributes. Sending UAttributes in the transport's payload as well as in the header increases overhead for little to no gain. To address this issue,  <<Send()>> allows flexibility for uTransport implementers to define however they see fit what gets mapped into their transport header vs payload. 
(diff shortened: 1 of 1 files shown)


sophokles73 (commented):
This is now consistent with the protobuf file. However, it feels very strange that when invoking the `send()` method, I need to specify the source instead of the destination that I want to send the message to. Intuitively, the destination address seems to be the most important parameter while only some messages will require including an origin (source) address, e.g. for specifying the reply-to address. So, from a developer's perspective I would expect to find the destination address as the first parameter to the `send` method and the reply-to/source address as an (optional) property of the UAttributes. Is there a particular reason why you did it the other way around?

stevenhartley:
@sophokles73 understood and agree, will revise based on our discussion.

up-l1/README.adoc
```diff
@@ -87,32 +87,24 @@ For example, if HartleyTransport can map only UPriority to its PDU header, then
 
 The API signature:
 
-`OUT UStatus send(UUri&, IN UPayload&, IN UAttributes&)`
+`OUT UStatus send(UMessage&)`
```
PLeVasseur, line 90: :+1:

up-l1/README.adoc
```diff
@@ -67,7 +67,7 @@ NOTE: uProtocol data model used by the uTransport interface are declared in http
 `UListener` is an interface used for receiving messages by the client uE from the uTransport layer. Listeners are used for receiving messages asynchronously by the uTransport layer.  
 
   * Type name *MUST* be `UListener`
-  * *MUST* contain the method `UStatus onReceive(UUri&, UPayload&, UAttributes&)` that is called by the Transport to notify (callback) the client
+  * *MUST* contain the method `UStatus onReceive(UMessage&)` that is called by the Transport to notify (callback) the client
```
sophokles73, line 70: does this mean that the listener can return a `UStatus` to the transport implementation? Is there a particular expectation how that should be used and what the transport implementation should do about it?

sophokles73 (commented):
IMHO you should also adapt the signature of `Receive()` accordingly.

up-l1/README.adoc
stevenhartley: @sophokles73 this has been a long standing debate and there are two school of thoughts from developers, some want to return (so the operations can be part of a flow) and others say "it is a callback, it shouldn't do anything". I would love to hear others opinions but to this date there is no defined behavior as to what we should do if the return is anything but OK.

stevenhartley: Done, I checked and up-java returns void anyways

tamarafischer:
@sophokles73 and people, uProtocol mission statement is connecting uE everywhere and anywhere.
We are using service oriented architecture with strong contracts using the protobuf files and we are optimizing on message sending using pub/sub patterns.
In the pub/sub patterns the sender **does not care** about receivers - it just sends out messages.
It can optimize that is no one is subscribed, nothing is sent.

" I need to specify the source instead of the destination that I want to send the message to" - correct! This is to enable routers/dispatchers to implement the mission statement of connecting everyone everywhere.

An owner of data (service oriented) can publish a state change - the platforms job is to deliver this state change to interested parties. Interested parties register as such by subscribing to said state change topic.

Routers and dispatchers are interested in the source since their job is knowing who is subscribed, not the owners of the data.

Separation of concerns in this way, gives the protocol all the wiggle room it is needs to optimize each system delivery.

tamarafischer:
If RPC is what is confusing, then I have a ton to say about that and it is more than a comment typing. But it should be solved in the RPC part and not in the actual UTransport part.

sophokles73:
@tamarafischer there is no dissent about what you stated. I also agree with

> In the pub/sub patterns the sender does not care about receivers - it just sends out messages.

and I would be inclined to append

> to an address.

because in the end, the receiver(s) of the message are represented by this address, regardless of whether it is zero, one or multiple receivers.

I also believe that when it comes to defining a developer-facing API, we should try to make the API as easy to use and comprehensible as possible. While I can easily see that a message is being published/sent-out *to* an address/topic/receiver (whatever you might want to call it), I have a hard time understanding why I need to put the address that I want to send the message *to* into a parameter that is called *source* (even if at the transport level, calling it like that might make sense).

Moving the *source* parameter into `UAttributes` does not really solve this issue FMPOV, but I guess we can alleviate this problem by means of the `UAttributesBuilder` factory methods which we can use to hide the mapping of the publish address to the corresponding `UAttributes` property.

stevenhartley:
> @tamarafischer there is no dissent about what you stated. I also agree with
> 
> > In the pub/sub patterns the sender does not care about receivers - it just sends out messages.
> 
> and I would be inclined to append
> 
> > to an address.
> 
> because in the end, the receiver(s) of the message are represented by this address, regardless of whether it is zero, one or multiple receivers.
> 
> I also believe that when it comes to defining a developer-facing API, we should try to make the API as easy to use and comprehensible as possible. While I can easily see that a message is being published/sent-out _to_ an address/topic/receiver (whatever you might want to call it), I have a hard time understanding why I need to put the address that I want to send the message _to_ into a parameter that is called _source_ (even if at the transport level, calling it like that might make sense).
> 
> Moving the _source_ parameter into `UAttributes` does not really solve this issue FMPOV, but I guess we can alleviate this problem by means of the `UAttributesBuilder` factory methods which we can use to hide the mapping of the publish address to the corresponding `UAttributes` property.

@tamarafischer we are not changing the definition of source and sink, just making the transport API more intuitive.

Issue #68, opened 2024-02-21
Validation of UUri when sending Publish, Request, Response type -- builders, send(), invokeMethod()


PLeVasseur:
In the [`UTransport::send()`](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#21-send) spec we can see that there's these lines:
> * SHOULD pass resolved UUris (contains both names and numbers)
> * SHOULD use the Miro Uri serialization format

And in the [`RpcClient::invokeMethod()`](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l2/rpcclient.adoc#3-implementation) spec it's written:
> Implementations communicate to their underlining transport through the [uTransport Interface](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/README.adoc#_utransport).

I think in-practice we've strayed away from the above requirement that `RpcClient::invokeMethod()` be using `UTransport::send()`, particularly in the Zenoh implementation, due to the ability to use Zenoh Queryables.

I see `SHOULD` is used to specify that resolved UUris are what should be passed in and also that it `SHOULD` use Micro Uri serialization format.

A few points I think I'd like to see addressed:

1. Should `up-foolang` client libraries be opinionated about checking the UUris fit Micro Uri format or are resolved Uris when they are making builders? I'm looking at the [`Source`](https://github.com/eclipse-uprotocol/up-spec/blob/ceb40c3e218ba29241f8cc8f10aa141a7a2d3c84/languages.adoc#2-source) section on the languages spec page where there's a discussion that builders can be provided.
   * In particular, there's [some work being done](https://github.com/eclipse-uprotocol/up-rust/pull/46) over in `up-rust` to update to add a `UMessageBuilder` and reviewing the code, I was reminded of requirements from the spec. Feedback here would help.
2. Should `up-client-bartransport-foolang` `UTransport::send()` implementations be checking one or both of the above `SHOULD` lines from the spec?
3. Should `up-client-bartransport-foolang` `RpcClient::invokeMethod()` implementations be forced to use `UTransport::send()`? If not we should probably:
   * update the spec to clarify they can have a completely separate implementation of `RpcClient::invokeMethod()` which does not depend on `UTransport::send()` and also
   * add any UUri requirements we come up with for the 2. item above

Tagging @sophokles73, @stevenhartley for visibility (for some odd reason Daniel's handle doesn't come up with I "@" in this issue)

sophokles73:
[...]

PR #75, opened 2024-03-05
MQTT5 Protocol spec

devkelley:
This is a draft of how a uTransport would communicate over MQTT:

I would envision that there is a device specific mqtt broker and a cloud based mqtt broker (or more). If a uMessage is being routed to the cloud only the uAuthority information is really needed in the topic. Once the uMessage reaches the correct uDevice, then the header is processed to get the uEntity and uResource from the sink uuri. The uMessage is then passed over the local mqtt broker to the correct component.

Things to discuss:
- Should microURI be used anywhere here?
- The uMessage payload is simply the uMessage protobuf object written to bytes. (vec<u8>)

### DIFF
--- a/up-l1/mqtt.adoc
+++ b/up-l1/mqtt.adoc
@@ -35,8 +35,85 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
 
 For more information, please refer to https://mqtt.org/
 
-
 == Specifications
 
+=== UAttribute Mapping
 
-_Coming Soon_
+MQTT 5 supports custom header fields, and we leverage this to map certain UAttribute values into the MQTT header.
+
+[cols="1,1"]
+|===
+| MQTT Header Property Code | Value (key:value)
+| UserProperty
+| ("upriority","{UAttribute.priority}")
+| CorrelationData
+| "{UAttribute.reqId}"
+| UserProperty
+| ("sink","{UAttribute.sink}") # Should this be serialized to MicroUri?
+|===
+
+The upriority field is needed as mqtt's QoS is very limited, so this will allow listeners to prioritize messages.
+
+The Correlation Data Property Code contains the request correlation id.
+
+The sink field is used to correctly map a message to the correct uE as the topic name may not contain all of the information.
+
+=== URI Mapping
+
+Depending on whether the UURI is local or remote, the MQTT Topic that the message will be published on will be different:
+
+* The remote UUri (with UAuthority): `UUri.UAthority.name`
+* The local UUri (without UAuthority): `UUri.UEntity/UUri.UResource`
+
+Examples:
+
+* The remote UUri (with UAuthority):
+
+[source]
+----
+UUri {
+  authority: UAuthority {
+    name: "Vehicle_1",
+    number: Id({01, 02, 03, 10, 11, 12})
+  },
+  entity: UEntity {
+    name: "body.access",
+    version_major: 1,
+    id: 1234,
+  }
+  resource: UResource {
+    name: "door",
+    instance: "front_left",
+    message: "Door",
+    id: 5678,
+  }
+}
+----
+
+Since this is a remote UUri, the MQTT Topic will simply be `Vehicle_1`.
+
+* The local UUri (without UAuthority):
+
+[source]
+----
+UUri {
+  authority: empty,
+  entity: UEntity {
+    name: "body.access",
+    version_major: 1,
+    id: 1234,
+  }
+  resource: UResource {
+    name: "door",
+    instance: "front_left",
+    message: "Door",
+    id: 5678,
+  }
+}
+----
+
+Since this is a local UUri, the MQTT Topic will be constructed from the UEnity name and the UResource name `body.access/door`.
+
+=== Payload Encoding
+
+The MQTT payload **MUST** be a UMessage that is represented as a byte array to reduce size.
--- a/up-l1/mqtt.adoc
+++ b/up-l1/mqtt.adoc
@@ -62,8 +62,8 @@ The sink field is used to correctly map a message to the correct uE as the topic
 
 Depending on whether the UURI is local or remote, the MQTT Topic that the message will be published on will be different:
 
-* The remote UUri (with UAuthority): `UUri.UAthority.name`
-* The local UUri (without UAuthority): `UUri.UEntity/UUri.UResource`
+* The remote UUri (with UAuthority): `upr/UUri.UAthority.name`
+* The local UUri (without UAuthority): `upl/UUri.UEntity/UUri.UResource`
 
 Examples:
 
@@ -90,7 +90,7 @@ UUri {
 }
 ----
 
-Since this is a remote UUri, the MQTT Topic will simply be `Vehicle_1`.
+Since this is a remote UUri, the MQTT Topic will simply be `upr/Vehicle_1`.
 
 * The local UUri (without UAuthority):
 
@@ -112,7 +112,7 @@ UUri {
 }
 ----
 
-Since this is a local UUri, the MQTT Topic will be constructed from the UEnity name and the UResource name `body.access/door`.
+Since this is a local UUri, the MQTT Topic will be constructed from the UEnity name and the UResource name `upl/body.access/door`.
 
 === Payload Encoding
 


lynn2009:
"the uAuthority information is really needed in the topic." -- MQTT broker provides multiple ways to distribute messages (MQTT pub/sub, streaming bridge etc), the streaming bridge may also use other uAttributes, for example proriity.

devkelley:
> "the uAuthority information is really needed in the topic." -- MQTT broker provides multiple ways to distribute messages (MQTT pub/sub, streaming bridge etc), the streaming bridge may also use other uAttributes, for example proriity.

Do you think other uAttributes should be present in the topic? The current proposal has several uAttributes being passed as custom header props. Would that work for something like the streaming bridge?

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,85 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+
+The Correlation Data Property Code contains the request correlation id.
+
+The sink field is used to correctly map a message to the correct uE as the topic name may not contain all of the information.
+
+=== URI Mapping
+
+Depending on whether the UURI is local or remote, the MQTT Topic that the message will be published on will be different:
+
+* The remote UUri (with UAuthority): `upr/UUri.UAthority.name`
```
PLeVasseur, line 65: I'm less familiar with what's considered 'typical' for MQTT5 identifiers that are published on / subscribed to.

Over in the Zenoh spec + implementation, we went with Micro Form, since it's [there's this bit in the spec](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#211-parameters):
image
which to my mind makes Micro Form kind of the bar to clear.

If Resolved URIs are available (satisfies both Micro + Long Forms) that works too, but that way at a minimum only Micro Form is needed.

This does degrade the "reading the URI on the wire in human-readable format" experience though.

WDYT?

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,85 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+
+[cols="1,1"]
+|===
+| MQTT Header Property Code | Value (key:value)
+| UserProperty
+| ("upriority","{UAttribute.priority}")
+| CorrelationData
+| ("reqId", "{UAttribute.reqId}")
+| UserProperty
+| ("sink","{UAttribute.sink}") # Should this be serialized to MicroUri?
```
PLeVasseur, line 52: I dunno enough about MQTT5, but from what I can see of the spec in [my previous comment over here](https://github.com/eclipse-uprotocol/up-spec/pull/75/files#r1522025611), probably?

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,85 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+}
+----
+
+Since this is a local UUri, the MQTT Topic will be constructed from the UEnity name and the UResource name `upl/body.access/door`.
+
+=== Payload Encoding
 
-_Coming Soon_
\ No newline at end of file
+The MQTT payload **MUST** be a UMessage that is represented as a byte array to reduce size.
```
matthewd0123, line 119: Thought on this one. Could we just use the UMessage Protobuf's "SerializeToString()" Method to serialize the payload before publishing it in MQTT? Would keep it consistent between different languages.

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,85 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+  resource: UResource {
+    name: "door",
+    instance: "front_left",
+    message: "Door",
+    id: 5678,
+  }
+}
+----
+
+Since this is a local UUri, the MQTT Topic will be constructed from the UEnity name and the UResource name `upl/body.access/door`.
```
matthewd0123, line 115: I'm wondering if it would be worth it to carry along the version_major inside the topic. That way, we could differentiate between people publishing to one version of the UEntity vs another. It could look something like this:

upl/body.access/1/door

up-l1/mqtt.adoc
devkelley: That is a good question. I am currently using a 'write_to_bytes' protobuf function that is provided by the Protobuf crate and I believe is used by up-client-zenoh. @stevenhartley Do you have an opinion on this?

up-l1/mqtt.adoc
devkelley: I went back and forth on this a bit, and I can see both sides (using micro uri vs a string representation). One thing to consider is that mqtt topics require a string representation so it had initially made sense to use the Authority name and if that became too cumbersome switch it to Micro Uri. One thing that was mentioned to me is that UTransport can expect a resolved uri in a message to transport (am I remembering this correctly @stevenhartley?) and if that is the case I dont think there is a ton of value in serializing the topic name.

up-l1/mqtt.adoc
PLeVasseur: Makes sense to use to me! But I'm not named Steven :slightly_smiling_face:

up-l1/mqtt.adoc
PLeVasseur: Yeah. Zenoh also requires a string for their key expression (KeyExpr). They end up serializing and chopping up, then arranging in a string. [Here's a link](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/blob/8855b9abd4bd27228d30d9061522194f330fa547/src/lib.rs#L89-L116) to where that's done for reference.

PLeVasseur: Related to #68

Specifically, [my last comment](https://github.com/eclipse-uprotocol/up-spec/issues/68#issuecomment-1994379074), with the remaining question from that issue.

stevenhartley: All of these discussions are making me believe we need to bring back the short form URI. This was a string version of the URI using the numbers in lieu of names (long) and in lieu of in byte[] (micro). 

Short form URI then would be:

`up://192.168.1.100/2/1.1/8000`

In the that example, the UAuthortiy is an IP address, the uService ID is 2 and version 1.1 and the topic ID is 8000. If this is valuable and can be used for MQTT, I can add it back to the spec and the language libraries but it will miss the 1.5.7 deadline tomorrow. @devkelley @sophokles73 @PLeVasseur WDYT?

up-l1/mqtt.adoc
stevenhartley: Payload needs to be UPayload and in it there is the UPayloadFormat that says how the data is serialized. We cannot assume a given format, this is also why I suggested that we use the MQTT5 binary form so we don't have to make the MQTT PDU a string (it could be binary)

up-l1/mqtt.adoc
stevenhartley: We need to use names.

lynn2009: Does the UResource include the character '#'? In MQTT topics, '#' is considered a reserved character. The structure of the topic is crucial for MQTT client subscriptions and for performing authorization checks.  

We have to have different topic strategies for message type:

-  For pub message, the MQTT Topic will be constructed from the source uEntity name and the UResource 

-  For req, res, and notification message,  the MQTT Topic will be constructed from the sink uEntity name and the UResource

Including a version in the topic allows for better management of message routing and ensures compatibility between different versions of the services or payloads being communicated. This practice can be particularly useful in scenarios where you need to update the format or content of the messages being published without disrupting existing subscribers who might expect data in an older format.

up-l1/mqtt.adoc
lynn2009: If the Payload is UPayload, then we have to map the UAttributes to MQTT 5 User Properties.  
If the Payload is UMessage, we still have to map some UAttributes to MQTT 5 User Properties for routing purpose.

up-l1/mqtt.adoc
PLeVasseur: > Does the UResource include the character '#'? In MQTT topics, '#' is considered a reserved character. The structure of the topic is crucial for MQTT client subscriptions and for performing authorization checks.

[The long form does](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#11-rfc3986) -- specifically for the "fragment", i.e. [the message](https://github.com/eclipse-uprotocol/up-core-api/blob/dc353878914b3459edf2e04531e9cadfc752a2e9/uprotocol/uri.proto#L96) portion of UResource.

sophokles73: I have created a mapping of UUris to MQTT topic names in https://github.com/eclipse-uprotocol/up-spec/issues/73#issuecomment-1993445795.

Maybe that can serve as inspiration?

up-l1/mqtt.adoc
devkelley: @lynn2009 That is a valid point. The current draft has all of UMessage being passed as the mqtt payload and select UAttributes being duplicated in the MQTT 5 User Properties section. I think the current spec for a UTransport allows for a variation of what is considered payload and what is in the header as long as the UPayload is not altered. I believe this is achieved with the current proposal but I am open to enforcing all of the UAttributes into the MQTT 5 User Properties and just pass the byte serialized uPayload. It will just require some reconstruction of the UAttributes if the message needs to be sent to a different MQTT Broker (ie. getting sent off device)

up-l1/mqtt.adoc
devkelley: @sophokles73 Overall I think that mapping would be good for cloud. Some questions would be:
1. Do we think the topic names would get to long in this scenario. If so, we could consider the short_form uri that @stevenhartley mentioned.
2. Would there be a reason to distinguish between local and remote uri's as currently proposed and implemented in Zenoh? I dont necessarily think there would beyond shortening the topic for local communication.

@PLeVasseur @lynn2009 Any thoughts? Ill go ahead and update the draft if not

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,85 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
-
 == Specifications
 
+=== UAttribute Mapping
+
+MQTT 5 supports custom header fields, and we leverage this to map certain UAttribute values into the MQTT header.
+
+[cols="1,1"]
+|===
+| MQTT Header Property Code | Value (key:value)
```
matthewd0123, line 46: Would we be able to include UMessageType in here as well?

devkelley: Sure I will add that to the spec

up-l1/mqtt.adoc
lynn2009: The mapping strategy is ok for the vehicle side. However, when considering the flow of messages between the device and the cloud, the mapping might cause confusion, even for the in-vehicle uStreamer. The in-vehicle uStreamer is configured to subscribe to any messages coming from the cloud to the vehicle using a wildcard, but it does not subscribe to messages it sends out. Due to the mapping, both C2D (Cloud-to-Device) and D2C (Device-to-Cloud) published message types begin with the same authority, leading to potential confusion.

To alleviate this issue, we could consider introducing a topic prefix, which seems like a viable option within UTransport. For instance, messages from the Cloud to Vehicle could use the prefix c2d/<authority>, while messages from the Vehicle to the Cloud could use d2c/<authority>.

lynn2009: For topics that are frequently used, having a lengthy topic name can lead to increased bandwidth consumption during message publication. Utilizing MQTT 5.0 Topic Aliases could minimize that.  But there's typically a limitation on the number of aliases in a session.

devkelley: If I am understanding correctly, because a 'publish' message would use the device's uauthority it would cause confusion because any messages sent to the device from the cloud would start with the same uauthority. And then your suggestion would be to have a prefix like `c2d/` that the device would subscribe to with its uauthority id to get messages on? From the local message perspective, it may make sense to have some prefix like `upl` still.

devkelley: For the topic naming itself, I think if we serialize the UURI to short form we can just use that for the topic names (with the appropriate prefix). That should be short enough

up-l1/mqtt.adoc
devkelley: I updated the spec so the sink uri be short form serialized in the header.

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,100 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+  resource: UResource {
+    name: "door",
+    instance: "front_left",
+    message: "Door",
+    id: 5678,
+  }
+}
+----
+
+* A Publish Message: `d2c/192.168.0.0/1234/1/5678`
```
matthewd0123, line 105: 2 questions:
1. Why are we selecting IPs and IDs instead of the names?
2. Question I had asked in slack: How do we build topics when the UMessageType is unknown, which is true for register_listener()

devkelley: 1. @lynn2009 and others I think have concerns about the topics getting too long, and this follows the short uri format for the UURIs. @stevenhartley any other thoughts?

matthewd0123: Oh okay all good, just wanted to understand why the change happened

PLeVasseur:
@devkelley -- I'm not familiar with MQTT5 yet, but in order to support the use-case of use by a uStreamer, could you add into the spec how we'd handle the situation [outlined in the Zenoh spec for URI Mapping](listener_wrapper) in the case that a uStreamer wants to subscribe to "everything" from a particular UAuthority?

In particular, talking with @evshary, we came up with this:
image
to allow creation of a Zenoh KeyExpr which would be able to "see" anything going on a certain UAuthority.

And then CY wrote `up-client-zenoh-rust` in such a way that that special `UUri` containing only a `UAuthority` would result in registering and unregistering from everything as appropriate (publish, notification, request, response). For reference you could take a look [here](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/blob/8855b9abd4bd27228d30d9061522194f330fa547/src/utransport.rs#L542-L581).

Hope that all made sense -- if not we can chat here or on Slack :slightly_smiling_face:

up-l1/mqtt.adoc
mnemili: If we are able to send UAttributes in MQTT 5 User Properties, why cant we just send the UPayload in the payload? We dont need to duplicate again and increase the data over the wire.

up-l1/mqtt.adoc
```diff
@@ -129,6 +132,23 @@ UUri {
...
+    name: "Vehicle_1",
+    number: 192.168.0.0,
+  },
+  entity: empty,
+  resource: empty,
+}
+----
+
+* If in cloud, uses wildcard to subscribe to all messages from the UAuthority: `d2c/192.168.0.0/+/+/+`
+* If on device, uses wildcard to subscribe to all messages from the UAuthority: `c2d/192.168.0.0/+/+/+`
```
PLeVasseur, line 150: Is "in cloud" vs "on device" determined by when we construct a `UPClientMQTT5`, passing in a parameter to dictate that?

PLeVasseur: What if we want to get messages from the cloud at `uauthority_1` and messages from other devices locally within the vehicle at `uauthority_2`?

Want to make sure we're sort of not limiting ourselves into a certain topology here prematurely :slightly_smiling_face:

Tagging in @stevenhartley for sanity check

devkelley: To your first question, yes my thought was the UPClient would have some initialization value to determine if its running on the device and connected to a remote mqtt, running on the device and connected to a local mqtt or running in the cloud.

devkelley: However, i do see your point that multiple hops may break this system down

up-l1/mqtt.adoc
devkelley: My original thought was to minimize the header size and not have to reconstruct the UAttributes from multiple places, but I am open to passing all the UAttributes in the header and just passing uPayload in the payload portion of the message. It will require mapping on both ends of the message (send and receive) but I dont think thats a big deal. Any concerns @matthewd0123 @PLeVasseur @lynn2009?

up-l1/mqtt.adoc
devkelley: Okay one concern I have in general is how EntityA on a device with no direct connection to the cloud signals to EntityB (with direct connection to the cloud) that it wants to receive information from a different uAuthority. EntityB will have 2 upClients (one connected to the cloud mqtt and one connected to the remote mqtt).

devkelley: A simple register listener /uAuthority_2/+/+/+ doesnt work for EntityA since EntityB doesnt know that EntityA is listening on that topic. EntityB would need to get some form of notification that EntityA is listening on that remote topic for the local mqtt broker so that it can listen to the remote topic as well (on the cloud mqtt broker) to forward along the messages to the local mqtt broker

up-l1/mqtt.adoc
PLeVasseur: @devkelley -- I was re-reading the [relevant part of the spec](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#21-send) to see if there was any guidance, as I recalled that there was explicit mention of putting info in transport-specific header and deduplicating in the transport payload.

I don't see any guidance on where to place the `UPayload`, but I suppose this is probably also an implementation detail of how `up-client-mqtt5-rust` would work. So long as both ends of the pipe know what to expect, should work.

As far as what's best to do -- I'm not clear on any performance gain/loss on where to put the UPayload.

tl;dr: I don't see any real problem with the proposal, but tagging @stevenhartley just to be safe :slightly_smiling_face:

mnemili: There are few reasons for the uAttributes to be in the clear as lot of commercial components may have to handle them. Even some components in cloud either might want to consume these uAttributes without having to deserialize uMessage proto. So another option is good where the uMessage has uHeaders in json and uPayload is proto. That probably is better option.

devkelley: I agree that being clear where uAttributes are in the transport message is important and I think it does make sense to keep the uAttributes together in an easily accessible location in the transport message to minimize routing work. My knowledge is lacking a bit here so I am unsure what the performance hit (from device to cloud) would be to have uAttributes serialized as json in the transport message. As far as mapping the values when switching from MQTT to another transport, my understanding is that a uTransport defines its own scheme for a uMessage, so regardless of how a MQTT uTransport handles a uMessage, the transport would need to convert back to a uMessage before passing onto the next transport.

mnemili: We will have to compare these options for data on the wire.  Payload may never go over 50 bytes but the uAttributes may come to 300 bytes however we send to make routing choices. In the other option, we have performance hit with deserializing and serializing billions of messages.

up-l1/mqtt.adoc
stevenhartley: > my understanding is that the short uri format can have the uAuthority be either the ip address or an id. @stevenhartley is that correct?

Correct. So we would use ID (the VIN) for talking with the vehicle.

up-l1/mqtt.adoc
stevenhartley: > Okay one concern I have in general is how EntityA on a device with no direct connection to the cloud signals to EntityB (with direct connection to the cloud) that it wants to receive information from a different uAuthority. EntityB will have 2 upClients (one connected to the cloud mqtt and one connected to the remote mqtt).

No the D2D communication is through the streamer vs uE-2-uE communication is through a MQTT5 broker. So a uE only uses one up-client library talking to a local broker.  The streamer yes will have 2 different up-client (uTransport) instances, one for talking to other devices (streamers) and one to talk to the local broker.

up-l1/mqtt.adoc
devkelley: I updated the spec to at least put all of the uATtributes in the mqtt header properties as a starting place.

up-l1/mqtt.adoc
devkelley: @PLeVasseur @stevenhartley @lynn2009  What are your thoughts on this approach?:

location_identifier:
- upr: Remote destination
- upl: Local destination

topic uri mapping:

  `{location_identifier}/{UUri.UAuthority.number (remote only)}/{UUri.UEntity.id}/{UUri.UEntity.version_major}/{UUri.UResource.id}`

Publish uses source UUri and all other message types use sink UUri when sending a message.

subscribing (registerListener):

  If subscribing to a different UAuthority, will subscribe to all Publish message types on the UUri:

    `upr/{UUri.UAuthority.number}/+/+/+` where `UAttribute.type == Publish`

  if subscribing to the own authority, will subscribe to all messages except Publish message type:

    `upr/{UUri.UAuthority.number}/+/+/+` where `UAttribute.type != Publish`

This should avoid the scenario where an authority is publishing messages to the cloud and trying to receive messages at the same topic since source and sink may match.

Alternatively, the Listener could sort messages based on whether the `source` and `topic` match at the uAuthority level.

PLeVasseur: Hey @devkelley -- sorry, I missed this.

We ended up rolling back our thoughts about separating `upl/` vs `upr/` over in the Zenoh case and Zenoh always does a "global" send onto `up/`. Is the concept of remote vs local at the key/header level still needed here for MQTT? (again, don't know MQTT at all yet).

-------

Speaking from a uStreamer perspective:

> if subscribing to the own authority, will subscribe to all messages except Publish message type:
> 
> ```
> `upr/{UUri.UAuthority.number}/+/+/+` where `UAttribute.type != Publish`
> ```

This matches what I'd expect. In fact, because I assume that this portion:
> where `UAttribute.type != Publish`

would not be filtered for, I currently [have an explicit check to throw away any Publish messages](https://github.com/eclipse-uprotocol/up-streamer-rust/pull/7/files#r1548303397) in my initial PR for the `up-streamer-rust`. :slightly_smiling_face: 

> If subscribing to a different UAuthority, will subscribe to all Publish message types on the UUri:
> 
> ```
> `upr/{UUri.UAuthority.number}/+/+/+` where `UAttribute.type == Publish`
> ```

Again, another feature that's unexpected, but welcome :slightly_smiling_face: I was assuming I'd have to do a similar check to throw away any non-Publish messages.

PLeVasseur: > Okay one concern I have in general is how EntityA on a device with no direct connection to the cloud signals to EntityB (with direct connection to the cloud) that it wants to receive information from a different uAuthority. EntityB will have 2 upClients (one connected to the cloud mqtt and one connected to the remote mqtt).

Could you draw me a simple picture / diagram for this case and help me understand the challenge for MQTT in this case? Perhaps this comes up because MQTT is brokered, unlike Zenoh? Again total MQTT newb.

devkelley: image
So maybe the context I was missing is that EntityB in this scenario would be a uStreamer. But the issue I was thinking about is that the EntityA will subscribe to `up/Authority2/+/+/+` on the local mqtt broker (Orange). How does EntityB get notified that EntityA has a listener for `up/Authority2/+/+/+` since there is no notification that a listener was setup by EntityA?

However, @stevenhartley mentioned that for device to device communication a streamer would be used, so maybe this doesnt apply?

PLeVasseur: I'm going to speak to this as if we're talking about Publish messages only for the moment.

I think that only a uStreamer would be subscribing on the UAuthority alone.

Therefore, if the uE `EntityA` wanted to register for a particular Publish message called `up/Authority2/EntityFoo/ResourceBar`, that'd be coordinated through uSubscription. `EntityA` would call [`Subscribe()`](https://github.com/eclipse-uprotocol/up-spec/blob/36c4a2ded83b5d69fdabca3d7122d70ffa7556b7/up-core-api/uprotocol/core/usubscription/v3/usubscription.proto#L62) on the client uSubscription object, which would then coordinate with the uSubscription service on the device for `Authority1`.

The uSubscription service in the background would then check that this topic is being piped over to us from the cloud (or any other connected device) and if not, make it so.

In addition, the uStreamer would also get signaled that there was a change in subscribers via [`RegisterForNotifications()`](https://github.com/eclipse-uprotocol/up-spec/blob/36c4a2ded83b5d69fdabca3d7122d70ffa7556b7/up-core-api/uprotocol/core/usubscription/v3/usubscription.proto#L93) and perform the routing appropriately so that any `up/Authority2/EntityFoo/ResourceBar` get picked up from the Cloud MQTT and put onto the Local MQTT.

------

Some questions to help me make sure I'm understanding correctly:
1. Are the Local MQTT and Cloud MQTT completely different implementations of `UTransport`? Same?
2. Are the Local MQTT and Cloud MQTT unable to "see" each other due to the brokered nature of MQTT?

devkelley: 1. My intention is that the Local and Cloud clients would be the same implementation of UTransport, just connecting to the different brokers. If this feels too difficult we could change to different implementations.
2. Correct, in this scenario they would not have knowledge of each other and the 'entity' that is using the upclients would be responsible for handling messages between the two clients.

PLeVasseur: Thanks for answers to 1 & 2. Makes sense to me.

I think what I wrote in my above response holds then. In this case `EntityB` would be a uStreamer to bridge the Local MQTT and Cloud MQTT.

That uStreamer would have 2 x instances of `up-client-mqtt5-rust`, one used with the Local MQTT broker and one for the Remote MQTT broker.

Upon getting a signal via uSubscription that _some entity_ on `UAuthority1` wanted to subscribe to `up/Authority2/EntityFoo/ResourceBar` which is on `UAuthority2`, then the uStreamer would subscribe to `up/Authority2/EntityFoo/ResourceBar` from the Cloud MQTT via the `up-client-mqtt5-rust` `UTransport` for the Cloud and publish that same message, but onto the `up-client-mqtt5-rust` `UTransport` for the Local MQTT broker.

devkelley: Okay, I think this was the context that I was missing. Essentially there is some higher level component that will handle signaling around the messages and the `up-client-mqtt5-rust` client will only be responsible for handling a single communication hop. Given this context I think we can drop the `remote` and `local` tagging.

devkelley: @PLeVasseur and just to confirm for zenoh: are there still two kinds of topics depending on local vs remote destination? ie.

up/{Authority}/{Entity}/{Resource}/  for remote

and

up/{Entity}/{Resource}/ for local?

PLeVasseur: @devkelley -- we decided on the Zenoh Key Expressions always being global. So when sending any Publish, Request, Response, or Notification they are `up/UAuthority/UEntity/UResource` in short form.

I don't think it's reflected into the spec yet, though.

@evshary -- can you sanity check me?

devkelley: Ah got it, and if the uAuthority is not present when passed to a function it is added?

devkelley: Okay I updated the uuri mapping to what my understanding currently is, let me know if it doesnt match your understanding :)

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,128 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
 == Specifications
 
+=== UAttribute Mapping
+
+MQTT 5 supports custom header fields, and we leverage this to map the UAttribute values into the MQTT header.
+
+[cols="1,1"]
+|===
+| MQTT Header Property Code | Value (key:value)
+| UserProperty
```
stevenhartley, line 47: In order to drastically save space (each byte is critical), lets use the protobuf defined field numbers in lieu of string names so that it is easier to convert and takes less space.

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,128 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+The type field is used by a UListener to select message types on the topic.
+
+=== URI Mapping
+
+The URI mapping follows the short form UUri format, with the addition of a location identifier to determine if the message is intended for a local or cloud destination.
+
+
+- `up/{UUri.UAuthority.number}/{UUri.UEntity.id}/{UUri.UEntity.version_major}/{UUri.UResource.id}`
+
+  Note: If the UUri does not contain a UAuthority, add the UAuthority.number for the UAuthority the transport client is running on.
```
stevenhartley, line 85: UAuthority number you mean id or IP address?

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,128 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+
+Subscribe to all messages to the current Authority
+
+- `up/{UUri.UAuthority.number}/+/+/+` where `UAttributes.type != Publish`
```
stevenhartley, line 89: If we want this to be a URI then the scheme is `up:` and I would recommend omitting the scheme as I don't think we will have anything other than uProtocol topics right?

stevenhartley (changes requested):

up-l1/mqtt.adoc
evshary: @PLeVasseur You're right. The spec hasn't been updated yet. I'm working on that.

evshary: Initial version here
https://github.com/eclipse-uprotocol/up-spec/pull/103

up-l1/mqtt.adoc
devkelley: Yes, since its oneof in the protobuf definition I think the labeling is 'number'

up-l1/mqtt.adoc
devkelley: Valid point, let me check that there are no concerns on starting off with a number

devkelley: removed

up-l1/mqtt.adoc
devkelley: Alright, I updated the table to use the protobuf defined field numbers. One thing to note, the UserProperty is still a Utf8 encoded string pair in the MQTT 5 specification.

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,126 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+| ("9", "{UAttribute.reqId}") # UUID reqid
+| UserProperty
+| ("10", "{UAttribute.token}") # Optional token
+| UserProperty
+| ("11", "{UAttribute.traceParent}") # Optional traceparent
+|===
+
+All uAttributes are mapped to a MQTT header UserProperty, where the key is the UAttribute protobuf field number.
+
+The source and sink fields **MUST** be a short form serialized UUri.
```
stevenhartley, line 73:
```suggestion
The source and sink fields **MUST** be link:../basics/uri.adoc[Short URI Format].
```

stevenhartley (changes requested):
@devkelley with the last change I think we are good to merge in the initial proposal and we can adapt afterwards. Lets sync up on the progress of up-client-mqtt5-rust later this week

matthewd0123 (approved):

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,126 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+
+The URI mapping follows the short form UUri format, with the addition of a location identifier to determine if the message is intended for a local or cloud destination.
+
+- `{UUri.UAuthority.number}/{UUri.UEntity.id}/{UUri.UEntity.version_major}/{UUri.UResource.id}`
+
+  Note: If the UUri does not contain a UAuthority, add the UAuthority.number for the UAuthority the transport client is running on.
+
+Subscribe to all messages to the current Authority
+
+- `{UUri.UAuthority.number}/+/+/+` where `UAttributes.type != Publish`
```
lynn2009, line 87: Filtering based on UserProperty is not inherently supported by MQTT's native subscribe function. Instead, this type of filtering have to be implemented at the application level.  
whether it's the cloud subscribing to messages from a vehicle or a vehicle subscribing to cloud messages, this approach is inefficient.  specially on Cloud side.

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,126 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+
+Subscribe to all messages from a different Authority
+
+- `{UUri.UAuthority.number}/+/+/+` where `UAttributes.type == Publish`
+
```
lynn2009, line 92: See comments as above

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,126 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+  Note: If the UUri does not contain a UAuthority, add the UAuthority.number for the UAuthority the transport client is running on.
+
+Subscribe to all messages to the current Authority
+
+- `{UUri.UAuthority.number}/+/+/+` where `UAttributes.type != Publish`
+
+Subscribe to all messages from a different Authority
+
+- `{UUri.UAuthority.number}/+/+/+` where `UAttributes.type == Publish`
```
lynn2009, line 91: If the subscribe operation has to use the UUri.uAuthority.number, it is not suitable for cloud to subscribe vehicle messages, given that the VIN is a 17-character string.

up-l1/mqtt.adoc
devkelley: Hmm, what do you think about appending the message type to the topic?

up-l1/mqtt.adoc
devkelley: @lynn2009 are you saying that the vin is too long for the topic name? I am unsure what other identifier to use, other than using the micro uri format for the topic name

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,126 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+| UserProperty
+| ("8", "{UAttribute.commStatus}") # Optional UCode commstatus
+| UserProperty
+| ("9", "{UAttribute.reqId}") # UUID reqid
+| UserProperty
+| ("10", "{UAttribute.token}") # Optional token
+| UserProperty
+| ("11", "{UAttribute.traceParent}") # Optional traceparent
+|===
+
```
mnemili, line 70: Can we have just "id", "type" instead of numerics in the key?

up-l1/mqtt.adoc
stevenhartley: > Filtering based on UserProperty is not inherently supported by MQTT's native subscribe function. Instead, this type of filtering have to be implemented at the application level. whether it's the cloud subscribing to messages from a vehicle or a vehicle subscribing to cloud messages, this approach is inefficient. specially on Cloud side.

@lynn2009 Right now the purpose of this is only for D2C and C2D and once the message is in the cloud, it will use the clouse native uTransport and uSubscription infrastructure (i.e. Eventhub or something else).

up-l1/mqtt.adoc
stevenhartley: UAuthority.number is either id (VIN) or IP address.

up-l1/mqtt.adoc
stevenhartley: I recommended we use numbers in lieu of strings so that we reduce the number of bytes that we send over the wire and given the numbers are fixed in uattributes.proto anyways.

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,126 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
 
 For more information, please refer to https://mqtt.org/
 
-
 == Specifications
 
+=== UAttribute Mapping
+
+MQTT 5 supports custom header fields, and we leverage this to map the UAttribute values into the MQTT header.
```
sophokles73, line 42: if this specification covers MQTT 5 only, then IMHO the document should be renamed to `mqtt5.adoc` and this fact should be made explicit in the Overview section already.

up-l1/mqtt.adoc
sophokles73: If message size is the main concern, then why don't we simply put the protobuf serialization of UAttributes in a single property like we do in the zenoh transport?

up-l1/mqtt.adoc
sophokles73: > Hmm, what do you think about appending the message type to the topic?

Or maybe even put it to the front as the most discriminant field.
Filtering on all Publish messages: `pub.v1/+/+/+/+`
Filtering on all messages destined to myself: `+/{my.authority.id}/+/+/+`

up-l1/mqtt.adoc
sophokles73: IMHO @lynn2009 refers to the fact that the cloud component that wants to receive all Publish messages from all vehicles would need to use a separate topic filter per subscribed topic/vehicle (`vin1/entity.id/1/resource.id`, `vin1/entity.id/1/resource.id`, etc) as opposed to only one filter (`pub.v1/+/+/+/+`) for Publish messages originating from any vehicle ...

mnemili:
@stevenhartley  Why cant we do /{vin}/body.access/{v}/rpc.unlock instead of /{vin}/body.access/{v}/rpc.command? We can verify the permission of the rpc without opening the payload..

up-l1/mqtt.adoc
mnemili: Headers should not have to be deserialized for routing purposes or other purposes.

up-l1/mqtt.adoc
```diff
@@ -35,8 +35,126 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+
+Subscribe to all messages to the current Authority
+
+- `{UUri.UAuthority.number}/+/+/+` where `UAttributes.type != Publish`
+
+Subscribe to all messages from a different Authority
+
+- `{UUri.UAuthority.number}/+/+/+` where `UAttributes.type == Publish`
+
+When sending a message, the source UUri is used for Publish message types and sink UUri is used for all other message types.
```
mnemili, line 93: @stevenhartley  If we do /{vin}/body.access/{v}/rpc.command, we cant verify permission without opening the payload. Instead if we have /{vin}/body.access/{v}/rpc.unlock, that would help - lot of other MQTT brokers had the same opinion.

up-l1/mqtt.adoc
lynn2009: Prefer to have the following patterns:
rpc/{authority}/+/+/+
notf/{authority}/+/+/+
pub/{authority}/+/+/+

This approach enables subscribers to subscribe without needing application-level filtering, making it simpler to implement authorization controls. For instance, the uAuthority would be restricted to subscribing only to topics under rpc/<UUri.uAuthority.number|name>/# and notf/<UUri.uAuthority.number|name>/#, but they would be allowed to publish to topics under pub/<UUri.uAuthority.number|name>/#.

Issue #81, opened 2024-03-13
Define how `UTransport::registerListener()` shall map which `UUri` onto Zenoh KeyExpr for the various `Notification`, `Publish`, `Request`, and `Response` message types


PLeVasseur:
Hey @evshary :wave: --

We were having an internal discussion about which `UUri` to map (source vs sink) and how for [`up-client-zenoh-cpp`](https://github.com/eclipse-uprotocol/up-client-zenoh-cpp).

I also saw the discussion pivot over on #73 via @sophokles73's latest message.

I think it would be very good to document similar to how @sophokles73 did, but for Zenoh so that we have consistency across all `up-client-zenoh-foo`:
![image](https://github.com/eclipse-uprotocol/up-spec/assets/11622119/04d839bd-b02f-4a1f-a7cb-0c1f5e316988)

@evshary: Would you like to take the ball and write-up what this should be in the spec here so that we're all on the same page?

I'd like to tag Michael on here as well, but his name doesn't show up when I @. I'll forward this issue to him.

evshary:
I forgot to link the reply here
https://github.com/eclipse-uprotocol/up-spec/issues/73#issuecomment-1996858847

@PLeVasseur Please let me know if we're not on the same page. 😀

MelamudMichael:
I am good with this

PLeVasseur:
> Possible solution to address this would be to change point 3 in the description above to use the source address as the zenoh key expression, then there are no issues. @evshary, @MelamudMichael, @PLeVasseur , Luca what do you think?

Do you mean the following?
![image](https://github.com/eclipse-uprotocol/up-spec/assets/11622119/da5125de-5e80-4cd7-9d76-911e84fc2c2c)

So the semantics are that for a notification, we register on the source instead of the sink, just like a publish?

stevenhartley:
> So the semantics are that for a notification, we register on the source instead of the sink, just like a publish?

That is correct.

PLeVasseur:
OK! Makes sense :slightly_smiling_face:

stevenhartley:
@evshary if you are ok with the proposed changes, please update the PR #92 and then we can merge 92 and close this issue.

evshary:
After the meeting, we decided to keep using "sink UUri" (or receiver_uuri in the example above) while register_listener for the Notification type.
Regarding the issue that all the notification messages go into the same callback, we agree that developers should discriminate the different types of notification messages in that callback for the time being.
[...]

PR #92, opened 2024-03-19
Update Zenoh Spec with work flow

evshary:
As @PLeVasseur requested in #81, add the work flow in Zenoh Spec.

### DIFF
--- a/up-l1/zenoh.adoc
+++ b/up-l1/zenoh.adoc
@@ -194,11 +194,11 @@ The section provides the details how it works in different message types.
     * source: `up://origin-authority/origin-entity/origin-version/origin-resource.instance#message`
     * sink: `-`
 
-2. `UTransport.send` maps source to Zenoh key
+2. `UTransport.send` maps source (**origin UUri**) to Zenoh key
     * local (w/o origin-authority): `upl/` + `origin-resource-id + origin-entity-id + origin-version`
     * remote (with origin-authority): `upr/` + `origin-authority-id/` + `origin-resource-id + origin-entity-id + origin-version`
 
-3. Subscriber registers a listener for
+3. Subscriber registers a listener on **origin UUri**
     * `up://origin-authority/origin-entity/origin-version/origin-resource.instance#message`
 
 4. which `UTransport.registerListener` maps to Zenoh key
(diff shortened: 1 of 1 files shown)


up-l1/zenoh.adoc
```diff
@@ -178,3 +183,91 @@ Encoding::WithSuffix(
...
+1. A uEntity (`origin-authority/origin-entity`) creates a message with
+    * source: `up://origin-authority/origin-entity/origin-version/origin-resource.instance`
+    * sink: `up://dest-authority/dest-entity/dest-version/dest-resource.instance`
+
+2. `UTransport.send` maps sink to Zenoh key
+    * local (w/o dest-authority): `upl/` + `dest-resource-id + dest-entity-id + dest-version`
+    * remote (with dest-authority): `upr/` + `dest-authority-id/` + `dest-resource-id + dest-entity-id + dest-version`
+
+3. Receiver (dest-authority/dest-entity) registers a listener for
+    * `up://dest-authority/dest-entity/dest-version/dest-resource.instance`
```
stevenhartley, line 221: Receiver registers a listener for `up://origin-authority/origin-entity/origin-version/origin-resource.instance` . When the message type is UMESSAGE_TYPE_NOTIFICATION, the up-client library matches the source (not the sink) to the listner.

up-l1/zenoh.adoc
```diff
@@ -178,3 +183,91 @@ Encoding::WithSuffix(
...
+
+4. which `UTransport.registerListener` maps to Zenoh key
```
stevenhartley, line 223: see comment above.

stevenhartley (changes requested):
Thank you for this CY, I think this will drastically help explain and I need to put this high level explainations in the UTransport API specification if we agree. 

Publish and notifications listener matching is to the source (client side), but request and response are to the sink.

up-l1/zenoh.adoc
stevenhartley: Michael identified a key issue now on the client side in registerListener() that I'll document in #81

up-l1/zenoh.adoc
```diff
@@ -178,3 +183,91 @@ Encoding::WithSuffix(
...
+    * remote (with origin-authority): `upr/` + `origin-authority-id/` + `origin-resource-id + origin-entity-id + origin-version`
+
+3. Subscriber registers a listener for
+    * `up://origin-authority/origin-entity/origin-version/origin-resource.instance#message`
+
+4. which `UTransport.registerListener` maps to Zenoh key
+    * local (w/o origin-authority): `upl/` + `origin-resource-id + origin-entity-id + origin-version`
+    * remote (with origin-authority): `upr/` + `origin-authority-id/` + `origin-resource-id + origin-entity-id + origin-version`
+
+5. Since this is **Publish URI**, `Zenoh subscriber` will be used to bind the `registerListener`
```
PLeVasseur, line 208: For the section on `registerListener` for Publish, Notification, Request, and Response I think it'd be a good idea to explicitly make note of which UUri we're registering on: destination or origin.

PLeVasseur:
@MelamudMichael -- could you please review this so that the cpp and rust implementations are on the same page?

up-l1/zenoh.adoc
evshary: I agree with you. It's clearer to readers.

PLeVasseur:
IMHO for traceability would be nice to keep separate, but I don't have a very strong opinion =^)

PLeVasseur (approved):
Yup, makes sense. Thanks for updating

evshary:
@PLeVasseur No. As you suggested [here](https://github.com/eclipse-uprotocol/up-spec/pull/92#issuecomment-2025171681), I'm creating another PR for the changes.
It would be good if this PR could be merged first.

stevenhartley (approved):
#103 will be used to fix the other part of the spec w.r.t URI mapping to zenoh keys

PR #94, opened 2024-03-19
Clarification of minimum viable UUri as being one that satisfies Micro Form

PLeVasseur:
From #68, namely from [this clarification](https://github.com/eclipse-uprotocol/up-spec/issues/68#issuecomment-2007523728)

Updated to reflect that the minimum viable UUri that we should accept in UTransport::send() is that which satisfies the Micro Form.

### DIFF
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -105,7 +105,7 @@ The API signature:
 
 * All UMessage attributes metadata *MUST* be preserved during transmission and available to the receiver
 * *MUST* not manipulate the UMessage payload during transmission
-* *MUST* pass UUris which satisfy criteria for xref:../basics/uri.adoc#_micro_uris[Micro Uris]
+* *MUST* pass UUris for which `isMicroForm()` return true
 * *SHOULD* pass Resolved UUris (contains both names and numbers)
 * *SHOULD* use the Micro Uri serialization format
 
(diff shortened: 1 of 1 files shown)


up-l1/README.adoc
```diff
@@ -105,8 +105,9 @@ The API signature:
 
 * All UMessage attributes metadata *MUST* be preserved during transmission and available to the receiver
 * *MUST* not manipulate the UMessage payload during transmission
-* *SHOULD* pass resolved UUris (contains both names and numbers)
-* *SHOULD* use the Miro Uri serialization format
+* *MUST* pass UUris which satisfy criteria for xref:../basics/uri.adoc#_micro_uris[Micro Uris]
+* *SHOULD* pass Resolved UUris (contains both names and numbers)
+* *SHOULD* use the Micro Uri serialization format
```
stevenhartley, line 110: The last bullet (from my original description) doesn't make sense if you have the first item.

up-l1/README.adoc
```diff
@@ -105,8 +105,9 @@ The API signature:
 
 * All UMessage attributes metadata *MUST* be preserved during transmission and available to the receiver
 * *MUST* not manipulate the UMessage payload during transmission
-* *SHOULD* pass resolved UUris (contains both names and numbers)
-* *SHOULD* use the Miro Uri serialization format
+* *MUST* pass UUris which satisfy criteria for xref:../basics/uri.adoc#_micro_uris[Micro Uris]
```
stevenhartley, line 108: isMicroForm() to align with spec.

stevenhartley (changes requested):

up-l1/README.adoc
stevenhartley: It is not explicit and it comes from my original text :-). and I can't think anymore how to better phrase it so I would just remove the original text please.

stevenhartley (approved):

Issue #98, opened 2024-03-26
What is the minimum viable UUri to pass `UTransport::registerListener() / unregisterListener()`?


PLeVasseur:
In #94 we clarified that for `UTransport::send()` the minimum viable UUri contained within the UMessage passed in must be in Micro Form

I'm wondering about `UTransport::registerListener() / unregisterListener()`

Perhaps the same language can be used for `UTransport::registerListener() / unregisterListener()`:
* **MUST** pass UUris for which `isMicroForm()` return true
* **SHOULD** pass Resolved UUris (contains both names and numbers)

WDYT?

Tagging in @stevenhartley & @sophokles73

sophokles73:
sounds good to me

stevenhartley:
Agree as well, @PLeVasseur please push the spec update. Thank you

PLeVasseur:
> Agree as well, @PLeVasseur please push the spec update. Thank you

Sounds good. Opened #101 for this :slightly_smiling_face:

PLeVasseur:
Done in #101, closing!

PR #101, opened 2024-03-27
Clarification of minimum viable UUri as being one that satisfies Micro Form for `UTransport::registerListener()` and `UTransport::unregisterListener`

PLeVasseur:
Updating UTransport::registerListener() and UTransport::unregisterListener() to clarify that they take _at a minimum_ a UUri which satisfies isMicroForm() but SHOULD be passed Resolved UUris.

For issue #98


stevenhartley (approved):
Thanks @PLeVasseur !

Issue #100, opened 2024-03-27
How UUri maps to Zenoh key


evshary:
As we discussed in the weekly meeting, let me put the conclusion here to see whether we're on the same page.
Then I can update the [Zenoh spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/zenoh.adoc#22-uri-mapping) in up-spec later.

1. We don't need to distinguish the local and remote UUri in the Zenoh key. The Zenoh key transformed from UUri is visible globally. No matter local entity or remote entity subscribes to the **same** Zenoh key.
2. If the UUri doesn't include UAuthority, we need to prepend it while transforming UUri into Zenoh key.
    Users should add UAuthority or we can create UAuthority internally in the UPClientZenoh constructor.
![image](https://github.com/eclipse-uprotocol/up-spec/assets/456210/a10d6f3d-3c5d-4392-b3cb-cf89a3320d4a)

If we agree on the conclusion, then Zenoh doesn't need to do any access control to the local UUri on Zenoh router and I'll close the issue.
https://github.com/eclipse-zenoh/zenoh/issues/773

Let me know your thoughts. @Mallets @sophokles73 @stevenhartley @PLeVasseur

Mallets:
A clarification on:
> 1. We don't need to distinguish the local and remote `UUri` in the Zenoh key.

From point of view the Zenoh network always operate on `remote` (i.e. global) `UUri`. If the uProtocol developer uses a local `UUri`, then it's the duty of the `UPClientZenoh` to properly add the `UAuthority` in such a way at Zenoh level only remote UUri are present. I hope my understanding is correct.

evshary:
> From point of view the Zenoh network always operate on remote (i.e. global) UUri. If the uProtocol developer uses a local UUri, then it's the duty of the UPClientZenoh to properly add the UAuthority in such a way at Zenoh level only remote UUri are present. I hope my understanding is correct.

Correct. This is also my understanding.

PLeVasseur:
I think this will work with the concept of the uStreamer so long as:
* The uStreamer can be configured to ignore Publish messages from certain authorities, even though uSubscription is saying there's a valid subscription
   * To allow for the use case of where we wish to allow Zenoh to handle the routing from A->B, instead of the uStreamer
* The uStreamer is configured carefully so that we _never_ add a forwarding rule from A->B so that Request, Response, and Notification message forwarding is handled entirely by the Zenoh Router

[...]

Issue #83, opened 2024-03-14
Should we Bring Back Short Form URI?


stevenhartley:
Older version of [uri.adoc](https://github.com/eclipse-uprotocol/up-spec/blob/e4ad95ff1a7e8d1dfb806f095131eb8172a123a2/basics/uri.adoc) use to also have a URI format called Short Form. Short URI was a string representation of the numbers inside of UUri object, example below:

`up://192.168.1.100/2/1.1/5000`

In the above example. UAuthority was the IP address, uService id is 2, verison of the service is 1.1, and the topic ID (the publish topic) was 5000. 
It was thought that this version would not be needed so it was removed from up-java and the specification **however** we see strong benefits of this URI format for zenoh and MQTT that require strings for topic names. 

This issue is to address if we should bring back short form URIs for zenoh & MQTT (and possibly other transports) in lieu of micro form that is difficult to use unless you're writing raw bytes to a transport "wire".

evshary:
From Zenoh perspective, using the short URI is easier to generate Zenoh key.

* Long URI: Need to deal with escaping illegal character in Zenoh
* Micro URI: Need to transform the binary into hex string

However, I'm just wondering whether this will cause a problem for users if they need to define Long/Short/Micro format while declaring the UURI?

PLeVasseur:
I think this makes a lot of sense for protocols like MQTT and Zenoh.

@evshary -- btw, just wanted to make sure that you're still able to take a UUri containing only a UAuthority and still form a Zenoh KeyExpr then like this: `up://192.168.1.100/**`

You may need to submit a PR upstream to `up-rust` to add functionality to serialize only a UAuthority to long form / string.

evshary:
Hi @PLeVasseur 

> btw, just wanted to make sure that you're still able to take a UUri containing only a UAuthority and still form a Zenoh KeyExpr then like this: up://192.168.1.100/**

I think I still need to do some minor transformations.

* with UAuthority: `upr/192.168.1.100/4/1`	
* w/o UAuthority: `upl/4/1`
* For the special UUri: `upr/192.168.1.100/**`

> You may need to submit a PR upstream to up-rust to add functionality to serialize only a UAuthority to long form / string.

Do you mean short form?

PLeVasseur:
> I think I still need to do some minor transformations.

Makes sense to me.

> Do you mean short form?

[...]

PR #103, opened 2024-04-03
Conform Zenoh spec to 1.5.8

evshary:
The PR is to deal with

* The mapping of UUri and Zenoh Key: https://github.com/eclipse-uprotocol/up-spec/issues/100
* Using short UUri for Zenoh Key: https://github.com/eclipse-uprotocol/up-spec/issues/83

Still need to wait for #92 before merging (I need to add some changes)

### DIFF
--- a/up-l1/zenoh.adoc
+++ b/up-l1/zenoh.adoc
@@ -114,29 +114,19 @@ Encoding::WithSuffix(
 )
 ----
 
+=== UPayload
+
+The data is sent with Zenoh directly without further processing.
+
 === URI Mapping
 
 UUri are transformed into Zenoh key expressions following the format defined below.
-The format of Zenoh key **MUST** be `up/[src authority]/[dst authority]/[entity_id]/[ue_version_major]/[resource_id]`
+The format of Zenoh key **MUST** be 
+`up/[source.authority_name]/[source.ue_id]/[source.ue_version_major]/[source.resource_id]/[sink.authority_name]/[sink.ue_id]/[sink.ue_version_major]/[sink.resource_id]`
 
 * `up/`: Zenoh key **MUST** always start with `up/`, which can be used to identify the uProtocol traffic.
-* `src authority`: Indicates where the message comes from.
-* `dst authority`: Indicates where the message is sent to. Must be `empty` in Publish type communication. `empty` is reserved word.
-* `entity_id`, `ue_version_major` and `resource_id`: Might come from source or destination UUri depending on the communication type. They *MUST* be mapped to the upper-case base16 encoding.
-
-Different type of communication in uProtocol maps to different Zenoh keys, just like the following table.
-
-[%autowidth]
-|===
-| Message Type | Zenoh Key
-
-| Publish | up/src-authority/{}/src-entity/src-version/src-resource
-| Notification | up/src-authority/dst-authority/dst-entity/dst-version/dst-resource
-| Request | up/src-authority/dst-authority/dst-entity/dst-version/dst-resource
-| Response | Don't need it since relying on Zenoh queryable mechanism
-|===
-
-Note that since authority is case-insensitvie and allows some special characters defined in link:https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2[RFC3986, Section 3.2.2]. Zenoh needs to use `{}` to escape these characters and transform all the alphbets to lowercase.
+* `authority_name`: Since authority is case-insensitive and allows some special characters defined in link:https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2[RFC3986, Section 3.2.2]. The implementation **MUST** transform all the alphabets to lowercase and use `{}` to escape special characters.
+* `entity_id`, `ue_version_major` and `resource_id`: They *MUST* be mapped to the upper-case base16 encoding.
 
 [%autowidth]
 |===
@@ -150,97 +140,13 @@ Take some examples:
 
 [%autowidth]
 |===
-| Use Case | Source UUri | Sink UUri | Zenoh Key
+| Use Case | Source | Sink | Zenoh Key
 
-| Publish | up://192.168.1.100/1002/3/800A | - | up/192.168.1.100/{}/1002/3/800A
-| Notification | up://192.168.1.99/... | up://192.168.1.100/1002/3/800A | up/192.168.1.99/192.168.1.100/1002/3/800A
-| Request | up://MYHOST1/... | up://myhost$/1002/3/B | up/myhost1/myhost{dollar}/1002/3/B
-| uStreamer | up://+++*+++/... | up://192.168.1.100/0000FFFF/FF/FFFF | up/+++*+++/192.168.1.100/+++*+++/+++*+++/+++*+++
+| Send Publish | up://192.168.1.100/10AB/3/80CD | - | up/192.168.1.100/10AB/3/80CD/{}/{}/{}/{}
+| Subscribe messages | up://192.168.1.100/10AB/3/80CD | - | up/192.168.1.100/10AB/3/80CD/{}/{}/{}/{}
+| Send Notification | up://192.168.1.100/10AB/3/80CD | up://192.168.1.101/20EF/4/80GH | up/192.168.1.100/10AB/3/80CD/192.168.1.101/20EF/4/80GH
+| Receive all Notifications | up://+++*+++/FFFF/FF/FFFF | up://192.168.1.101/20EF/4/80GH | up/+++*+++/+++*+++/+++*+++/+++*+++/192.168.1.101/20EF/4/80GH
+| Send Request | up://MY-HOST1/10AB/3/0 | up://my-host$/20EF/4/B | up/my-host1/10AB/3/0/my-host{dollar}/20EF/4/B
+| Receive all Requests | up://+++*+++/FFFF/FF/FFFF | up://my-host$/20EF/4/B | up/+++*+++/+++*+++/+++*+++/+++*+++/my-host{dollar}/20EF/4/B
+| Receive all messages to a device | up://+++*+++/FFFF/FF/FFFF | up://192.168.1.101/FFFF/FF/FFFF | up/+++*+++/+++*+++/+++*+++/+++*+++/192.168.1.101/+++*+++/+++*+++/+++*+++
 |===
-
-=== UPayload
-
-The data is sent with Zenoh directly without further processing.
-
-=== Work Flow
-
-The section provides the details how it works in different message types.
-
-==== Transfer Publish message
-
-1. A uEntity (`origin-authority/origin-entity`) creates a message with
-    * source: `up://origin-authority/origin-entity/origin-version/origin-resource`
-    * sink: `-`
-
-2. `UTransport.send` maps source (**origin UUri**) to Zenoh key
-    * `up/` + `origin-authority/{}/origin-entity/origin-version/origin-resource`
-
-3. Subscriber registers a listener
-
-    * source: `up://origin-authority/origin-entity/origin-version/origin-resource`
-    * sink: `-`
-
-4. which `UTransport.registerListener` maps to Zenoh key
-    * `up/` + `origin-authority/{}/origin-entity/origin-version/origin-resource`
-
-5. Since this is **Publish URI**, `Zenoh subscriber` will be used to bind the `registerListener`
-
-==== Transfer Notification message
-
-1. A uEntity (`origin-authority/origin-entity`) creates a message with
-    * source: `up://origin-authority/origin-entity/origin-version/origin-resource.instance`
-    * sink: `up://dest-authority/dest-entity/dest-version/dest-resource.instance`
-
-2. `UTransport.send` maps sink (**destination UUri**) to Zenoh key
-    * `up/` + `origin-authority/dest-authority/dest-entity/dest-version/dest-resource`
-
-3. Receiver (dest-authority/dest-entity) registers a listener on **destination UUri**
-    * source: `-`
-    * sink: `up://dest-authority/dest-entity/dest-version/dest-resource.instance`
-
-4. which `UTransport.registerListener` maps to Zenoh key
-    * `up/` + `+++*+++/dest-authority/dest-entity/dest-version/dest-resource`
-
-5. Since this is **Notification URI**, `Zenoh subscriber` will be used to bind the `registerListener`
-
-==== Transfer Request message
-
-1. Service consumer (`consumer-authority/consumer-entity`) creates a message with
-    * source: `up://consumer-authority/consumer-entity/consumer-version/rpc.response`
-    * sink: `up://provider-authority/provider-entity/provider-version/rpc.methodname`
-
-2. `UTransport.send` maps sink (**destination UUri**) to Zenoh key
-    * `up/` + `consumer-authority/provider-authority/provider-entity/provider-version/provider-resource`
-
-3. Service provider (provider-authority/provider-entity) registers a listener on **destination UUri**
-    * source: `-`
-    * sink: `up://provider-authority/provider-entity/provider-version/rpc.methodname`
-
-4. which `UTransport.registerListener` maps to Zenoh key
-    * `up/` + `+++*+++/provider-authority/provider-entity/provider-version/provider-resource`
-
-5. Since this is **Request URI**, `Zenoh queryable` will be used to bind the `registerListener`
-
-==== Transfer Response message
-
-The way to transfer Response message is different from others, because Zenoh can reply the message with its own mechanism and doesn't need to map the UUri to another Zenoh key.
-
-1. Service provider receives a Request message with `Zenoh queryable`, and it can store this query with `UAttributes.id`.
-
-2. `UTransport.send` will retrieve the specific query by comparing the `UAttributes.reqid`, and then send back the message with that query.
-
-3. There are two ways for service consumer to receive the response message:
-    * Using `invoke_method`:
-        - After sending query with `Zenoh get`, we can receive the reply by using `recv()`.
-    * Using `register_listener`:
-        - Service consumer (consumer-authority/consumer-entity) registers a listener on **Response UUri** (`up://consumer-authority/consumer-entity/consumer-version/rpc.response`)
-        - While the request is sent by `UTransport.send`, the listener will be registered in `Zenoh get` and triggered if receiving the response message.
-
-==== Message routing
-
-1. A streamer that is interested in all incoming messages for dest-authority registers a listener for
-    * source: `-`
-    * sink: `up://dest-authority/0000FFFF/FF/FFFF`
-
-2. which `UTransport.registerListener` maps to Zenoh key
-    * `up/` + `+++*+++/dest-authority-id/+++*+++/+++*+++/+++*+++`
--- a/up-l1/zenoh.adoc
+++ b/up-l1/zenoh.adoc
@@ -30,7 +30,17 @@ For more information, please visit https://projects.eclipse.org/projects/iot.zen
 
 === Zenoh Version
 
-We **MUST** use Zenoh version `0.10.1-rc` to ensure the interoperability in different language bindings.
+We **MUST** use Zenoh version `0.11.0-rc.3` to ensure the interoperability in different language bindings.
+
+=== UPClientZenoh initialization
+
+While initializing up-client library, we **MUST** include the following information.
+
+* Zenoh Config: If user wants to configure how Zenoh works, they can adjust the configuration.
+* UAuthority: UAuthority is used in two scenarios:
+    - UAuthority is used to generate source authority field in Zenoh key.
+    - While calling `invoke_method`, we need UAuthority and UEntity to generate the response UUri.
+* UEntity: While calling `invoke_method`, we need UAuthority and UEntity to generate the response UUri.
 
 === UAttribute Mapping
 
@@ -86,83 +96,15 @@ Different message types **MUST** use different Zenoh API.
 | CS6 | REAL_TIME
 |===
 
-=== URI Mapping
+==== TTL
 
-UUri **MUST** be able to be serialized into link:../basics/uri.adoc#42-micro-uris[micro form].
-Then we can transform the micro-form UUri, which is `u8` array, into string in lowercase hexadecimal representation.
+While sending Request messages, TTL **MUST** be mapped to the timeout configuration in Zenoh query.
 
-Zenoh key for uProtocol consists of three parts and they are separated by `/`.
-Zenoh key **MUST** start with `upl/` or `upr/`.
-While the local UUri maps to `upl/`, the remote UUri maps to `upr/`.
-If UAuthority is defined in UUri, it **MUST** be put in the second part.
-Then the rest of the UUri **MUST** be put in the last part.
+==== Payload Format
 
-To be more specific:
-
-* The local UUri without UAuthority: `upl/[hexadecimal representation of the UUri]`
-* The remote UUri with UAuthority: `upr/[hexadecimal representation of the UAuthority]/[hexadecimal representation of the rest of the UUri]`
-
-Note that there is a special case. If the UUri only contains UAuthority with empty UEntity and UResource,
-it's defined as a special form of Zenoh key `upr/[byte hexadecimal of UAuthority]/**`.
-It can listen to all the messages sending to a certain UAuthority.
-
-Some examples:
-
-* The local UUri without UAuthority:
-
-[source]
-----
-UUri {
-  authority: empty,
-  entity: UEntity {
-    name: "body.access",
-    version_major: 1,
-    id: 1234,
-  }
-  resource: UResource {
-    name: "door",
-    instance: "front_left",
-    message: "Door",
-    id: 5678,
-  }
-}
-----
-
-It's a local UUri, so it starts with `upl`.
-Also, we leave the second part blank since it doesn't have UAuthority.
-Only ID in UEntity and UResource matters while serializing into link:../basics/uri.adoc#421-local-micro-uri[micro-form Uri].
-Then transform it into lowercase hexadecimal representation.
-The Zenoh key is `upl/0100162e04d20100`.
-
-* The remote UUri with UAuthority
-
-[source]
-----
-UUri {
-  authority: UAuthority {
-    name: UAuthName,
-    number: Id({01, 02, 03, 10, 11, 12})
-  },
-  entity: empty,
-  resource: empty,
-}
-----
-
-It's a remote UUri, so it starts with `upr`.
-The UAuthority is serialized into mirco form.
-Given how a UAuthority containing an ID is serialized into micro form as outlined in link:../basics/uri.adoc#422-remote-micro-uris[the spec],
-it is length + serialized UUri in lowercase hexadecimal representation, which is `06` + `0102030a0b0c`.
-The UEntity and UResource are all blank, so it is the special form UUri.
-It means that the third part is `+++**+++`.
-Therefore, the Zenoh key is `upr/060102030a0b0c/+++**+++`.
-
-=== UPayload
-
-Although UPayload can be encoded into protobuf, it saves a lot of time without encoding.
-UPayload consists of three parts: length, format, data.
-
-* Length only works with shared memory, and we ignore it temporarily.
-* Format **SHOULD** be sent with Zenoh encoding. Zenoh supports `Encoding::WithSuffix`. The encoding here is `KnownEncoding::AppCustom` and the UPayloadFormat (int32) is transformed into the string.
+Payload Format **MUST** be sent with Zenoh encoding.
+Zenoh supports `Encoding::WithSuffix`.
+The encoding here is `KnownEncoding::AppCustom` and the UPayloadFormat (int32) is transformed into the string.
 
 [source, rust]
 ----
@@ -172,7 +114,53 @@ Encoding::WithSuffix(
 )
 ----
 
-* Zenoh takes data as a buffer simply and sends it directly. Ignore the shared memory (reference) temporarily.
+=== URI Mapping
+
+UUri are transformed into Zenoh key expressions following the format defined below.
+The format of Zenoh key **MUST** be `up/[src authority]/[dst authority]/[entity_id]/[ue_version_major]/[resource_id]`
+
+* `up/`: Zenoh key **MUST** always start with `up/`, which can be used to identify the uProtocol traffic.
+* `src authority`: Indicates where the message comes from.
+* `dst authority`: Indicates where the message is sent to. Must be `empty` in Publish type communication. `empty` is reserved word.
+* `entity_id`, `ue_version_major` and `resource_id`: Might come from source or destination UUri depending on the communication type. They *MUST* be mapped to the upper-case base16 encoding.
+
+Different type of communication in uProtocol maps to different Zenoh keys, just like the following table.
+
+[%autowidth]
+|===
+| Message Type | Zenoh Key
+
+| Publish | up/src-authority/{}/src-entity/src-version/src-resource
+| Notification | up/src-authority/dst-authority/dst-entity/dst-version/dst-resource
+| Request | up/src-authority/dst-authority/dst-entity/dst-version/dst-resource
+| Response | Don't need it since relying on Zenoh queryable mechanism
+|===
+
+Note that since authority is case-insensitvie and allows some special characters defined in link:https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2[RFC3986, Section 3.2.2]. Zenoh needs to use `{}` to escape these characters and transform all the alphbets to lowercase.
+
+[%autowidth]
+|===
+| Special characters in authority | Zenoh Key
+
+| empty | `{}`
+| `$` | `{dollar}`
+|===
+
+Take some examples:
+
+[%autowidth]
+|===
+| Use Case | Source UUri | Sink UUri | Zenoh Key
+
+| Publish | up://192.168.1.100/1002/3/800A | - | up/192.168.1.100/{}/1002/3/800A
+| Notification | up://192.168.1.99/... | up://192.168.1.100/1002/3/800A | up/192.168.1.99/192.168.1.100/1002/3/800A
+| Request | up://MYHOST1/... | up://myhost$/1002/3/B | up/myhost1/myhost{dollar}/1002/3/B
+| uStreamer | up://+++*+++/... | up://192.168.1.100/0000FFFF/FF/FFFF | up/+++*+++/192.168.1.100/+++*+++/+++*+++/+++*+++
+|===
+
+=== UPayload
+
+The data is sent with Zenoh directly without further processing.
 
 === Work Flow
 
@@ -181,19 +169,19 @@ The section provides the details how it works in different message types.
 ==== Transfer Publish message
 
 1. A uEntity (`origin-authority/origin-entity`) creates a message with
-    * source: `up://origin-authority/origin-entity/origin-version/origin-resource.instance#message`
+    * source: `up://origin-authority/origin-entity/origin-version/origin-resource`
     * sink: `-`
 
 2. `UTransport.send` maps source (**origin UUri**) to Zenoh key
-    * local (w/o origin-authority): `upl/` + `origin-resource-id + origin-entity-id + origin-version`
-    * remote (with origin-authority): `upr/` + `origin-authority-id/` + `origin-resource-id + origin-entity-id + origin-version`
+    * `up/` + `origin-authority/{}/origin-entity/origin-version/origin-resource`
 
-3. Subscriber registers a listener on **origin UUri**
-    * `up://origin-authority/origin-entity/origin-version/origin-resource.instance#message`
+3. Subscriber registers a listener
+
+    * source: `up://origin-authority/origin-entity/origin-version/origin-resource`
+    * sink: `-`
 
 4. which `UTransport.registerListener` maps to Zenoh key
-    * local (w/o origin-authority): `upl/` + `origin-resource-id + origin-entity-id + origin-version`
-    * remote (with origin-authority): `upr/` + `origin-authority-id/` + `origin-resource-id + origin-entity-id + origin-version`
+    * `up/` + `origin-authority/{}/origin-entity/origin-version/origin-resource`
 
 5. Since this is **Publish URI**, `Zenoh subscriber` will be used to bind the `registerListener`
 
@@ -204,15 +192,14 @@ The section provides the details how it works in different message types.
     * sink: `up://dest-authority/dest-entity/dest-version/dest-resource.instance`
 
 2. `UTransport.send` maps sink (**destination UUri**) to Zenoh key
-    * local (w/o dest-authority): `upl/` + `dest-resource-id + dest-entity-id + dest-version`
-    * remote (with dest-authority): `upr/` + `dest-authority-id/` + `dest-resource-id + dest-entity-id + dest-version`
+    * `up/` + `origin-authority/dest-authority/dest-entity/dest-version/dest-resource`
 
 3. Receiver (dest-authority/dest-entity) registers a listener on **destination UUri**
-    * `up://dest-authority/dest-entity/dest-version/dest-resource.instance`
+    * source: `-`
+    * sink: `up://dest-authority/dest-entity/dest-version/dest-resource.instance`
 
 4. which `UTransport.registerListener` maps to Zenoh key
-    * local (w/o dest-authority): `upl/` + `dest-resource-id + dest-entity-id + dest-version`
-    * remote (with dest-authority): `upr/` + `dest-authority-id/` + `dest-resource-id + dest-entity-id + dest-version`
+    * `up/` + `+++*+++/dest-authority/dest-entity/dest-version/dest-resource`
 
 5. Since this is **Notification URI**, `Zenoh subscriber` will be used to bind the `registerListener`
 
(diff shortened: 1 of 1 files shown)


up-l1/zenoh.adoc
```diff
@@ -93,22 +93,21 @@ Different message types **MUST** use different Zenoh API.
...
 Zenoh key for uProtocol consists of three parts and they are separated by `/`.
-Zenoh key **MUST** start with `upl/` or `upr/`.
-While the local UUri maps to `upl/`, the remote UUri maps to `upr/`.
-If UAuthority is defined in UUri, it **MUST** be put in the second part.
-Then the rest of the UUri **MUST** be put in the last part.
 
-To be more specific:
+* `up/`: Zenoh key **MUST** always start with `up/` and the scheme (`up://`) in the short UUri is always removed.
+* UAuthority: UAuthority is not mandatory, and it can be ignored in local UUri. However, it **MUST** always exist in Zenoh key.
+* `UE_ID`, `UE_VER_MAJOR`, and `URESOURCE_ID`: We reserve them in Zenoh Key.
```
PLeVasseur, line 103: Could you help me understand what reserve means in this case?

up-l1/zenoh.adoc
```diff
@@ -93,22 +93,21 @@ Different message types **MUST** use different Zenoh API.
...
 
 Zenoh key for uProtocol consists of three parts and they are separated by `/`.
-Zenoh key **MUST** start with `upl/` or `upr/`.
-While the local UUri maps to `upl/`, the remote UUri maps to `upr/`.
-If UAuthority is defined in UUri, it **MUST** be put in the second part.
-Then the rest of the UUri **MUST** be put in the last part.
 
-To be more specific:
+* `up/`: Zenoh key **MUST** always start with `up/` and the scheme (`up://`) in the short UUri is always removed.
+* UAuthority: UAuthority is not mandatory, and it can be ignored in local UUri. However, it **MUST** always exist in Zenoh key.
```
PLeVasseur, line 102: Could you expand on this a bit? I'm having a hard time parsing it out.

Where is UAuthority not mandatory? Within the UURIs in a UMessage when calling `UTransport::send()`? When calling `UTransport::register_listener()`?

And if it's not mandatory, how do we ensure it always exists in the Zenoh key? The constructor which takes a UAuthority?

up-l1/zenoh.adoc
evshary: I changed another way to describe it and hope it's clearer now :smile:

up-l1/zenoh.adoc
evshary: I think I didn't describe it well. How about now?

evshary:
I've also updated the workflow and added the requirement of arguments in constructors.
Let me implement it in up-client-zenoh-rust first before merging it.

up-l1/zenoh.adoc
```diff
@@ -98,22 +108,21 @@ Different message types **MUST** use different Zenoh API.
...
+Then we can transform the short-form UUri into string for Zenoh key.
 
 Zenoh key for uProtocol consists of three parts and they are separated by `/`.
-Zenoh key **MUST** start with `upl/` or `upr/`.
-While the local UUri maps to `upl/`, the remote UUri maps to `upr/`.
-If UAuthority is defined in UUri, it **MUST** be put in the second part.
-Then the rest of the UUri **MUST** be put in the last part.
 
-To be more specific:
+* `up/`: Zenoh key **MUST** always start with `up/` and the scheme (`up://`) in the short UUri is always removed.
```
gregmedd, line 116: Out of curiosity, is there a reason not to just keep the scheme from the short URI? It's only two extra characters, but removes a whole extra string operation that is unique to Zenoh and applied to a string that is otherwise already formatted correctly. Does Zenoh treat `:` or `/` as special characters?

If there's no specific need for this, we could reduce the amount of code that has to be written, tested, and maintained by keeping the scheme from the short URI.

up-l1/zenoh.adoc
```diff
@@ -138,33 +147,28 @@ UUri {
 }
 ----
 
-It's a local UUri, so it starts with `upl`.
-Also, we leave the second part blank since it doesn't have UAuthority.
-Only ID in UEntity and UResource matters while serializing into link:../basics/uri.adoc#421-local-micro-uri[micro-form Uri].
-Then transform it into lowercase hexadecimal representation.
-The Zenoh key is `upl/0100162e04d20100`.
+It's a local URI, and the short format of the UUri should be `/1/1234/5678` in this case.
+Let's assume the UAuthority is `192.168.1.100`.
```
gregmedd, line 151: I'm unclear how we can assume that when we have just said UAuthority is not set because it's a local URI. This should probably be clarified.

up-l1/zenoh.adoc
```diff
@@ -98,22 +108,21 @@ Different message types **MUST** use different Zenoh API.
 
 === URI Mapping
 
-UUri **MUST** be able to be serialized into link:../basics/uri.adoc#42-micro-uris[micro form].
-Then we can transform the micro-form UUri, which is `u8` array, into string in lowercase hexadecimal representation.
+UUri **MUST** be able to be serialized into link:../basics/uri.adoc#short-uris[short form].
+Then we can transform the short-form UUri into string for Zenoh key.
```
gregmedd, line 112: Two questions:
- Does zenoh comply with the short form URI requirement that URIs are case insensitive? This matters for ID and IPv6 authorities - if zenoh is not case insensitive, then we need to add a requirement that case is normalized (e.g. lower-case the whole key)
- How are IP addresses normalized? Particularly with IPv6 (but also to a lesser extent IPv4), there are many ways to write the same address. If zenoh keys are a literal match, then we should probably define what the normalized address format is.

gregmedd (changes requested):

up-l1/zenoh.adoc
evshary: You're right. We can't have `//` in Zenoh key.

up-l1/zenoh.adoc
```diff
@@ -98,22 +108,21 @@ Different message types **MUST** use different Zenoh API.
...
 
 Zenoh key for uProtocol consists of three parts and they are separated by `/`.
-Zenoh key **MUST** start with `upl/` or `upr/`.
-While the local UUri maps to `upl/`, the remote UUri maps to `upr/`.
-If UAuthority is defined in UUri, it **MUST** be put in the second part.
-Then the rest of the UUri **MUST** be put in the last part.
 
-To be more specific:
+* `up/`: Zenoh key **MUST** always start with `up/` and the scheme (`up://`) in the short UUri is always removed.
+* UAuthority: Although UAuthority does not exist in local UUri, it **MUST** be used in Zenoh key. We can use the UAuthority which is initialized in the constructor in this case.
```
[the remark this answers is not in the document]
evshary: We need to define UAuthority and UEntity while creating UPClientZenoh.
It's necessary to fill the source address in the UAttribute when we call `invoke_method`
https://github.com/eclipse-uprotocol/up-spec/pull/103/files#diff-540fb00cfbe38ece62de18b3a95419f71516df5fb08cc89402bf33235411ae7fR50

up-l1/zenoh.adoc
evshary: Maybe I should add more description. UAuthority is defined when creating the UPClientZenoh.

up-l1/zenoh.adoc
evshary: We use the definition of short-form in up-spec
https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#51-remote-uris
However, you raise a good point. I'm not sure whether short form URI should take care of case sensitive or not.
@stevenhartley Do you know that?

evshary:
@gregmedd Thanks for your review. Feel free to advise me on how to make the spec clearer. Sometimes I might ignore some premise while writing down the spec.

up-l1/zenoh.adoc
stevenhartley: The case sensitivity of URIs are covered in the URI spec:

```
Schema and authority MUST be case-insensitive per RFC3986
Path, query, and message MUST be case-sensitive
```

If Zenoh key expressions cannot handle the case differences then I recommend we normalize it and define that in the zenoh.adoc spec.

evshary: I thought the definition here was only for long-format before.
Ok, then we should normalize it to lowercase in Zenoh key.

up-l1/zenoh.adoc
evshary: Done.

up-l1/zenoh.adoc
gregmedd: @evshary

I've looked at the change, and I'm not sure that it helps clarify the needs of normalization. The big concern for me is how IP addresses are normalized - for example, there are many ways to write an IPv6 address that all represent the same address but have different strings. It's not just whether the characters are upper case or lower case. Do we keep or drop leading zeros (both IPv6 and IPv4)? Always compact groups of zeros with `::` or always expand `::`? Do we convert the dual-format addresses (e.g. `2001:db8::f00f:192.168.1.4` or `::1.1.1.1`) to native IPv6 formats?

One option would be to parse the IP address string then convert it back to a string using the APIs provided by Linux, and that behavior becomes spec. Another would be to look at RFCs to see if there is a spec for normalization of addresses.

The second concern I have with the change is that the new language on converting to lower case is unclear. It sounds like the entire key would be converted to lower case, but _only the schema and authority_ should be case insensitive (and lower cased in the key), if I'm understanding what @stevenhartley said earlier..

stevenhartley: @gregmedd I think the best way forward is to explicitly state that the IP address must follow [RFC5952](https://www.rfc-editor.org/rfc/rfc5952). I added this to https://github.com/eclipse-uprotocol/up-spec/issues/115

evshary: @gregmedd
For the first one, I think it will not be a problem if the IP address always follows RFC5952 in up-spec.
For the second one, yes, I should limit the transformation only to the schema and authority.

evshary: Done.

up-l1/zenoh.adoc
```diff
@@ -98,22 +108,24 @@ Different message types **MUST** use different Zenoh API.
...
+Then we can transform the short-form UUri into string for Zenoh key.
 
 Zenoh key for uProtocol consists of three parts and they are separated by `/`.
-Zenoh key **MUST** start with `upl/` or `upr/`.
-While the local UUri maps to `upl/`, the remote UUri maps to `upr/`.
-If UAuthority is defined in UUri, it **MUST** be put in the second part.
-Then the rest of the UUri **MUST** be put in the last part.
 
-To be more specific:
+* `up/`: Zenoh key **MUST** always start with `up/` and the scheme (`up://`) in the short UUri is always removed.
```
stevenhartley, line 116: Serialize to short and long with Java doesn't add the scheme, does up-rust add the scheme? Perhaps we should remove it as it was optional anyways.

up-l1/zenoh.adoc
```diff
@@ -98,22 +108,24 @@ Different message types **MUST** use different Zenoh API.
...
+* UAuthority: Although UAuthority does not exist in local UUri, it **MUST** be used in Zenoh key. We can use the UAuthority which is initialized in the constructor in this case.
+* `UE_ID`, `UE_VER_MAJOR`, and `URESOURCE_ID`: These should be kept in Zenoh Key and also separated by `/`.
+
+To be more specific, the format of Zenoh Key is:
+
+`up/[UAuthoriy]/[UE_ID]/[UE_VER_MAJOR]/[URESOURCE_ID]`
```
stevenhartley, line 122: This is 1:1 mapping of short form URI now so I would just state that you use the short form URI for zenoh key expressions.

up-l1/zenoh.adoc
```diff
@@ -98,22 +108,24 @@ Different message types **MUST** use different Zenoh API.
...
 
-* The local UUri without UAuthority: `upl/[hexadecimal representation of the UUri]`
-* The remote UUri with UAuthority: `upr/[hexadecimal representation of the UAuthority]/[hexadecimal representation of the rest of the UUri]`
+The UAuthority of short-form UUri is case-insensitive, but Zenoh key is case-sensitive.
+We **MUST** transform all the alphabets of the UAuthority into lowercase.
```
stevenhartley, line 125: hum.... perhaps the serializer should do this in lieu of inside of the transport and we make short form URI case sensitive, what say you @sophokles73 & @AnotherDaniel ?

up-l1/zenoh.adoc
```diff
@@ -138,33 +150,28 @@ UUri {
...
 
 * The remote UUri with UAuthority
 
 [source]
 ----
 UUri {
   authority: UAuthority {
-    name: UAuthName,
-    number: Id({01, 02, 03, 10, 11, 12})
+    name: "UAuthName",
```
stevenhartley, line 163: Please see https://github.com/eclipse-uprotocol/up-spec/pull/121, we are dropping id and ip from UAuthority.

up-l1/zenoh.adoc
evshary: You're right. No scheme while serializing to short and long in up-rust. I'll remove this description.

up-l1/zenoh.adoc
evshary: OK, I'll update it after the PR is merged.

up-l1/zenoh.adoc
```diff
@@ -40,7 +40,17 @@ For more information, please visit https://projects.eclipse.org/projects/iot.zen
 
 === Zenoh Version
 
-We **MUST** use Zenoh version `0.10.1-rc` to ensure the interoperability in different language bindings.
+We **MUST** use Zenoh version `0.11.0-rc` to ensure the interoperability in different language bindings.
```
evshary, line 43: @gregmedd Just want to raise it up that we'll use 0.11.0-rc in up-spec 1.5.8. I think you're using it now.

PLeVasseur:
Hey @evshary -- today in the MQTT meeting when discussing @sophokles73's proposal for MQTT, I thought this could also serve a nice purpose for Zenoh as well.

The idea is to include both the source and the destination UAuthorities s.t. we can easily determine when a message is a publish vs notification message and bring the message through.

I think we'd also modify s.t. when we say we want to "subscribe to all", what we really mean is we'd like to subscribe to all Request, Response, and Notification messages, and none of Publish messages.

Publish messages will be handled on a topic-by-topic basis based on what uSubscription tells us to forward, so we do not have a need to get "all" Publish messages.

Essentially, I think that if we modify the Zenoh KeyExpr from:
`up://origin-authority/origin-entity/origin-version/origin-resource.instance`
to include the destination authority as well:
`up://origin-authority/dest-authority/origin-entity/origin-version/origin-resource.instance`
then I think we can disambiguate.

So based on the issue I raised in [this PR](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/31), I have a proposal:

## Setup

A uEntity (origin-authority/origin-entity) creates a message with
* source: `up://origin-authority/origin-entity/origin-version/origin-resource.instance#message`
* sink: -

## Subscribing uE

Registers a listener on origin UUri
* `up://origin-authority/origin-entity/origin-version/origin-resource.instance#message`

which `UTransport.registerListener` maps to Zenoh key
* `up/` + `origin-authority-id/` + `empty/` + `origin-entity-id/origin-version/origin-resource-id`

The MQTT proposal would have the destination as `/`, but I understand that `//` is not valid in a Zenoh KeyExpr. We'd need `empty` to be something reasonable and consistent.

## Subscribing uStreamer

Registers a listener on special origin-authority UUri to catch out-going messages coming from Zenoh, intended for SOME/IP
* `up://origin-authority/*/*/*`

which `UTransport.registerListener` maps to Zenoh key
* `up/` + `*/` + `origin-authority-id/` + `*/*/*`

## uStreamer receives messages and forwards

uStreamer catches the following message on SOME/IP UTransport:
* `up://origin-authority/origin-entity/origin-version/origin-resource.instance#message`

uStreamer then uses `UTransport.send` which maps to Zenoh Key:
* `up/` + `origin-authority-id/` + `empty/` + `origin-entity-id/origin-version/origin-resource-id`

which would now _not_ match for the uStreamer's `registerListener()` because we have discriminated between Publish vs Notification messages via the two slots:
* first the source authority
* second the sink authority

sophokles73:
> Subscribing uStreamer
> Registers a listener on special origin-authority UUri to catch out-going messages coming from > Zenoh, intended for SOME/IP
>
> `up://origin-authority/*/*/*`
> which UTransport.registerListener maps to Zenoh key
>
> `up/` + `*/` + `origin-authority-id/` + `*/*/*`

If the `origin-authority-id` represents the _source_ of the messages, then I do not see how this Zenoh Key could possibly match any messages that are _intended for SOME/IP_, because the third segment in the key represents the _destination_ (sink) of the message, not the source.

I guess it would help if, in your example, you would use terms other than `origin-authority` but instead clearly distinguish between the non-SOME/IP authority vs the SOME/IP authority that you want to route between ...

evshary:
@PLeVasseur I guess I missed some context. I still can't understand why we need to have both the source and sink authority in the Zenoh key. If we don't use this way to discriminate between Publish and Notification, will we face any problems in uStreamer?

Another question is that if we use `up://origin-authority/dest-authority/origin-entity/origin-version/origin-resource.instance` in Notification, it means that the one receiving notifications can only receive data from `origin-authority` in this Zenoh key. If I want to receive notifications that possibly come from A, B, or C, I need to subscribe to three Zenoh keys separately.

* `up://A/dest-authority/origin-entity/origin-version/origin-resource.instance`
* `up://B/dest-authority/origin-entity/origin-version/origin-resource.instance`
* `up://C/dest-authority/origin-entity/origin-version/origin-resource.instance`

Or maybe I mistake something?

sophokles73:
> Another question is that if we use `up://origin-authority/dest-authority/origin-entity/origin-version/origin-resource.instance` in Notification,

Not sure if you are talking about a UUri here or the Zenoh key created from a UUri?

sophokles73:
The problem with all of this is that UTransport.register_listener currently does not support indicating if I want to register a listener for Publish or other types of messages. This, however, would be necessary in order for the transport impl to be able to create the proper Zenoh key to match on, i.e. whether to fill the authority into the `source-authority` or the `sink-authority` segment of the key.

We had this discussion weeks ago already but never really came to a conclusion. Maybe it is time to create a corresponding PR in up-rust to add that flexibility to UTransport.

WDYT? @PLeVasseur @evshary

PLeVasseur:
> We had this discussion weeks ago already but never really came to a conclusion. Maybe it is time to create a corresponding PR in up-rust to add that flexibility to UTransport.
> 
> WDYT? @PLeVasseur @evshary

Yeah I like the idea of being action-oriented here in some feature branches:
1. defining an updated `register_listener()` proposal for discriminating between Publish and Notification in `up-rust`
2. reflecting usage of new `register_listener()` API into `up-client-zenoh-rust`
3. seeing how we all feel about it
4. based on our thoughts and feelings, bringing a change to `up-spec`
5. merge `up-rust` & `up-client-zenoh-rust` changes from steps 1. and 2. (with any feedbacks from 3. & 4.)

WDYT @sophokles73, @evshary?

If sounds good, who would like to take 1. and do some API design?

PLeVasseur:
@evshary 

> Maybe I got your point. Is this your problem?
> https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/31#discussion_r1579533930

Yes, exactly.

> I'm thinking of whether using allowed_origin in subscriber can solve the problem or not.
> https://github.com/eclipse-zenoh/zenoh/blob/0.10.1-rc/zenoh/src/subscriber.rs#L474

There may be a Zenoh-specific way to solve this, but AFAIK this will also pose an issue to MQTT5, will it not @sophokles73?

And if so, it feels cleaner to be explicit in the implementation of `register_listener()` within `up-rust` (and more generally in the spec) on how to handle Notification vs Publish.

evshary:
> On the receiving side, a uEntity would subscribe to up://sink-authority/sink-entity/sink-version/sink-resource.instance which the Zenoh uTransport would turn into key
up/*/sink-authority/sink-entity/sink-version/sink-resource.instance so that it can receive Notifications from all origins.

OK, it makes sense to use `*` to listen to all notifications.

> The problem with all of this is that UTransport.register_listener currently does not support indicating if I want to register a listener for Publish or other types of messages. This, however, would be necessary in order for the transport impl to be able to create the proper Zenoh key to match on, i.e. whether to fill the authority into the source-authority or the sink-authority segment of the key.

I always support the idea of introducing type while calling register_listener, just as we discussed [it](https://github.com/eclipse-uprotocol/up-spec/issues/73#issuecomment-1979181712) in the infamous #73 
But, since we are also discussing the change of UUri now, how about introducing the type concept in UUri?
We can discriminate the requests and responses from others (publish & notification) based on the fact that they have special format in UResource.
Now we have:

* request: name=rpc, instance=xxxx
* response: name=rpc, instance=response

We can also have something like

* publish: name=pub, instance=xxxx
* notification: name=notify, instance=xxxx

WDYT, @sophokles73 @PLeVasseur 

> There may be a Zenoh-specific way to solve this, but AFAIK this will also pose an issue to MQTT5, will it not @sophokles73?

I know little about MQTT, but it seems possible.
https://stackoverflow.com/questions/66785306/paho-mqtt-client-how-to-ignore-messages-published-by-myself

up-l1/zenoh.adoc
sophokles73: So far, we have tried to keep very close to the URI definition of RFC3986. As I pointed out to @evshary in https://github.com/eclipse-uprotocol/up-spec/pull/121#issuecomment-2078684079, we should be able to make small restrictions while maintaining overall compatibility. Making authority case-sensitive would potentially start breaking compatibility IMHO. I do not see an issue with letting the transport decide what to do with the authority in order to make it work for them. Restricting the allowed characters in the authority name might still be worthwhile, though, in order to increase general _interoperability_ with transports.

sophokles73:
@evshary we are currently discussing to get rid of the long form completely and use numeric identifiers for the resource only. So, using such a naming convention would not really work. I also do not see, why we would need to introduce the message type into the UUri, given that a message already has the type in the UAttributes. I'd rather extend the UTransport interface to support specifying the message type. I'll give it a shot and create a corresponding PR in up-rust.

sophokles73:
I have created https://github.com/eclipse-uprotocol/up-rust/pull/92 in the hope that this would allow the Zenoh client to use the source and sink in the Zenoh key ...

sophokles73:
> Is there any reserved word for authority name? I mean when we're using Publish, we need to keep sink authority empty and I'm not sure if this should be defined in uri spec or every protocol spec.

Sorry, I do not understand your point. Can you elaborate?

PLeVasseur:
> > Is there any reserved word for authority name? I mean when we're using Publish, we need to keep sink authority empty and I'm not sure if this should be defined in uri spec or every protocol spec.
> 
> Sorry, I do not understand your point. Can you elaborate?

@evshary is asking this because it is disallowed in a Zenoh key expression to have two slashes following one another, i.e. if it were a publish and in say, MQTT `//` could be used to represent that in the topic name, the same is not possible in Zenoh.

So I think the ask is -- is there some reserved or disallowed character in uProtocol that we could use instead? If say `$` were disallowed in a uProtocol, then a publish UUri when converted to a Zenoh key expression may instead contain `/$/` to note that the sink UUri is empty.

sophokles73:
> So I think the ask is -- is there some reserved or disallowed character or word in uProtocol that we could use instead? If say $ were disallowed in a uProtocol UAuthority, then a publish message when converted to a Zenoh key expression may instead contain /$/ to note that the sink UUri is empty.

What would be in the key if the sink were not empty?

PLeVasseur:
Yes, exactly. Thanks for wording it much better 😅

up-l1/zenoh.adoc
```diff
@@ -86,82 +96,107 @@ Different message types **MUST** use different Zenoh API.
...
+TTL **SHOULD** be mapped to the timeout configuration in Zenoh query.
 
-Zenoh key for uProtocol consists of three parts and they are separated by `/`.
-Zenoh key **MUST** start with `upl/` or `upr/`.
-While the local UUri maps to `upl/`, the remote UUri maps to `upr/`.
-If UAuthority is defined in UUri, it **MUST** be put in the second part.
-Then the rest of the UUri **MUST** be put in the last part.
+=== URI Mapping
+
+UUri is transformed into Zenoh key in the implementation.
```
stevenhartley, line 106:
```suggestion
UUris are transformed into Zenoh key expressions following the format defined below.
```

up-l1/zenoh.adoc
```diff
@@ -86,82 +96,107 @@ Different message types **MUST** use different Zenoh API.
...
-Note that there is a special case. If the UUri only contains UAuthority with empty UEntity and UResource,
-it's defined as a special form of Zenoh key `upr/[byte hexadecimal of UAuthority]/**`.
-It can listen to all the messages sending to a certain UAuthority.
+| Publish | up/src-authority/empty/src-entity/src-version/src-resource
+| Notification | up/src-authority/dst-authority/dst-entity/dst-version/dst-resource
+| Reqest | up/src-authority/dst-authority/dst-entity/dst-version/dst-resource
+| Response | Don't need it since relying on Zenoh queryable mechanism
+|===
 
 Some examples:
```
stevenhartley, line 124: @evshary , I would just map the `URI notation` (a.k.a short form) to zenoh key expressions in the above table per message type in lieu of the code below to make it simpler to read

up-l1/zenoh.adoc
```diff
@@ -86,82 +96,107 @@ Different message types **MUST** use different Zenoh API.
...
-It means that the third part is `+++**+++`.
-Therefore, the Zenoh key is `upr/060102030a0b0c/+++**+++`.
+This means we want to register the messages to a certain destination.
+The zenoh_key should be `up/+++*+++/192.168.1.100/+++*+++/+++*+++/+++*+++`
 
 === UPayload
 
-Although UPayload can be encoded into protobuf, it saves a lot of time without encoding.
-UPayload consists of three parts: length, format, data.
+UPayload consists of two parts: data, format.
```
stevenhartley, line 197: Please see recent changes, there is no more UPayload. All we have is payload in UMessage and the UPayloadFormat is in UAttributes

stevenhartley (changes requested):

evshary:
> @evshary is asking this because it is disallowed in a Zenoh key expression to have two slashes following one another, i.e. if it were a publish and in say, MQTT // could be used to represent that in the topic name, the same is not possible in Zenoh.

@PLeVasseur Thank you for the clarification. This is exactly what I'm asking :laughing: 
I didn't know that MQTT doesn't have this limitation before. What I can think of is that I can choose the special character which is not included in sub-delims to achieve the same effect. Or maybe I need to use escape character to indicate empty. I guess it's fine for me to define it in Zenoh spec, but we need to confirm the scope of the legal character in authority name first.

```
sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                 / "*" / "+" / "," / ";" / "="
```

up-l1/zenoh.adoc
```diff
@@ -86,82 +96,107 @@ Different message types **MUST** use different Zenoh API.
 | CS6 | REAL_TIME
 |===
 
-=== URI Mapping
+==== TTL
 
-UUri **MUST** be able to be serialized into link:../basics/uri.adoc#42-micro-uris[micro form].
-Then we can transform the micro-form UUri, which is `u8` array, into string in lowercase hexadecimal representation.
+The TTL only takes effect while sending RPC in the Zenoh implementation.
+TTL **SHOULD** be mapped to the timeout configuration in Zenoh query.
```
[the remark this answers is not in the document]
evshary: Then TTL will not take effect. It'll use Zenoh default timeout.
Do you think we should use **MUST** here?

up-l1/zenoh.adoc
evshary: OK, it's merged now. Then I'll update it.

up-l1/zenoh.adoc
evshary: Agree. Let me update it.

evshary:
These are unavailable characters in Zenoh key
image
I'm thinking of using `[]` to escape the characters.
For example, 

* `[]` means empty
* `[question]` means `?`

etc.

evshary:
Oops, you're right. But IPv6 also uses `:`, for example `::1` means localhost.
Maybe using `{}` or `\` might be better.

sophokles73:
At first sight, I don't see any reason why we shouldn't further restrict the `reg-name` production used for to contain

> Oops, you're right. But IPv6 also uses `:`, for example `::1` means localhost. Maybe using `{}` or `\` might be better.

A corresponding IPv6 address would still look like `[::1]` but I guess `{}` is the best candidate yet as these characters are not allowed anywhere in URIs :+1:

up-l1/zenoh.adoc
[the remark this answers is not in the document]
evshary: Done.

evshary:
> At first sight, I don't see any reason why we shouldn't further restrict the reg-name production used to contain

I'm fine with that. It's also easier to me.

> A corresponding IPv6 address would still look like [::1] but I guess {} is the best candidate yet as these characters are not allowed anywhere in URIs 👍

Great! I'll use `{}` to indicate empty.

Here is the initial proposal on how to escape special characters. I'll update it if we decide to restrict the reg-name.
https://github.com/eclipse-uprotocol/up-spec/pull/103/files#diff-540fb00cfbe38ece62de18b3a95419f71516df5fb08cc89402bf33235411ae7fR140

up-l1/zenoh.adoc
```diff
@@ -86,93 +96,72 @@ Different message types **MUST** use different Zenoh API.
...
-It's a remote UUri, so it starts with `upr`.
-The UAuthority is serialized into mirco form.
-Given how a UAuthority containing an ID is serialized into micro form as outlined in link:../basics/uri.adoc#422-remote-micro-uris[the spec],
-it is length + serialized UUri in lowercase hexadecimal representation, which is `06` + `0102030a0b0c`.
-The UEntity and UResource are all blank, so it is the special form UUri.
-It means that the third part is `+++**+++`.
-Therefore, the Zenoh key is `upr/060102030a0b0c/+++**+++`.
+[%autowidth]
+|===
+| Speical Characters in authority | Zenoh Key
```
sophokles73, line 144: Special characters ...

up-l1/zenoh.adoc
```diff
@@ -86,93 +96,72 @@ Different message types **MUST** use different Zenoh API.
...
+| Use Case | Source UUri | Sink UUri | Zenoh Key
 
-[source, rust]
-----
-Encoding::WithSuffix(
-    KnownEncoding::AppCustom,
-    payload.format.value().to_string().into(),
-)
-----
+| Publish | up://192.168.1.100/1002/3/800A | - | up/192.168.1.100/{}/10002/3/800A
```
sophokles73, line 156: `up/192.168.1.100/{}/1002/3/800A`

up-l1/zenoh.adoc
```diff
@@ -86,93 +96,72 @@ Different message types **MUST** use different Zenoh API.
...
+| Notification | up://192.168.1.99/... | up://192.168.1.100/1002/3/800A | up/192.168.1.99/192.168.1.100/10002/3/800A
```
sophokles73, line 157: `up/192.168.1.100/{}/1002/3/800A`

up-l1/zenoh.adoc
```diff
@@ -86,93 +96,72 @@ Different message types **MUST** use different Zenoh API.
...
+| Request | up://MYHOST1/... | up://myhost$/1002/3/B | up/myhost1/myhost{dollar}/10002/3/B
```
sophokles73, line 158: `up/myhost1/myhost{dollar}/1002/3/B`

up-l1/zenoh.adoc
```diff
@@ -86,93 +96,72 @@ Different message types **MUST** use different Zenoh API.
...
+| uStreamer | up://+++*+++/... | up://192.168.1.100/FFFF/FF/FFFF | up/+++*+++/192.168.1.100/+++*+++/+++*+++/+++*+++
```
sophokles73, line 159: Where do these `+` signs come from?

up-l1/zenoh.adoc
```diff
@@ -181,19 +170,19 @@ The section provides the details how it works in different message types.
...
 
 1. A uEntity (`origin-authority/origin-entity`) creates a message with
-    * source: `up://origin-authority/origin-entity/origin-version/origin-resource.instance#message`
+    * source: `up://origin-authority/origin-entity/origin-version/origin-resource`
     * sink: `-`
 
 2. `UTransport.send` maps source (**origin UUri**) to Zenoh key
-    * local (w/o origin-authority): `upl/` + `origin-resource-id + origin-entity-id + origin-version`
-    * remote (with origin-authority): `upr/` + `origin-authority-id/` + `origin-resource-id + origin-entity-id + origin-version`
+    * `up/` + `origin-authority/empty/origin-entity/origin-version/origin-resource`
```
sophokles73, line 177: That would then be
`up/` + `origin-authority/{}/origin-entity/origin-version/origin-resource`
I guess

up-l1/zenoh.adoc
```diff
@@ -181,19 +170,19 @@ The section provides the details how it works in different message types.
...
 
-3. Subscriber registers a listener on **origin UUri**
-    * `up://origin-authority/origin-entity/origin-version/origin-resource.instance#message`
+3. Subscriber registers a listener
+
+    * source: `up://origin-authority/empty/origin-entity/origin-version/origin-resource`
```
sophokles73, line 181: source: `up://origin-authority/origin-entity/origin-version/origin-resource`

up-l1/zenoh.adoc
```diff
@@ -181,19 +170,19 @@ The section provides the details how it works in different message types.
...
+    * sink: `-`
 
 4. which `UTransport.registerListener` maps to Zenoh key
-    * local (w/o origin-authority): `upl/` + `origin-resource-id + origin-entity-id + origin-version`
-    * remote (with origin-authority): `upr/` + `origin-authority-id/` + `origin-resource-id + origin-entity-id + origin-version`
+    * `up/` + `origin-authority/empty/origin-entity/origin-version/origin-resource`
```
sophokles73, line 185: `up/` + `origin-authority/{}/origin-entity/origin-version/origin-resource`

sophokles73 (commented):
In general, I wonder if you could simply use standard URL/Percent-Encoding for special characters?

up-l1/zenoh.adoc
evshary: Used to escape asterisk.
https://asciidoc-py.github.io/userguide.html#X77

up-l1/zenoh.adoc
evshary: oops...

up-l1/zenoh.adoc
evshary: You're right. I forgot to update here...

up-l1/zenoh.adoc
```diff
@@ -141,7 +141,7 @@ Note that since authority is case-insensitvie and allows some special characters
 
 [%autowidth]
 |===
-| Speical Characters in authority | Zenoh Key
+| Speical characters in authority | Zenoh Key
```
sophokles73, line 144: I was actually more concerned with the typo in `Speical` ;-)

evshary: oh, I missed that typos :laughing:

up-l1/zenoh.adoc
```diff
@@ -86,93 +96,72 @@ Different message types **MUST** use different Zenoh API.
...
-Then transform it into lowercase hexadecimal representation.
-The Zenoh key is `upl/0100162e04d20100`.
+[%autowidth]
+|===
+| Message Type | Zenoh Key
 
-* The remote UUri with UAuthority
+| Publish | up/src-authority/{}/src-entity/src-version/src-resource
+| Notification | up/src-authority/dst-authority/dst-entity/dst-version/dst-resource
+| Reqest | up/src-authority/dst-authority/dst-entity/dst-version/dst-resource
```
sophokles73, line 136: Request

up-l1/zenoh.adoc
```diff
@@ -86,93 +96,72 @@ Different message types **MUST** use different Zenoh API.
 | CS6 | REAL_TIME
 |===
 
-=== URI Mapping
+==== TTL
 
-UUri **MUST** be able to be serialized into link:../basics/uri.adoc#42-micro-uris[micro form].
-Then we can transform the micro-form UUri, which is `u8` array, into string in lowercase hexadecimal representation.
+The TTL only takes effect while sending RPC in the Zenoh implementation.
+TTL **MUST** be mapped to the timeout configuration in Zenoh query.
```
sophokles73, line 102: what about other messages that have timed out already on arrival? Would you still deliver those to registered listeners?

evshary: Maybe I missed something, but is there any way for us to know when the message is sent? How does the receiver side know the message is timed out only with TTL?

stevenhartley: The UUIDv8 defined in the specifications includes the timestamp from when the request message was generated.

up-l1/zenoh.adoc
```diff
@@ -253,7 +240,8 @@ The way to transfer Response message is different from others, because Zenoh can
 ==== Message routing
 
 1. A streamer that is interested in all incoming messages for dest-authority registers a listener for
-    * `up://dest-authority////`
+    * source: `-`
+    * sink: `up://dest-authority/FFFF/FF/FFFF`
```
PLeVasseur, line 244: @evshary -- I'm honestly not sure what's better from a readability standpoint. I think we could also represent this as:
```
* sink: `up://dest-authority/0000FFFF/FF/FFFF`
```
according to the [spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#611-matching-on-authority), right?

up-l1/zenoh.adoc
```diff
@@ -253,7 +240,8 @@ The way to transfer Response message is different from others, because Zenoh can
 ==== Message routing
 
 1. A streamer that is interested in all incoming messages for dest-authority registers a listener for
-    * `up://dest-authority////`
+    * source: `-`
```
PLeVasseur, line 243: Can't leave the comment high up enough. I'm referring to [this line](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#611-matching-on-authority).

With the separation coming between uP-L1 and uP-L2, I have a question: would the `up-client-foo-rust` libraries be responsible for / be expected to implement uP-L2? Or would they _optionally_ implement uP-L2, but allow for the posibility of plugging in the UTransport implementation they made into e.g. the RpcClient "default" implementation we'd have in `up-rust`?

Tagging @sophokles73 in for this one as well to hear his thoughts.

up-l1/zenoh.adoc
```diff
@@ -204,15 +193,14 @@ The section provides the details how it works in different message types.
...
 
 3. Receiver (dest-authority/dest-entity) registers a listener on **destination UUri**
-    * `up://dest-authority/dest-entity/dest-version/dest-resource.instance`
+    * source: `-`
+    * sink: `up://dest-authority/dest-entity/dest-version/dest-resource.instance`
 
 4. which `UTransport.registerListener` maps to Zenoh key
-    * local (w/o dest-authority): `upl/` + `dest-resource-id + dest-entity-id + dest-version`
-    * remote (with dest-authority): `upr/` + `dest-authority-id/` + `dest-resource-id + dest-entity-id + dest-version`
+    * `up/` + `+++*+++/dest-authority/dest-entity/dest-version/dest-resource`
```
PLeVasseur, line 203: I feel like this has been settled at this point, I just don't recall the conversation...

So for Notification messages, the latter parts of the Zenoh key (entity, version, resource) are pulled from the destination, not the source?

Would we want the ability to discriminate at the `registerListener` level as to which source the Notification is coming from? If that were the case, I suppose we'd have to put the latter parts (entity, version, resource) from the source and modify the `registerListener` example to show us using the `source` field.

PLeVasseur (commented):
Thanks for continuing to pull this together @evshary!

Want to hear your feedback on a few points :slightly_smiling_face:

up-l1/zenoh.adoc
evshary: OK, I missed that.

> what about other messages that have timed out already on arrival? Would you still deliver those to registered listeners? Or send out messages in send() that have already timed out?

Then we should check whether messages time out or not on arrival in the transport implementation. Let me update it.

up-l1/zenoh.adoc
evshary: I agree with you that this is more readable.
The reason why I omit the `0000` is that the example in spec always does so.
Since the entity includes service and instance, it might be better to show the whole hex string in entity and avoid the ambiguity.

up-l1/zenoh.adoc
evshary: > So for Notification messages, the latter parts of the Zenoh key (entity, version, resource) are pulled from the destination, not the source?

Only Publish messages use source as the latter part of Zenoh key. Notification and Request only care about the destination.

> Would we want the ability to discriminate at the registerListener level as to which source the Notification is coming from? If that were the case, I suppose we'd have to put the latter parts (entity, version, resource) from the source and modify the registerListener example to show us using the source field.

I don't quite get your point. Maybe retrieving the source information from uAttributes is enough?

up-l1/zenoh.adoc
evshary: After a second look, it might be better for the designer to give us some direction.
@sophokles73 Do you think we should use `0000FFFF` instead of `FFFF` to match all any instances and any types in ue_id?

BTW, out of curiosity, why do you choose `0000` to stand for any instance?

up-l1/zenoh.adoc
evshary: Hmm... do you think we should drop the expired messages on the transport layer (Zenoh, MQTT..), or just passed to users and leave the decisions to them?

up-l1/zenoh.adoc
sophokles73: > BTW, out of curiosity, why do you choose 0000 to stand for any instance?

In almost all cases, the sender of a message does not care about which service instance its request is being sent to. Using `0x0000` for the 16 MSB in this case will lead to a URI that is four characters shorter than including `0xFFFF`, because we can simply omit the leading zeros.

sophokles73: > Do you think we should use 0000FFFF instead of FFFF to match all any instances and any types in ue_id?

It doesn't make any difference FMPOV

stevenhartley:
@evshary today on the MQTT call we concluded that we need to construct the MQTT topic including the full source and sink URI information so that we can register for notifications that require sink and source. We register a listener for notifications passing the source UUri and the destination UUri (sink) so that the message is routed directly to the uE that consumes the notification message. 

Here is more information:

> Based on our live discussion just now, we have concluded that in order to support notifications correctly (that will have source and sink addresses) we need to expand and update the MQTT5 topic to be the following:
>   
>   {client_identifier}/{source.authority_name}/{source.ue_id}/{source.ue_version_major}/{source.resource_id}/{sink.authority_name}/{sink.ue_id}/{sink.ue_version_major}/{sink.resource_id}
>   
>   Below are some examples using this updated notation:
>   
>   Specific Publish Topic: d/{source.authority_name}/{source.ue_id}/{source.ue_version_major}/{source.resource_id}/////
>   All (incoming) requests to a specific Method (what a service would want to do): d/+/+/+/+/{sink.authority_name}/{sink.ue_id}/{sink.ue_version_major}/{sink.resource_id}
>   Streamer registering to receive any sent notifications, requests, or responses destine to its device:
>   d/+/+/+/+/{sink.authority_name}/+/+/0
>   Specific Notification topic: d/{source.authority_name}/{source.ue_id}/{source.ue_version_major}/{source.resource_id}/{sink.authority_name}/{sink.ue_id}/{sink.ue_version_major}/0

evshary:
@stevenhartley Updated. TBH, I'm not in favor of the complexity of Zenoh key or MQTT topic, but the majority rules.
I also tidied up the examples of Publish, Notification, and Request after referring to the MQTT spec.
@sophokles73 @PLeVasseur I would like to hear your comments. Is it clear enough to developers?

stevenhartley (approved):
LGTM

Issue #96, opened 2024-03-20
Remove Topic Creation and Deprecation from uSubscription APIs


stevenhartley:
Early on in the development of uProtocol we baked in an optimization for subscribers to know existence of topics in a system by adding the APIs to uSubscription CreateTopic() and (originally) DeleteTopic(). The optimization required producers to call this API before subscribers could subscribe however this optimization only causing issues with initialization and added additional unnecessary messaging each time a uService is created. 

Many other pub/sub communication protocols such as MQTT, SOME/IP, and zenoh do not require you to "create" a topic before you can subscribe to the topic, not to mention we have uDiscovery that could be leveraged to find out if a topic "exists" before you subscribe to it. 

This, issue shall be use to deprecate the CreateTopic() and DeprecateTopic APIs from uSubscription APIs and implementation to simplify the subscription flow for developers.

stevenhartley:
> @stevenhartley, is it correct to assume that producers would still do RegisterForNotifications() on a topic they're going to produce, and that this is mandatory for a producer?

No it is not mandatory for Producers, only for those who care if there are subscribers or not before they start producing (ex. we would not produce camera frames topic if there is no one who wants to view the camera as this is resource intensive opperation).

AnotherDaniel:
Ok - one more Q then, even though this is not immediately relevant for me: in the camera example, and with the CreateTopic call gone - a camera frame subscriber would then "blindly" subscribe to the topic they think is the correct one for getting camera pics, and only then would the producer begin to actually publish?

stevenhartley:
Merged

PR #108, opened 2024-04-05
Remove USubscription CreateTopic() and DeprecateTopic() APIs

stevenhartley:
The topic lifecycle APIs were added as an optimization early on so that subscribers could know if the topic they are subscribing to is valid and the producer is alive however this optimization has given us nothing but heartburn in production as it forces all publishers to call the API when then are initialized and it causes race conditions on initial boot-up. Furthermore, validation of topics could already be done by calling UDiscovery so there was no added benefit of adding the extra states to the subscription service.

#96


up-l3/usubscription/v3/README.adoc
```diff
@@ -142,53 +106,26 @@ Subscribers are automatically registered to receive subscription change notifica
 
 === Publishers
 
-* *MUST* *NOT* be permitted to call `RegisterForNotifications()` for topics they do not create through the  `CreateTopic()`, only the topics they produce
+* *MUST* *NOT* be permitted to call `RegisterForNotifications()` for topics they do not produce (i.e. `UEntity` of the producer is not equal to the `UEntity` in the passed `UUri`)
```
sophokles73, line 109:
```suggestion
* *MUST* *NOT* be permitted to call `RegisterForNotifications()` for topics they do not produce, i.e. the uEntity sending the request is not equal to the `UEntity` of the request message's `UUri` parameter.
```

up-l3/usubscription/v3/README.adoc
```diff
@@ -142,53 +106,26 @@ Subscribers are automatically registered to receive subscription change notifica
...
+Dispatchers are also permitted to register for subscription change notifications to facilitate the multicasting of messages. The mechanics and requirements of the dispatcher and uSubscription communication are platform deployment specific. 
 
 == Event Delivery
 
-uSubscription service, in addition to managing subscriptions for subscribers to topics, also plays a key role in providing event delivery configuration details as not every platform delivers events using the same mechanism. 
-
-the `EventDeliveryConfig` message is one of the fields returned to the subscriber to provide details of how to consume the data (how events are delivered). For example if the subscriber has to consume from a different topic or from different event infrastructure, this message will store those delivery semantics.
+uSubscription service, also provides a means to notify subscribers of how they shall consume the published events though the  `EventDeliveryConfig` returned in the `SubscriptionResponse` message. For example if the subscriber has to consume from a different topic or from different message infrastructure, this message will store those delivery semantics.
```
sophokles73, line 118:
```suggestion
The uSubscription service also allows to indicate to subscribers how they shall consume published events by means of the `EventDeliveryConfig` returned in the `SubscriptionResponse` message. For example, if the subscriber has to consume from a different topic or from different messaging infrastructure, this message will store the corresponding delivery semantics.
```

up-l3/usubscription/v3/README.adoc
```diff
@@ -142,53 +106,26 @@ Subscribers are automatically registered to receive subscription change notifica
...
 
 NOTE: Delivery semantics (if any) are deployment specific and not covered in this specification
 
 == uSubscription Sequences
 
-In the following section, we will elaborate on the various subscription flows for local and remote topics. When a consumer subscribes to a remote topic, it is the responsibility of the Subscription Service to relay the subscription request to the remote Subscription Service as can be illustrated in the sequence diagrams below.
-
-For the flows below we will use a fictitious device called Device1 hosting a service called uExample that has a resource that emits a Message.
+In the following section, we will elaborate on the various subscription flows for local and remote topics. When a consumer subscribes to a remote topic, it is the responsibility of the Subscription Service to relay the subscription request to the remote Subscription Service as can be seen in the sequence diagrams below.
```
sophokles73, line 124:
```suggestion
In the following section, we will elaborate on the various subscription flows for local and remote topics. When a consumer subscribes to a remote topic, it is the responsibility of the (local) uSubscription service to relay the subscription request to the remote uSubscription service as can be seen in the sequence diagrams below.
```

sophokles73 (approved):
LGTM

sophokles73:
@AnotherDaniel this might be of interest to you as well, since you are working on an implementation ...

lynn2009:
@stevenhartley , The uSubscription still requires verification of whether the requested topic is valid. By removing the createTopic method, does this imply that uSubscription is depending on uDiscovery for functionality?

stevenhartley:
> @stevenhartley , The uSubscription still requires verification of whether the requested topic is valid. By removing the createTopic method, this implies a strong dependency on uDiscovery for its functionality.

@lynn2009 , why do you need to check the topic is valid or not?

lynn2009:
If uSubscription does not check the topic is valid or not, we will run into the case the subscriber can subscribe an invalid topic and get success response.

stevenhartley:
> If uSubscription does not check the topic is valid or not, we will run into the case the subscriber can subscribe an invalid topic and get success response.

correct

mayureshkar:
@stevenhartley  @lynn2009  with createTopic removed, what constitutes an 'invalid topic'?

stevenhartley:
invalid UUri

Issue #102, opened 2024-04-02
Update up-core-api/uprotocol/uri.proto to make microuri mandatory and longuri optional.


ruchirchauhan:
@stevenhartley, as per our discussion earlier, I think we need to update uri.proto to support mandatory micro and optional long uri formats. 

Proposed change:
1. UEntity. name = currently is required field, should become optional
2. UEntity.id = currently is optional field, should become required
3. UEntity.version_major/minor = currently optional, Not sure, do we need to update these too, may be not ?
4. UResource.name = currently is required field, should become optional
5. UResource.instance = Not sure ?
6. UResource.id = currently is optional field, should become required

Note: There is a check corresponding to UResource.name in up-client-zenoh-cpp, which checks for its value to be "rpc", for rpc type messages. Are we planning on doing that still ? If yes then, may be we can't make longuri optional ans that requires UResource.name to be present, thoughts ?

stevenhartley:
Hi @ruchirchauhan ,
I agree with your proposed changes (for 5 it should be optional). Could you make the corresponding change in up-spec now that I merged up-core-api ?

sophokles73:
Based on the changes proposed here (and implemented in #110 already), we also need to change the [UUri specification](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#3-validators) as well. Actually, IMHO we should remove the whole section from the UUri spec and instead add corresponding informtation to the other documents where we actually require UUris to contain certain information, e.g. in a the context of UTransport, where a UUri needs to be resolved. I had already created https://github.com/eclipse-uprotocol/up-rust/issues/78 because the current validation is too restrictive/inappropriate for certain use cases.

@stevenhartley @PLeVasseur WDYT?

stevenhartley:
@sophokles73 agree.

evshary:
I guess short-form UUri can always be transformed from micro-form UUri, right?
Want to make sure we don't need to make sure the UUri is short-form in up-client-zenoh.

> Note: There is a check corresponding to UResource.name in up-client-zenoh-cpp, which checks for its value to be "rpc", for rpc type messages. Are we planning on doing that still ? If yes then, may be we can't make longuri optional ans that requires UResource.name to be present, thoughts ?

There should be a way to check RPC type messages in micro-form UUri. (id < 0x800)
https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#3-validators
[...]

PR #110, opened 2024-04-09
Making Uri proto, microuri support mandatory and longuri support optional

ruchirchauhan:
This PR implements issue #102 

It modifies uri.proto:

1. UEntity name is now optional
2. UResource name is now optional

### DIFF
--- a/up-core-api/uprotocol/uri.proto
+++ b/up-core-api/uprotocol/uri.proto
@@ -74,8 +74,8 @@ message UAuthority {
 // A uE that publishes events is a <b>Service</b> role.<br>
 // A uE that consumes events is an <b>Application</b> role.
 message UEntity {
-  optional string name = 1;    // Name of the entity
-  uint32 id = 2;      // The numeric ID for the uEntity assigned from the name/number registry
+  optional string name = 1;          // Name of the entity
+  uint32 id = 2;                     // The numeric ID for the uEntity assigned from the name/number registry
   optional uint32 version_major = 3; // optional major version of the uEntity
   optional uint32 version_minor = 4; // optional minor version of the uEntity
 }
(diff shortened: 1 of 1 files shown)


up-core-api/uprotocol/uri.proto
```diff
@@ -74,8 +74,8 @@ message UAuthority {
 // A uE that publishes events is a <b>Service</b> role.<br>
 // A uE that consumes events is an <b>Application</b> role.
 message UEntity {
-  string name = 1;    // Name of the entity
-  optional uint32 id = 2;      // The numeric ID for the uEntity assigned from the name/number registry
+  optional string name = 1;          // Name of the entity
+  uint32 id = 2;                     // The numeric ID for the uEntity assigned from the name/number registry
```
stevenhartley, line 78:
```suggestion
  optional uint32 id = 2;                     // The numeric ID for the uEntity assigned from the name/number registry
```

up-core-api/uprotocol/uri.proto
```diff
@@ -91,10 +91,10 @@ message UEntity {
 // Resources are unique when prepended with UAuthority that represents the device and
 // UEntity that represents the service.
 message UResource {
-  string name = 1;      // Name of the resource
+  optional string name = 1;      // Name of the resource
   optional string instance = 2;  // Instance of the resource
   optional string message = 3;   // Message type for the resource
-  optional uint32 id = 4;        // Numerical representation of the UResource portion of the UUri
+  uint32 id = 4;                 // Numerical representation of the UResource portion of the UUri
```
stevenhartley, line 97:
```suggestion
  optional uint32 id = 4;                 // Numerical representation of the UResource portion of the UUri
```

stevenhartley (changes requested):
The purpose of this change is to allow us to build URIs that contain only numbers that could be used for micro and short URI format however if we put id as mandatory and name optional, we will not be able to build URIs that are in long format only (meaning URIs that only contain names). Since UUri is a container to store the URI information, it needs to support the ability to contain in its belly:

1. Names only
2. Numbers only
3. Names & numbers

The changes I've proposed allows the UUri to contain any of the above information, please consider the suggestions

up-core-api/uprotocol/uri.proto
ruchirchauhan: Done

stevenhartley (approved):
Approved

sophokles73:
@stevenhartley it would be nice if we could squash PRs like this before merging in the future to keep the commit log in a clean and consistent state. We now have three commits in the log, each of which represents intermediate work and which only make sense when applied together ...

Issue #93, opened 2024-03-19
Update SOME/IP specification to uProtocol 1.5.x


stevenhartley:
The current SOME/IP specification did not take into consideration the uPClient library concept and ideas in that the translation to/from uMessages to SOME/IP messages will be done inside of the up-client-someip-xxx library under uTransport implementation. the spec requires some cleanup and clarification to align with latest architecture and concepts.

PR #117, opened 2024-04-18
SOME/IP Specification Cleanup

stevenhartley:
The following change is a reboot of the SOME/IP specifications in an attempt to drastically simplify to show how we map up-l1 messages to SOME/IP messages and uSubscription messages to SOME/IP-SD messages.  the older specifications had a lot of uProtocol version 1.3.6 concepts (cloudevents) that were rather complicating the situation.

#93


up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+----
+
+== Overview
+
+Scalable service-Oriented MiddlewarE over IP (SOME/IP) is a communication protocol used in the automotive industry for exchanging data between electronic control units (ECUs) in a vehicle. 
+SOME/IP provides a framework for efficient and reliable communication between ECUs over IP networks. It is designed to support real-time and high-bandwidth communication requirements in automotive systems. The protocol allows ECUs to discover and communicate with each other, exchange messages, and provide services in a distributed system.
+SOME/IP is based on the Internet Protocol (IP) and uses UDP (User Datagram Protocol) as the transport layer. It supports various communication patterns, including unicast, multicast, and broadcast. The protocol defines a set of message formats and procedures for encoding and decoding data, handling service discovery, and managing communication sessions.
```
int0x27, line 32: SOME/IP also uses TCP transport

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+.Sink & Source URI Mapping
+[sink-source-uri-mapping]
+|===
+|*UUri Portion* |*SOME/IP Field* |*Description*
+|*`UAuthority::ip`* |IP address & port |IP address & port (destination
+endpoint) of the mDevice
+
+|*`UEntity::id`* |Service ID |Software entity ID that is placed in the upper 16
+bits of the SOME/IP Message ID header field
+
```
int0x27, line 68: SOME/IP header does not include Instance ID, but it is used in SOME/IP SD and also required for vsomeip/sommr API.
Can we map it to UUri? e.g. ServiceID | InstanceID both u16 as `UAuthority::ip`: u32

[UUri spec](https://github.com/eclipse-uprotocol/up-spec/blob/someip-cleanup/basics/uri.adoc#uuri-specification) also refers to UResource instance, but maybe it is outdated.

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+|*`UResource::id`* |Method/Event ID |Identifier of the method/resource placed in the
+lower 15 bits of the SOME/IP Message ID header field
```
int0x27, line 70: 16 bits?

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+
+|===
+
+
+==== UUIDs
+
+link:../basics/uuid.adoc[uProtocol UUID specifications] create a unique identifier for each message along with timestamp information. the ID is used for correlate between request and response as well. SOME/IP instead defines the ID as 16bit client ID + 16 bit session ID (that is incremented). When messages are converted SOME/IP to/from uProtocol, care must be taken to ensure that the SOME/IP ID and uProtocol UUID are properly mapped, especially when corelating a request to a response. 
+
+* Generated SOME/IP events *SHOULD* set the 16 bit client ID to 0 per SOME/IP specifications 4.1.2.4
```
int0x27, line 79: Maybe you should refer to `[PRS_SOMEIP_00925]`?

> • Set the Client ID to 0x00. Set the Session ID according to
[PRS_SOMEIP_00932], [PRS_SOMEIP_00933], and [PRS_SOMEIP_00521].
In case of active Session Handling the Session ID shall be incremented upon
each transmission.

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+
+.UMessageType Mapping
+[umessagetype-mapping]
+|===
+| UMessageType | SOME/IP Type | Details
+
+|`UMESSAGE_TYPE_PUBLISH` or `UMESSAGE_TYPE_NOTIFICATION` |`NOTIFICATION` |Both publication and notifications events are the same for SOME/IP
+
+| `UMESSAGE_TYPE_REQUEST` |`REQUEST` |Requests type message
+
```
int0x27, line 94: Shall we handle `REQUEST_NO_RETURN` as per `[PRS_SOMEIP_00055]` for fire and forget requests?
Could be handled internally for the client

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+|===
+| UCode | SOME/IP Error Codes
+
+|`OK` |`E_OK`
+|`INVALID_ARGUMENT` |`E_WRONG_MESSAGE_TYPE` or `E_UNKNOWN_METHOD`
+|`DEADLINE_EXCEEDED` |`E_TIMEOUT`
+|`NOT_FOUND` |`E_UNKNOWN_SERVICE`
+|`UNAVAILABLE` |`E_NOT_READY`
+|`DATA_LOSS` |`E_MALFORMED_MESSAGE`
+|`INTERNAL` |`E_NOT_REACHABLE`
```
int0x27, line 132: Also add: `E_NOT_OK` here

- `E_WRONG_PROTOCOL_VERSION`, `E_WRONG_INTERFACE_VERSION` not described as well

For more error codes add link to SOME/IP Protocol specification: `[PRS_SOMEIP_00191]`

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+
+== Application Layer Translation
+
+Application (or message payload) translation is the process of converting SOME/IP-SD subscription and discovery messages, to/from uDiscovery and uSubscription Messages. 
+
+=== uSubscription
+
+The following section will elaborate only on the translation of uSubscription messages to/from SOME/IP-SD messages. Subscription state (persistent or not) is handled in the uSubscription services and not at the transport layer or this component.
+
+The following section we will elaborate on how Eventgroup Entry types (5.1.3.1) are mapped to uSubscription messages for the subscribe
```
int0x27, line 154: As it was harder to find this reference, can you also mention `[PRS_SOMEIPSD_00385]`

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+==== Common Fields
+
+<<common-field-mappings>> table below illustrates the common SOME/IP-SD EventGroup Entry fields that are present in for all SOME/IP-SD Eventgroup entry types (`SubscribeEventgroup`, `SubscribeEventGroupAck`, `SubscribeEventgroupNack`, `StopSubscribeEventGroup`). These fields are then mapped to uProtocol `UUri` attributes usined in uProtocol `UMessage` for performing subscription operations. 
+
+.Common Field Mappings
+[#common-field-mappings, width=70%]
+|===
+|Eventgroup Entry Field |UUri
+
+|Service ID |`UEntity.id`
```
int0x27, line 166: **NOTE:** For API calls also Instance ID is required, but could be defined as static configuration in the client.
Can we add it to `UAuthority::ip`?

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+
+|Major Version | `UEntity.major_version`
+
+|Eventgroup ID | `UResource.id`
+
+|===
+
+NOTE: `UUri.authority.ip` is translated to/from IPv4 (and/or IPv6) Endpoint Option of the SOME/IP-SD message
```
int0x27, line 174: vsomeip API does not provide anything to get remote endpoint IP, it uses <service/instance/..> mappings internally to find which tcp/udp endpoint is used.
For me setting remote IPs in the authority for SOME/IP seems wrong.

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+.EventGroup Entry Type Mapping
+[#eventgroup-entry-mapping, cols="1,1,2"]
+|===
+|Eventgroup Entry Type |uSubscription Message | Additional Details
+
+|`SubscribeEventGroup` 
+|`SubscriptionRequest` 
+a|The message is used to subscribe to a topic.
+
+* Instance ID *SHALL* assume to be 0
```
int0x27, line 187: Instance ID should be arbitrary value expected from the SOME/IP service. e.g. if you hardcode 0 and the service only handles instance 1, subscription will be NACKed.

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+* *SHALL* assume version 0 to match uProtocol backwards compatibility requirements
+* TTL (`SubscribeAttributes.ttl`) *SHALL* set to 0xFFFFFF to indicate that the subscription should remain for the duration of the ignition cycle
```
int0x27, line 189: Remark about vsomeip: there is global Service Discovery TTL (static config) and it is used for all subscriptions. It's not possible to have it per service or dynamic. Probably it should match remote service shortest TTL

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+[#eventgroup-entry-mapping, cols="1,1,2"]
+|===
+|Eventgroup Entry Type |uSubscription Message | Additional Details
+
+|`SubscribeEventGroup` 
+|`SubscriptionRequest` 
+a|The message is used to subscribe to a topic.
+
+* Instance ID *SHALL* assume to be 0
+* *SHALL* assume version 0 to match uProtocol backwards compatibility requirements
```
int0x27, line 188: do you mean Major Version here? The service may NACK the subscription if it requests different major version.

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+When receiving uProtocol initiated requests:
+
+* *MUST* cache the request `UAttributes` for a maximum of `ttl` so that it can be used to build a response `UAttributes` when receiving a response from SOME/IP
+
+```
+response.priority = request.priority
+response.reqid = request.id
+```
+
+* Auto-generated SOME/IP REQUEST message, *MAY* set the 16 bit client ID to the 0 per SOME/IP specifications
```
int0x27, line 110: This is only for notifications, but not requests

up-l1/someip.adoc
```diff
@@ -0,0 +1,221 @@
...
+.Common Field Mappings
+[#common-field-mappings, width=70%]
+|===
+|Eventgroup Entry Field |UUri
+
+|Service ID |`UEntity.id`
+
+|Major Version | `UEntity.major_version`
+
+|Eventgroup ID | `UResource.id`
```
ruchirchauhan, line 170: Since Eventgroup ID and Event ID both are being mapped to UResource.id, I am guessing it means that this still doesn't support multiple event ids in an eventgroup and that eventgroup id should be equal to event id ?

up-l1/someip.adoc
int0x27: As I checked again UUri spec, we don't have u32 id to use for ServiceID + InstanceID.
It could be handled as a client option per Service ID along with major/minor service version.

up-l1/someip.adoc
int0x27: Even if we have 1 : 1 mapping for events and groups, they should be different IDs for the client library.
Everything is hardcoded in the SOME/IP service and the client library must follow those defined IDs.

I suggest we use Event ID as the primary mapping (as `UResource.id`) also for subscribe(). 
The client implementation may look up its static configuration to get additional information about Eventgroup ID to call `subcribe()`, but then events will be coming with Event ID, so subscribing to that Event ID makes more sense.

As a side effect subscribing to ECU with multiple events in the event group may trigger receive `on_message()` notifications for other events as well (those can be filtered on client level)

sagarsshah: I liked Ivan's suggestion.  lets keep event group logical entity (i.e. not defined in uProtocol). we can map an event group or event groups to a service.  also we can consider adding new api in sdk for registerListnergroup() which can be used for registering logical events (event id) for a services/event group.

stevenhartley: fixed in latest change.

up-l1/someip.adoc
stevenhartley: @int0x27 do we need to explicitly state this?

up-l1/someip.adoc
stevenhartley: No the UService.id

up-l1/someip.adoc
stevenhartley: @int0x27 , given your comment, should we assume live forever?

up-l1/someip.adoc
stevenhartley: Per discussion thread I'm changing this to 1 in lieu of 0

up-l1/someip.adoc
stevenhartley: We are merging back name, ip, and id per #120 so it is no longer ambiguous (and split into three parts). This means that the name has to be either ip address, a hostname, or an id and has to be globally unique.

up-l1/someip.adoc
stevenhartley: done, @int0x27 please check

up-l1/someip.adoc
stevenhartley: done

up-l1/someip.adoc
stevenhartley: done

up-l1/someip.adoc
stevenhartley: fixed

up-l1/someip.adoc
stevenhartley: per discussion below, we will stick with instance 1. Need to discuss instance id further though.

stevenhartley:
Closing so we can use #165 instead

Issue #115, opened 2024-04-17
Removal of UAuthority id & ip


stevenhartley:
**ISSUE SHIFTING TO DROPPING IP and ID**

The purpose of UAuthority.name was to holds the [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2) UAuthority containing username, hostname, and port notation.

UAuthority.number.ip purpose was to hold the IP address in 4 byte (IPv4) or 16 byte (IPv6) notation.

UAuthority.number.id purpose was to contain a simple id that could be used in micro form (i.e. fit into 255 bytes) 
UAuthority's id property is [defined as a byte array](https://github.com/eclipse-uprotocol/up-spec/blob/8fcb37da046193599aa961a78f05af4eedb90025/up-core-api/uprotocol/uri.proto#L63) without any further assumptions regarding its semantics nor encoding. [ShortUriSerializer](https://github.com/eclipse-uprotocol/up-java/blob/90c3e1ed9b80ca03629b3ecba41f99fb3e2a5e69/src/main/java/org/eclipse/uprotocol/uri/serializer/ShortUriSerializer.java#L80) assumes that the bytes in the array actually represent a UTF8 encoded string but it is not explicitly defined as such. 

The following PR shall be used to explicitly set the format of ID to be UTF8 by changing the type from byte to string.

Additionally, there are no explicit guidelines for how to represent the IP address in string form so the recommendation will be to update the uri spec to state that it MUST follow [rfc5952](https://www.rfc-editor.org/rfc/rfc5952)

sophokles73:
> The purpose of UAuthority.name was to holds the [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2) UAuthority containing username, hostname, and port notation.

I was not aware of the fact that it should also hold userinfo and port. How does that work when using IP addresses? Is the userinfo and port irrelevant in that case?

My understanding was (and is) that UAutority can contain a [URI _host_](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2) only, but not a port nor userinfo ...

sophokles73:
I guess my point is that UAuthority, the way it is currently specified, _could_ only contain userinfo and/or port in its `name` property but there is no way of storing neither userinfo nor port in either `ip` or `id`.

FMPOV the spec therefore needs to be changed to explicitly rule out the inclusion of userinfo and/or port in the `name`.

sophokles73:
[...]

PR #121, opened 2024-04-22
Make UUri specification more formal

sophokles73:
This addresses #115 and other issues.

The UUri specification has been lacking many details that are relevant
for correctly implementing the data model in client libraries.

A lot of the specification's textual content has been transformed into
formal requirements based on invariants and predicates expressed using
the Object Constraint Language.

Last but not least, the 'id' and 'ip' properties have been removed from
UAuthority's data model because they did not really serve a specific
pupose beyond the semantics of the 'name' property.


sophokles73:
The 8-bit length field of the Micro Format allows for the authority to use up to 255 bytes. The `name` property, on the other hand, contains (unicode) characters which may be encoded into up to three bytes (using UTF8) per character.

This shouldn't necessarily be an issue because most characters used for the logical name will consist of ASCII characters only and those are encoded into a single byte (using UTF8).

However, IMHO we should properly handle cases in which the authority would be encoded into more than 255 bytes. Given that the Long and Short Format do not put any constraints on the authority length, I wonder if we really need to constrain it in the Micro Format.

If we need/want to set a limit, then IMHO we should define this constraint on the Object Model itself. We should then probably only allow ASCII characters in the authority name so that we know upfront, how many bytes it would be encoded into ...

sophokles73:
@stevenhartley @PLeVasseur @tamarafischer @AnotherDaniel @evshary

any feedback would be welcome

stevenhartley:
> Here's my first shot at a UUri spec formalization.
> 
> One thing that still bothers me: The 8-bit length field of the Micro Format allows for the authority to use up to 255 bytes. The `name` property, on the other hand, contains (unicode) characters which may be encoded into up to three bytes (using UTF8) per character.
> 
> This shouldn't necessarily be an issue because most characters used for the logical name will consist of ASCII characters only and those are encoded into a single byte (using UTF8).
> 
> However, IMHO we should properly handle cases in which the authority would be encoded into more than 255 bytes. Given that the Long and Short Format do not put any constraints on the authority length, I wonder if we really need to constrain it in the Micro Format.
> 
> If we need/want to set a limit, then IMHO we should define this constraint on the Object Model itself. We should then probably only allow ASCII characters in the authority name so that we know upfront, how many bytes it would be encoded into ...

@sophokles73 we need to follow [RFC3968](https://datatracker.ietf.org/doc/html/rfc3986) which means it cannot be any form of UTF-8 character only those characters listed in [Appendix A](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), a collection of ABNF for URI.

sophokles73:
> we need to follow [RFC3968](https://datatracker.ietf.org/doc/html/rfc3986) which means it cannot be any form of UTF-8 character only those characters listed in [Appendix A](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), a collection of ABNF for URI.

Well, for the long and short form URIs that is done by means of _percent encoding_ any illegal characters. However, that also increases the number of ASCII characters in the URI.

For the micro form serialization I do not see why we would be bound by RFC3986. We are not producing a (textual) URI but a custom binary representation and as long as we can recreate the original UUri from it, we should be safe, shouldn't we?

Another approach, of course, would be to restrict UAuthortity.name to only contain ASCII characters in the first place, if that is what you mean.

stevenhartley:
> > we need to follow [RFC3968](https://datatracker.ietf.org/doc/html/rfc3986) which means it cannot be any form of UTF-8 character only those characters listed in [Appendix A](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), a collection of ABNF for URI.
> 
> Well, for the long and short form URIs that is done by means of _percent encoding_ any illegal characters. However, that also increases the number of ASCII characters in the URI.
> 
> For the micro form serialization I do not see why we would be bound by RFC3986. We are not producing a (textual) URI but a custom binary representation and as long as we can recreate the original UUri from it, we should be safe, shouldn't we?
> 
> Another approach, of course, would be to restrict UAuthortity.name to only contain ASCII characters in the first place, if that is what you mean.

If we rescript to ASCII characters, we will be in compliance with RFC3986 so sure lets do it

basics/uri.adoc
```diff
@@ -25,253 +26,488 @@ specific language governing permissions and limitations
...
-|Item |Value |Description
+.REQ~UURI-RESOLVED
+[#req~uuri-resolved~1]
+--
+[sidebar]
+Is this really needed? It is unclear where this check might be needed.
 
-|*`SCHEME`* |`up:` |Scheme (protocol name)
-|*`UAUTHORITY_NAME`* |unreserved |Canonical hostname of the device following https://datatracker.ietf.org/doc/html/rfc1035[RFC1035] requirements.
+A UUri which contains corresponding numerical identifiers for logical names used in its components, is called a _resolved_ UUri.
```
stevenhartley, line 268: A UUri that contains names and numbers is resolved

basics/uri.adoc
```diff
@@ -25,253 +26,488 @@ specific language governing permissions and limitations
...
+----
+context UUri
+def: isResolved() : Boolean =
+  ( entity.name->notEmpty() implies entity.id->notEmpty() )
+  and
+  ( resource.name->notEmpty() implies resource.id.notEmpty() )
+----
 
-|*`UE_VER_MAJOR`* |pchar | uEntity major version number only.
+NOTE: This means that a UUri which _only_ contains numerical identifiers is a _resolved_ UUri.
```
stevenhartley, line 279: no it needs names and numbers

basics/uri.adoc
```diff
@@ -25,253 +26,488 @@ specific language governing permissions and limitations
...
+Otherwise, a long form URI's _fragment_ *MUST* be mapped to/from the UUri's `resource.message` property following the rules defined in https://datatracker.ietf.org/doc/html/rfc3986#section-3.5[RFC3986, Section 3.5].
+--
 
- *MAY* be a valid vehicle VIN (17 MSB)
- *MUST* fit within 255 bytes (when used for micro form)
+.REQ~LONG_FORM_VALIDATION
+[#req~long-form-validation~1]
+--
+[sidebar]
+Is this needed? It currently only verifies that the invariants hold (which, by definition, they should do anyway) ...
```
stevenhartley, line 349: If you build a UUri from Short form string then this would be false as there is no name for entity or resource

basics/uri.adoc
```diff
@@ -282,46 +518,45 @@ Local UUris do not contain authority and *SHOULD* be used at all times when addr
...
-[#vin-micro-uri,java]     
+|===
+--
+
+=== Remote Micro Uri
+
+.REQ~MICRO-FORM-REMOTE
+[#req~micro-form-remote~1]
+--
+A UUri that contains a non-empty authority *MUST* be mapped to a byte array in network byte order as defined below. The structure extends the one defined by <<_local_micro_uri>>.
```
stevenhartley, line 557:
```suggestion
A UUri that contains a non-empty authority *MUST* be mapped to a byte array in network byte order as defined below. The structure extends the one defined by <<Local Micro Uri>>.
```

basics/uri.adoc
```diff
@@ -330,99 +565,424 @@ Remote UUris mean that the uThing is running on a different device and we need t
...
+|TYPE
+a| The type of authority contained in this micro URI. Occupies bits 8 through 15 (octet 1). +
+*MUST* be set to `0x03` for a URI with an authority.
 
-The following examples are based off the USubscription notification topic (see usubscription.proto).
+|NAME_LEN
+a|The length of the `UAUTHORITY_NAME`. Occupies bits 64 through 71 (octet 24). +
+*MUST* contain the length of the `UAUTHORITY_NAME` field in bytes.
 
+|UAUTHORITY_NAME
```
stevenhartley, line 585: @sophokles73, This is simpler in that there is only local vs remote however it means that a 4 byte IPv4 address could take up to 15 bytes and the IPv6 16 bytes could take up to 39 bytes.

basics/uri.adoc
```diff
@@ -330,99 +565,424 @@ Remote UUris mean that the uThing is running on a different device and we need t
...
+l|[
+  0x01, // version
+  0x00, // type
+  0x80, 0x00, // UResource ID
+  0x00, 0x00, // UEntity ID
+  0x03, // UEntity major version
+  0x00, // unused
+]
+
+l|`1G1YZ23J9P5800001`
```
stevenhartley, line 615:
```suggestion
l|1G1YZ23J9P5800001
```

basics/uri.adoc
```diff
@@ -330,99 +565,424 @@ Remote UUris mean that the uThing is running on a different device and we need t
...
 
-|Short
-| `//192.168.1.100/0/3/32768`
+Alternatively, the `UUri` class might provide corresponding methods.
 
-|Micro
-|`[1, 1, -128, 0, 0, 0, 3, 0, -64, -88, 1, 100]`
+[source,java]
+----
+public class UUri {
```
stevenhartley, line 667: UUri is defined in proto and the classes are generated to which we cannot extend. I recommend we do not show coding examples in this spec but state that serializers and deserializers and validators are required.

basics/uri.adoc
```diff
@@ -330,99 +565,424 @@ Remote UUris mean that the uThing is running on a different device and we need t
...
 
+[source,ocl]
+----
+context UAuthority
+def: matches(candidate : UAuthority) : Boolean =
+  ( self.name->notEmpty() implies candidate.name->notEmpty()
+    -- authority.name is case-insensitive per RFC3986
+    and self.name.toLower() = candidate.name.toLower() )
+  and
+  ( self.ip->notEmpty() implies candidate.ip->notEmpty() and self.ip = candidate.ip )
```
stevenhartley, line 830: we can get rid of the ip and id

basics/uri.adoc
```diff
@@ -330,99 +565,424 @@ Remote UUris mean that the uThing is running on a different device and we need t
...
+  self.authority.matches(candidate.authority)
+  and
+  self.entity.matches(candidate.entity)
+  and
+  self.resource.matches(candidate.resource)
+----
+
+[NOTE,#matching-local-uuris]
+====
+An empty authority pattern matches _any_ authority according to the `UAuthority::matches` predicate. In particular, it is not possible to use an empty authority pattern to filter out _non-local_ candidate UUris.
```
stevenhartley, line 869: I think I get it, you're saying (had to scratch my head multiple times).

basics/uri.adoc
```diff
@@ -330,99 +565,424 @@ Remote UUris mean that the uThing is running on a different device and we need t
...
+When a UUri is not resolved, i.e. only contains names, the only way to resolve the UUri is to lookup in uDiscovery the missing information. UEntity name and numbers are declared in the uService's protos by adding the service metadata attributes using https://protobuf.dev/programming-guides/proto2/#options[Protobuf Options]. This information for the uE is then exposed through the uService generated stubs.
  
 UResource portion of the UUri is also encoded in the uService protos and can be fetched by application developers using https://protobuf.dev/reference/csharp/api-docs/class/google/protobuf/reflection/message-descriptor[MessageDescriptors]. Application developers are then able to easily construct resolved UUris using the above mentioned generated code. link:../up-l2/README.adoc[Dispatchers] however do not have the name-to-number mapping metadata as they are only provided either serialized Long or Micro Uris so if they have to send a Uri in a different format, they have no choice but to dynamically lookup (resolve) the Uri.
```
stevenhartley, line 986:
```suggestion
UResource portion of the UUri is also encoded in the uService protos and can be fetched by application developers using https://protobuf.dev/reference/csharp/api-docs/class/google/protobuf/reflection/message-descriptor[MessageDescriptors]. Application developers are then able to easily construct resolved UUris using the above mentioned generated code. link:../up-l2/README.adoc[Dispatchers] however do not have the name-to-number mapping metadata as they are only provided serialized Uris so if they have to send a Uri in a different format, they have no choice but to dynamically lookup (resolve) the Uri. 
```

basics/uri.adoc
```diff
@@ -330,99 +565,424 @@ Remote UUris mean that the uThing is running on a different device and we need t
...
  
 UResource portion of the UUri is also encoded in the uService protos and can be fetched by application developers using https://protobuf.dev/reference/csharp/api-docs/class/google/protobuf/reflection/message-descriptor[MessageDescriptors]. Application developers are then able to easily construct resolved UUris using the above mentioned generated code. link:../up-l2/README.adoc[Dispatchers] however do not have the name-to-number mapping metadata as they are only provided either serialized Long or Micro Uris so if they have to send a Uri in a different format, they have no choice but to dynamically lookup (resolve) the Uri. 
 
-IMPORTANT: If a uE has a specific Uri format requirements (Long vs Micro), they *MUST* document them in their respective APIs specifications. Please see link:../up-l3/README.adoc#_core_platform_ues[Core uProtocol uEs] for more specific requirements for uProtocol core services.
+IMPORTANT: If a uE has a specific UUri format requirements (Long vs Micro), they *MUST* document them in their respective APIs specifications. Please see link:../up-l3/README.adoc#_core_platform_ues[Core uProtocol uEs] for more specific requirements for uProtocol core services.
```
stevenhartley, line 988:
```suggestion
IMPORTANT: If a uE has a specific UUri format requirements (UUri object, `string` Long/Short, or `bytes` format  Micro), they *MUST* document them in their respective APIs specifications. Please see link:../up-l3/README.adoc#_core_platform_ues[Core uProtocol uEs] for more specific requirements for uProtocol core services.
```

stevenhartley (changes requested):
Please see comments. Much better than before for sure!

basics/uri.adoc
```diff
@@ -25,253 +26,488 @@ specific language governing permissions and limitations
...
 .URI
 image::rfc3986.png[RFC3986 URI]
 
-URI specifications are covered at length in RFC3986 and will not be covered in this document, we will focus on the additions to RFC3986 that are specific to the uProtocol.
+Using this terminology, uProtocol URIs (UUri) can be structured like this:
 
-The following is taken from https://datatracker.ietf.org/doc/html/rfc3986#appendix-A[RFC3986 Appendix-A] and shall be used as reference in this specification:
+.UUri Specification
+[#uuri-specification]
+image::uri.drawio.svg[uProtocol Uri Specification]
```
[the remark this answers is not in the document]
sophokles73: I'll do my best :-)

basics/uri.adoc
```diff
@@ -25,253 +26,488 @@ specific language governing permissions and limitations
...
 
-[source]
+A UUri can be represented as
+
+* an instantiation of the <<_data_model,UUri Data Model>> in a supported programming language,
```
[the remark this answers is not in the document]
sophokles73: I'd like to refer to the section, not the UML class diagram here ...

basics/uri.adoc
```diff
@@ -25,253 +26,488 @@ specific language governing permissions and limitations
...
+* a URI containing (logical) names only as defined by <<_long_form>>,
```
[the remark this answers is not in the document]
sophokles73: is there a particular reason why you would rather use the section title instead of the identifier created by adoc for it?

basics/uri.adoc
```diff
@@ -25,253 +26,488 @@ specific language governing permissions and limitations
...
+In the remainder of this document, the term _UUri_ is used to refer to an instantiation of the data model, whereas the terms _long form URI_, _short form URI_ and _micro form URI_ are used to refer to its other representations respectively.
+
+== Data Model
+
+The UUri data model is defined in link:../up-core-api/uprotocol/uri.proto[uri.proto] using the proto3 type system.
+The diagram below provides an overview of the types and their relations using UML2 notation.
+
+.UUri Data Model
+[#uuri-data-model]
+[plantuml, format="png", width="100%", id="UUri Object Model"]
```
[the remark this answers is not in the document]
sophokles73: Ok, I will replace this with a Mermaid class diagram

basics/uri.adoc
```diff
@@ -25,253 +26,488 @@ specific language governing permissions and limitations
...
+hide empty members
+
+UAuthority "authority" <--* UUri
+UEntity "entity" <--* UUri
+UResource "resource" <--* UUri
+@enduml
 ----
 
-<<uuri-specification>> diagram below provides an overview of the specifications that we will elaborate in this document. 
+NOTE: All properties of UAuthority, UEntity and UResource are _optional_, whereas all of UUri's aggregation relations have cardinality 1, i.e. the `authority`, `entity` and `resource` properties are _never_ empty.
```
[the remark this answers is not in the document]
sophokles73: In the proto3 file, all of UUri's properties are optional, i.e. they can be empty. In the class model, UUri's properties (represented by its aggregations) are **not** optional because they have cardinality 1

basics/uri.adoc
```diff
@@ -25,253 +26,488 @@ specific language governing permissions and limitations
...
-.Long Form URI Description
-[#long-form-uri]
-[width="100%",cols="23%,18%,59%",options="header"]
-|===
-|Item |Value |Description
+.REQ~UURI-RESOLVED
+[#req~uuri-resolved~1]
+--
+[sidebar]
+Is this really needed? It is unclear where this check might be needed.
```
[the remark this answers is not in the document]
sophokles73: Does this mean that we can **only** serialize to short form, if UUri has a non-empty entity.id and resource.id? If so, that would not match the other direction, i.e. we currently are able to deserialize a short form URI that does not contain an entity.id and/or resource.id (-> wildcards) ...

basics/uri.adoc
sophokles73: Is this about the following?
1. Short form URI `up://my-vin/1234/1/5678` gets deserialized into a UUri
2. UUri shall now be serialized to a long form URI but that would result in an empty URI because the entity.id and resource.id may not have been resolved to their corresponding names

I see your point but it (again) raises the question if we allow empty properties during serialization or not. The latter would mean that we do not support generating URIs that contain wildcards and that serialization behavior is inconsistent with deserialization, where we (want to?) support wilcards (i.e. empty path segments) ...

basics/uri.adoc
sophokles73: I just wanted to point out that you **can** also use IP address strings for the name. But our overall idea was to suggest that uEntities use logical names only for the authority ...

basics/uri.adoc
sophokles73: hmm, the VIN also is a literal value. We format the hex values accordingly so, for consistency, I'd rather keep this as is. Alternatively we can remove the formatting for all values ...

basics/uri.adoc
sophokles73: ok, so simply remove the code examples?

basics/uri.adoc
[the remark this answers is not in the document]
stevenhartley: It should refer to the section with that title unless you named the title and the diagram the same. The reason why I proposed the change is that the link did not work for me when I rendered it in the PR.

basics/uri.adoc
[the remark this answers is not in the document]
stevenhartley: Links were not working for me (see previous comment). If it works for you in github then we can resolve these changes

sophokles73:
@stevenhartley I have updated according to your feedback but also have added some questions to the comments ...

basics/uri.adoc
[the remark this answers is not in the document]
sophokles73: Ah, I see what you mean. The links work in the vscode preview :-( I will try to fix this ...

stevenhartley:
> @stevenhartley I have updated according to your feedback but also have added some questions to the comments ...

I answered your questions this morning.

sophokles73:
> > @stevenhartley I have updated according to your feedback but also have added some questions to the comments ...
> 
> I answered your questions this morning.

It looks like some of them slipped through ... maybe you haven't seen them because they have been collapsed by GitHub?

sophokles73:
@stevenhartley I just had a little chat with @AnotherDaniel about his problems using UUris from within his uSubscription implementation which are mainly related to the duality of identifiers (name vs. id) :-)
We ended up thinking out loud about trying to get rid of the numerical id properties in the same way as we did in UAuthority, i.e. leave the resolving of logical names to identifiers to the UTransport that actually requires it. Currently, this would only affect the SOME/IP client if I am not mistaken and we could imagine it maintaining a mapping of names <-> (SOME/IP) ids in very much the same way as we let it do the mapping of names <-> IP addresses.

Would it make sense to think a little more into that direction or is it totally out of the question from your point of view?

stevenhartley:
> @stevenhartley I just had a little chat with @AnotherDaniel about his problems using UUris from within his uSubscription implementation which are mainly related to the duality of identifiers (name vs. id) :-) We ended up thinking out loud about trying to get rid of the numerical id properties in the same way as we did in UAuthority, i.e. leave the resolving of logical names to identifiers to the UTransport that actually requires it. Currently, this would only affect the SOME/IP client if I am not mistaken and we could imagine it maintaining a mapping of names <-> (SOME/IP) ids in very much the same way as we let it do the mapping of names <-> IP addresses.
> 
> Would it make sense to think a little more into that direction or is it totally out of the question from your point of view?

No we cannot do this (leave it up to the transport layer). The source of truth for service ids is the proto and is known to the application layer code or uDiscovery (not the transport layer). UAuthority to ip address can be done using linux resolver or static IP address table.

AnotherDaniel:
Hm, I hear what you're saying. So pushing ahead with uDiscovery, and making that the source of truth for this kind of lookup - based in the proto definitions - would that be an option?

I really don't like this information-duality between names and ids in the UURi, also the fact that we are redefining/extending what an uri is... feels off. Would much rather have uris just be uris, and have a well defined, separate way to translate to id representation and back (like url - DNS - ip)

stevenhartley:
I'm open to other suggestions and agree when we only had long form URI it was much simpler. I haven't been able to think of a different solution.

AnotherDaniel:
> I'm open to other suggestions and agree when we only had long form URI it was much simpler. I haven't been able to think of a different solution.

This might become too big a discussion for this PR - take this out into a dedicated issue maybe?

I'm asking this from my usual clueless point of view - but my thinking would be: shouldn't the up-client adapters be able to perform name-to-id translation for the messages that flow by, with the relevant information retrieved from uDiscovery (DNS-style) services? 

These uDiscovery entities in a certain scope (i.e. vehicle) might then address the question of how to efficiently manage their translation tables as some form of distributed configuration, that can be managed as such...

stevenhartley:
> > I'm open to other suggestions and agree when we only had long form URI it was much simpler. I haven't been able to think of a different solution.
> 
> This might become too big a discussion for this PR - take this out into a dedicated issue maybe?
> 
> I'm asking this from my usual clueless point of view - but my thinking would be: shouldn't the up-client/boundary-ustreamers adapters be able to perform name-to-id-to-name translation for the messages that flow by, with the relevant information retrieved from uDiscovery (DNS-style) services?



> 
> These uDiscovery entities, within a certain scope (i.e. vehicle), might then address the question of how to efficiently manage their translation tables as some form of distributed configuration, that can be managed as a separate problem to solve (so these tables might be just built from proto definitions, as default/semi-static config, but obviously some form of updatability might be desirable in the future)...

@AnotherDaniel , the problem is

evshary:
> > Another approach, of course, would be to restrict UAuthortity.name to only contain ASCII characters in the first place, if that is what you mean.
> Yes, lets restrict to ASCII characters

Sorry for disrupting the discussion. I also like the restriction of UAuthority.name, but I'm not sure what the range of ASCII characters here is. Does it only limit to use alphabet and number, or it can also use something like `!`, `+`, `@`, etc?
Also, is the UAuthority.name still case-insensitive?
These are related to whether the character are valid when transforming into protocol topic / key.

sophokles73:
> Sorry for disrupting the discussion. I also like the restriction of UAuthority.name, but I'm not sure what the range of ASCII characters here is. Does it only limit to use alphabet and number, or it can also use something like !, +, @, etc?

In the PR I am currently restricting the authority to be a _host_ following the corresponding production in https://datatracker.ietf.org/doc/html/rfc3986#appendix-A

In particular, this also includes characters from the _sub-delim_ production which also includes `!`, `&` etc

> Also, is the UAuthority.name still case-insensitive?

Yes, FMPOV that is the plan.

evshary:
> In the PR I am currently restricting the authority to be a _host_ following the corresponding production in https://datatracker.ietf.org/doc/html/rfc3986#appendix-A
> 
> In particular, this also includes characters from the _sub-delim_ production which also includes `!`, `&` etc

OK, then I still need to do some transformation while generating Zenoh key. At least, `$` and `*` are forbidden in Zenoh key.

> Yes, FMPOV that is the plan.

Then, no matter MQTT topic or Zenoh key should transform the authority to either lowercase or uppercase, since they are case-sensitive.

Is there any reason why we stick to RFC3986 standard here? If not, how about defining our own UAuthority for our own purpose?

sophokles73:
> Is there any reason why we stick to RFC3986 standard here? If not, how about defining our own UAuthority for our own purpose?

IMHO the biggest advantage is to be able to use tooling that is available for RFC3986 URIs, e.g. for parsing, comparing etc. Most programming languages have support for URIs in some way.

That said, as long as we make sure that a UUri _is-a_ RFC3986 URI, we should be on the safe side. This should indeed allow us to e.g. restrict the supported characters in the _host_ production if we feel that this would be beneficial. Do you have a concrete proposal for that?

evshary:
> IMHO the biggest advantage is to be able to use tooling that is available for RFC3986 URIs, e.g. for parsing, comparing etc. Most programming languages have support for URIs in some way.

Totally agree with you.

> Do you have a concrete proposal for that?

No, but I'm thinking perhaps using unserved here in UAuthority is enough.

```
unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
```

Also, limiting the using lowercase (or uppercase) alphabets here might be simpler.
But it's still ok to transform UAuthority into lowercase (or uppercase) in the transport layer (MQTT, Zenoh...).

stevenhartley:
@sophokles73 (et. al.),

I want to summarize conversations over email so we can conclude on this topic. I really don't think we will get a single UUri definition that will map perfectly to all transports so need to to compromise, taking into consideration the ease of use, $ for transmission, and other factors. 
Given that we will have to statically map service/topic information for mechatronics anyways (for SOME/IP), I think the "good enough" solution shall be:

- Go back to a single definition of a UUri that takes the best of long & short while removing micro all together
- Make UAuthority mandatory: Removes ambiguity if present or not and allows for simplifies normalization of UUris
- Constrain UAuthority::name and UEntity::name to "unreserved" defined above (and possibly limit size)
- Represent UResource as an ID in lieu of name+instance+Message strings: This reduces the size of the UUri while still providing uniqueness for the UResource and flexibility for UEntity naming to number mapping. We keep the definition of id ranges to what we've already defined (i.e. 0 is repsonse, 0-0x7FFFF is request, 0x8000- is publish/notification)

If we settle on above, there is no more need for isResolved(), isLong(0, isShort(), isMicro() and all the other validator logic that is overly complicated. Serializers are simplified back to a single toString() and fromString() and the string format is now:

`up://[UAUTHORITY_NAME]/[UENTITY_NAME]/[UENTITY_VERSION_MAJOR][.UENTITY_VERSION_MINOR]/[URESOURCE_ID]`

I've captured these changes in #126 . The reason for keeping UAuthority, UEntity, and UResource is to minimize the amount of changes in all the existing code and not to re-invent UUri completely.

sophokles73:
> ... and the string format is now:
>
> `up://[UAUTHORITY_NAME]/[UENTITY_NAME]/[UENTITY_VERSION_MAJOR][.UENTITY_VERSION_MINOR]/[URESOURCE_ID]`

So far, the minor version had not been part of any of UUris serializations. Just want to make sure that we really need it in there. And if we do, I guess the inclusion of minor version should be dependent on the presence of major version

`up://[UAUTHORITY_NAME]/[UENTITY_NAME]/[UENTITY_VERSION_MAJOR[.UENTITY_VERSION_MINOR]]/[URESOURCE_ID]`

or be put into a separate path segment

`up://[UAUTHORITY_NAME]/[UENTITY_NAME]/[UENTITY_VERSION_MAJOR/[UENTITY_VERSION_MINOR]/[URESOURCE_ID]`

However, I'd prefer the former, if we really need minor version in it ...

sophokles73:
@stevenhartley I have adapted the UUri spec according to the discussed simplifications of the UUri data model. Some things I want to bring to your attention:

* I propose to use base16 encoding for the identifiers in the URI which makes them easier to read FMPOV and will also save a few bytes compared to decimal encoding
* The _host_ production used for the authority name has not been adapted (yet) to use only _unreserved_ characters as proposed by @evshary. We might want to consider this, though.
* With the introduction of explicit wildcard values we no longer support URIs having empty path components. FMPOV this is actually good because it makes parsing much easier and should not have any real impact on message exchanges because the wildcards will only be used in the definition of forwarding rules in a streamer.
* I have added OpenFastTrace specification item markup where appropriate. We can use these for testing the OpenFastTrace toolchain before rolling out to other spec documents.

stevenhartley:
> @stevenhartley I have adapted the UUri spec according to the discussed simplifications of the UUri data model. Some things I want to bring to your attention:
> 
> * I propose to use base16 encoding for the identifiers in the URI which makes them easier to read FMPOV and will also save a few bytes compared to decimal encoding
> * The _host_ production used for the authority name has not been adapted (yet) to use only _unreserved_ characters as proposed by @evshary. We might want to consider this, though.
> * With the introduction of explicit wildcard values we no longer support URIs having empty path components. FMPOV this is actually good because it makes parsing much easier and should not have any real impact on message exchanges because the wildcards will only be used in the definition of forwarding rules in a streamer.
> * I have added OpenFastTrace specification item markup where appropriate. We can use these for testing the OpenFastTrace toolchain before rolling out to other spec documents.

I'll take a look. I started to code these things on the 1.5.8 branch of up-java and the serializers and deserializers were drastically simplified as well.

basics/uri.drawio.svg
stevenhartley, line 1: You removed the white background so it is difficult to see the image, please add it back.

basics/uri.adoc
```diff
@@ -25,404 +26,522 @@ specific language governing permissions and limitations
...
-| Returns true if the object contains both names and numbers (ids) such that it can be serialize to both <<Long URIs>> and <<Micro URIs>>
+[.specitem,oft-sid="dsn~authority-name-length~1",oft-needs="impl,utest"]
+--
+A UUri's `authority_name` *MUST NOT* exceed 128 characters in length.
 
-| `isMicroForm(UUri): boolean`
-| Returns true if the Uri part contains the required ids to serialized to Micro format and the fields of the Uri can fit within the specified number of bits and bytes. Both sets of details can be obtained under <<Micro URIs>>
+[source,ocl]
+----
+inv: authority_name.length() <= 128
```
stevenhartley, line 96: Where did this requirement come from?

basics/uri.adoc
```diff
@@ -25,404 +26,522 @@ specific language governing permissions and limitations
...
+NOTE: uEntities which produce events for consumption by other uEntities assume a _Service_ role. +
+uEntities which consume events produced by other uEntities assume an _Application_ role. +
+uEntities *MAY* assume both the Service and Application roles.
 
-|===
+[.specitem,oft-sid="dsn~entity-id~1"]
+--
+A UUri's `ue_id` property value determines the type and instance of service being referred to. +
+The value's _most_ significant 16 bits represent the _service instance_. +
+The value's _least_ significant 16 bits represent the _service type_.
```
stevenhartley, line 113: identifier and not type

basics/uri.adoc
```diff
@@ -25,404 +26,522 @@ specific language governing permissions and limitations
...
-| Returns true if: `id==0  && name==rpc && instance==response`
+NOTE: uEntities which produce events for consumption by other uEntities assume a _Service_ role. +
+uEntities which consume events produced by other uEntities assume an _Application_ role. +
+uEntities *MAY* assume both the Service and Application roles.
 
-|===
+[.specitem,oft-sid="dsn~entity-id~1"]
+--
+A UUri's `ue_id` property value determines the type and instance of service being referred to. +
+The value's _most_ significant 16 bits represent the _service instance_. +
```
stevenhartley, line 112: We should put a note that a service instance of 0 means "don't care".

basics/uri.adoc
```diff
@@ -25,404 +26,522 @@ specific language governing permissions and limitations
...
 
+A UUri can be used to define a _pattern_ that other UUris can then be matched against.
 
-.Remote IPv4 Uri
-[width="80%",cols="30%,70%",options="header"]
-|===
-|Format | Value
+[.specitem,oft-sid="dsn~pattern-matching~1",oft-needs="impl,utest"]
+--
+A UUri that is used as a pattern to match other UUris against
```
stevenhartley, line 398: This is duplicated from the sentence above

basics/uri.adoc
```diff
@@ -25,404 +26,522 @@ specific language governing permissions and limitations
...
 
-|Long
-|`//vcu.veh.gm.com/core.usubscription/3/SubscriptionChange#Update`
+* *MAY* have its `authority_name` set to the `*` (`U+002A`, Asterisk) character in order to match _any_ (including _no_) authority,
+* *MAY* have the _service type_ part of its `ue_id` set to `0xFFFF` in order to match _any_ type,
```
stevenhartley, line 401: maybe we need to call it service number to differentiate the instance id?

basics/uri.adoc
```diff
@@ -25,404 +26,522 @@ specific language governing permissions and limitations
...
+* *MAY* have the _service instance_ part of its `ue_id` set to `0x0000` in order to match _any_ instance,
```
stevenhartley, line 402: I like this!! better than what I was thinking and coding in Java...

stevenhartley (changes requested):
@sophokles73 only minor comments left but looking much better than before!

basics/uri.adoc
sophokles73: We had been discussing that we wanted to limit the length of the authority name to some degree. I just picked 128 as a first shot. Do you still think that we should limit the length?

basics/uri.adoc
sophokles73: We usually talk about _instances_ of a certain _type_, don't we? A service _identifier_ to me suggests that we are talking about a particular instance already, instead of its type.

basics/uri.adoc
sophokles73: I have added that to the section on pattern matching ...

basics/uri.drawio.svg
sophokles73: I agree that this makes it harder to see in vscode (with a dark theme/background). However, when viewing on GitHub this is not an issue, right? The rfc3986.png also has a transparent background so I thought we might want to do the same for the URI diagram ...

basics/uri.adoc
sophokles73: not sure what you mean, replace _service type_ with _service number_ or replace _service instance_ with _service number_?

sophokles73:
@stevenhartley I have pushed changes and amended your comments. Would you mind taking another look?

basics/uri.adoc
stevenhartley: Re-reading some/ip spec for a little inspiration, they have service_id and service_instance_id, no where do we refer to service_id as type hence why I wanted to keep the id or number but not type.

stevenhartley:
@sophokles73 not sure if my comment got lost but the diagram needs to have a white background (it was removed) otherwise I cannot view it when using dark mode in github.

basics/uri.adoc
sophokles73: then let's stick to _service id_ and service instance id_. I'll update accordingly ...

sophokles73:
@stevenhartley I have added the white background and updated the terminology used for ue_id

stevenhartley (approved):
LGTM, thanks @sophokles73 !

sophokles73:
@stevenhartley any idea why the `verify-pr` check is not completing? Or what it actually does/is, for that matter?

stevenhartley:
> @stevenhartley any idea why the `verify-pr` check is not completing? Or what it actually does/is, for that matter?

It builds the proto to make sure there are no issues with syntax.

stevenhartley (approved):

sophokles73:
@stevenhartley this PR seems to be stuck, I have tried to push some changes but that doesn't seem to resolve the situation. I will now close this PR and create a new one instead ...

sophokles73:
Reopening to test a theory ...

PR #123, opened 2024-04-25
ShortUUri Definition

stevenhartley:
The following works to address #121 by making the contract explicit and to differentiate the old LongForm URI and short form URI.

### DIFF
--- a/up-core-api/uprotocol/uri.proto
+++ b/up-core-api/uprotocol/uri.proto
@@ -58,11 +58,6 @@ message UUri {
 // is local.
 message UAuthority {
   optional string name = 1; // Domain & device name as a string
-  oneof number {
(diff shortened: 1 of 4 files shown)


up-core-api/uprotocol/uri.proto
```diff
@@ -94,11 +88,36 @@ message UResource {
...
 
 // Repeated list of URIs that can be used for batching APIs
 message UUriBatch {
   repeated UUri uris = 1; // Repeated list of URIs
 }
+
+// Short Form uProtocol <b>URI</b>.
+// Condensed version of the UUri that uses numbers (ids) in lieu of names to  
+// uniquely identify devices, services, and resources on the  network.<br>
+message ShortUUri {
```
sophokles73, line 102: what's the reason for adding ShortUUri instead of changing the existing UUri?

PR #126, opened 2024-04-30
Simplification of UUri Data model

stevenhartley:
The following adds to #121 to simplify the datamodel for UUri object.

### DIFF
--- a/up-core-api/uprotocol/uri.proto
+++ b/up-core-api/uprotocol/uri.proto
@@ -54,47 +54,31 @@ message UUri {
 // An Authority represents the deployment location of a specific Software Entity.
 // Authority can be represented in either a name (i.e example.com), ip address (205.236.147.1)
 // or an ID (i.e. VIN, SHA 128, or any other identifier that is less than 255 bytes.
-// *NOTE:* When Authority is empty (neither name, ip, or id are set) the authority
-// is local.
 message UAuthority {
-  optional string name = 1; // Domain & device name as a string
-  oneof number {
-    bytes ip = 2; 	// IPv4 or IPv6 Address in byte format
-    bytes id = 3;		// Unique ID for the device, could be a VIN, SHA 128, or any other identifier
-                    // *NOTE:* MAX length is 255 bytes
-  }
+  string name = 1; // Domain & device name as a string
 }
 
 
 // Data representation of an <b> Software Entity - uE</b><br>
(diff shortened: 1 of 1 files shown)


up-core-api/uprotocol/uri.proto
```diff
@@ -54,46 +54,30 @@ message UUri {
...
-// *NOTE:* When Authority is empty (neither name, ip, or id are set) the authority
-// is local.
 message UAuthority {
-  optional string name = 1; // Domain & device name as a string
-  oneof number {
-    bytes ip = 2; 	// IPv4 or IPv6 Address in byte format
-    bytes id = 3;		// Unique ID for the device, could be a VIN, SHA 128, or any other identifier
-                    // *NOTE:* MAX length is 255 bytes
-  }
+  string name = 1; // Domain & device name as a string
```
sophokles73, line 58: FMPOV this would need to remain `optional` in order to consistently support matching _any_ value across UUri's components.

Alternatively, we could think about defining explicit wildcard characters/values ...

up-core-api/uprotocol/uri.proto
```diff
@@ -54,46 +54,30 @@ message UUri {
...
 // A uE that publishes events is a <b>Service</b> role.<br>
 // A uE that consumes events is an <b>Application</b> role.
 message UEntity {
-  optional string name = 1;          // Name of the entity
-  optional uint32 id = 2;            // The numeric ID for the uEntity assigned from the name/number registry
-  optional uint32 version_major = 3; // optional major version of the uEntity
-  optional uint32 version_minor = 4; // optional minor version of the uEntity
+  optional uint32 id = 1;            // UEntity ID
+  optional uint32 version_major = 2; // optional major version of the uEntity
+  optional uint32 version_minor = 3; // optional minor version of the uEntity
```
sophokles73, line 72: I haven't seen any place yet where minor version is being used. Is this really needed?

up-core-api/uprotocol/uri.proto
stevenhartley: We discussed this and we are using explicit value definitions so there is no optional

up-core-api/uprotocol/uri.proto
stevenhartley: I have removed it. Minor is really only needed at the application layer.

sophokles73 (approved):
LGTM

Issue #128, opened 2024-05-01
Remove or replace the concept of references in UPayload


gregmedd:
### The problem

In [upayload.proto](https://github.com/eclipse-uprotocol/up-spec/blob/4a04ce999026a2dfccb060742d8c02d980603d65/up-core-api/uprotocol/upayload.proto), the `UPayload` object supports two forms of data: _value_ and _reference_. The _value_ form copies the payload data into the `UPayload` object, while the _reference_ form stores a raw pointer as a 64-bit integer.

There are lots of problems with the concept of passing a raw pointer in a payload. For one, it is very hard to guarantee any data safety with a raw pointer. And it would significantly complicate transport implementations - when would they need to convert references to copies and when would they not?

For data safety, we would need to be able to make these guarantees:

- The pointer MUST remain valid for the entire life of all copies of the `UPayload`.
- The pointer MUST refer to the same memory location for all recipients of the `UPayload`.
- The data referenced MUST NOT be altered while being accessed through a `UPayload`.
- There MUST be a signal to the data provider when all copies of `UPayload` are done with the pointer so the memory can be altered or released.

I see no clear way for any uProtocol implementation to provide those data safety guarantees without extending the `UPayload` object, and even then that is unlikely to cover all intended use cases.

Additionally, many transport implementations may have their own shared memory systems built-in that are not compatible with this separate pointer concept. For example, Zenoh can automatically use shared memory for local exchanges and multicast when publishing across the network.

For other scenarios, such as data in DMA (e.g. via [dma-buf](https://docs.kernel.org/driver-api/dma-buf.html)), significantly more than just a pointer would be needed in the message. There would need to be a flag that the data is host-local (or that it needs to be copied when exiting the host), and parameters for mapping the correct memory blocks would have to be added.

I've included three proposals for next steps below. Admittedly, I haven't spent a lot of time thinking on the problem, so I would appreciate alternate ideas on how to proceed.

----

### Proposal 1: Handle references at a higher layer

Remove the reference concept and add message scoping to allow layers above uProtocol to define their own memory sharing schemes. This would involve these steps:

- Remove the `reference` field from `UPayload`
- Add a `scope` or `locality` field to `UAttributes` that allows for a message to be restricted to scopes like `process` or `host`
[...]

PR #130, opened 2024-05-02
Cleanup of UPayload

stevenhartley:
The following change removes the notion of passing data by reference in UPayload till we finalize the shared memory APIs.

#128


gregmedd (approved):
Looks good to me

gregmedd (approved):
That was supposed to be an approval 😅

Still looks good to me

sophokles73 (approved):
LGTM

Issue #132, opened 2024-05-04
Remove CallOptions from uattributes.proto


gregmedd:
The `CallOptions` message does not get included in any message or sent on the wire. It appears to be intended as a data wrapper for uE code interacting with the Layer 2 client API.

Since it is not a part of the actual message data model ("Layer 0"), it should be removed in favor of allowing language-specific APIs to implement the best fit for their language.

stevenhartley:
Although your statement is true, I want to know why we couldn't define it here?

gregmedd:
Because interacting with protobuf objects is _really unpleasant_ in situations where it is not required. It is simpler to provide a context-specific interface that takes the appropriate parameters for that situation. See `UMessageBuilder` in `up-rust`. We are mirroring that API in the `up-cpp` updates.

Protobuf messages intended for use as purely internal APIs are only a hindrance to defining idiomatic language APIs. This goes back to the L2 update where the required methods are listed, but their signatures are left to the language-specific APIs to define.

sophokles73:
merged in https://github.com/eclipse-uprotocol/up-spec/pull/133

sophokles73:
Need to reopen because the spec has not been updated yet ...

sophokles73:
be my guest 😁

stevenhartley:
@sophokles73 we are rewriting rpcclient.adoc anyways for the new high level L2 API specifications, shouldn't we do it at that point? The only reference to CallOptions are in that file.

stevenhartley:
Closing, we will resolve reference in the L2 PR

PR #133, opened 2024-05-04
Remove the CallOptions message

gregmedd:
CallOptions is never sent as a uProtocol message. Instead of being a layer-0 message, it was used as a layer-2 client API.

It is being removed in favor of allowing language-specific API libraries to define the most appropriate function signatures for the language.

Addresses issue #132

### DIFF
--- a/up-core-api/uprotocol/uattributes.proto
+++ b/up-core-api/uprotocol/uattributes.proto
@@ -74,19 +74,6 @@ message UAttributes {
 }
 
 
-// A subset of UAttributes used for RPC method invocation.
-message CallOptions {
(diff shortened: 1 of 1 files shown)


sophokles73 (commented):
Removing the message from the proto file is ok FMPOV.
However, we should then also remove all references to `CallOptions` throughout the rest of the spec, e.g. from `rpcclient.adoc`

stevenhartley (approved):
@sophokles73 & @sophokles73 we will refactor the L2 documentation on a separate PR so I'm merging this in so I can pull it on my up-java 1.5.8 branch

PR #134, opened 2024-05-06
Update UTransport interface specification

### LINKED ISSUE #73: Does UTransport::register_listener match on source or sink address?
The [RegisterListener() spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/README.adoc#22-registerlistener) does not explicitly define the (type of) address to match on.

> Register a UListener to receive message(s) for a given UUri (topic).

seems to suggest that the given topic should be matched on a message's destination address (*sink* attribute). However, that would mean that *publish* messages cannot be matched because they only contain an origin address (*source* attribute).

sophokles73:
Changed registerListener method signature to support
matching on message source and sink addresses by means
of UUri patterns.

This change allows UTransport to be used by higher level
APIs for implementing all message exchange patterns
supported by uProtocol.

Fixes #73

### DIFF
--- a/upclient.adoc
+++ b/upclient.adoc
@@ -20,7 +20,7 @@ SPDX-License-Identifier: Apache-2.0
 
 == Overview
 
-A uProtocol Client library is the application facing library that uEntities use to interact with other uEntities. The library contains a programminb language and transport protocol specific implementation of the low level link:up-l1/README.adoc[Transport Layer API] and the link:up-l2/rpcclient.adoc[Communication Layer API]. The library also includes implementation specific client-side implementations of the core uServices uDiscovery, uTwin, and uSubscription.
+A uProtocol Client library is the application facing library that uEntities use to interact with other uEntities. The library contains a programming language and transport protocol specific implementation of the low level link:up-l1/README.adoc[Transport Layer API] and the link:up-l2/rpcclient.adoc[Communication Layer API]. The library also includes implementation specific client-side implementations of the core uServices uDiscovery, uTwin, and uSubscription.
 
 == Requirements
 


up-l1/README.adoc
```diff
@@ -50,157 +42,218 @@ We will use the following terminology during this document:
...
 
-Implements the _pull_ <<Delivery Method>> to fetch a message from the transport for a given UUri (topic).
+Clients use this method to receive a single message matching given filter criteria.
+This method implements the _pull_ <<Delivery Method>> for the underlying messaging infrastructure.
 
-`OUT <<UStatus>> receive(IN UUri, OUT UPayload&, OUT UAttributes&)`
+[source]
+----
+receive(
+  IN sourceFilter: UUri,
```
stevenhartley, line 219: Is the new line for each attribute like this a standard for defining APIs?

up-l1/README.adoc
```diff
@@ -50,157 +42,218 @@ We will use the following terminology during this document:
...
+| A succeeded future if the listener has been unregistered successfully. +
+Otherwise, a failed future containing a `UStatus` indicating the reason for the failure.
 |===
 
-* *MUST* pass a UUri for which `isMicroForm()` returns true
-* *SHOULD* pass a Resolved UUri (contains both names and numbers)
-
 
-=== Receive()
+=== Receive
```
stevenhartley, line 211: The receive API is a non-blocking synchronous API that returns immediately if there are no messages to consume. I also noticed that I didn't explicitly state that this API is optional to implement only if the transport utilizes pull delivery method.

up-l1/README.adoc
```diff
@@ -50,157 +42,218 @@ We will use the following terminology during this document:
...
- * *MUST* be passed by reference
+| result
+| Future<Void, link:../basics/error_model.adoc[UStatus]>
+| A succeeded future containing the least recent message that matches the given filter criteria and has not expired yet. +
+A failed future if no message can be retrieved from the messaging infrastructure. The `UStatus` will contain a `UCode` indicating the reason for the failure.
 
 |===
 
- * *MUST* return `NOT_FOUND` if there are no messages for the given topic
+[oft-sid="req~utransport-register-listener~1"]
```
stevenhartley, line 245: I would like to discuss adding this metadata to the rest of the specs, was this one added to test out the links and then if we agree we would add them to all requirements? How do we link this in our code?

up-l2/rpcclient.adoc
```diff
@@ -39,10 +39,10 @@ SPDX-License-Identifier: Apache-2.0
 
 | `IN` | Input parameter
 | `OUT`| Output parameter
```
stevenhartley, line 41: OUT was meant to be a passed parameter that is an output and not the return value from a method but you prefer to refer to IN as a parameter and out as the return value from the function?

stevenhartley (changes requested):
see comments

up-l2/rpcclient.adoc
sophokles73: To be honest: I have no clue. The semantics of IN, OUT, IN/OUT are, at best, unclear to me.

I had already thought about proposing to use an IDL that has some defined semantics instead of _pseudo code_. Maybe using UML2 semantics or just using Java Interfaces and Methods?

up-l1/README.adoc
sophokles73: Adding these might indeed be a little premature. Ideally, they would serve as OpenFastTrace specification item IDs which could then be included in source code, tests etc to provide for traceability across all artifacts. I haven't yet tested this, I have to admit. In fact, OFT does not (yet) support AsciiDoc files. However, from what I can tell, it should not be too hard to add this to OpenFastTrace.

up-l1/README.adoc
sophokles73: Ok, makes sense. Conversely, does this also mean that the un-/register_listener methods are optional to implement only if the transport utilizes push delivery?

up-l1/README.adoc
sophokles73: I just thought that it would improve readability because you could at once see which are the inputs and the outputs.

up-l1/README.adoc
```diff
@@ -50,157 +42,218 @@ We will use the following terminology during this document:
...
 
-The API signature:
+Clients use this method to transmit a single message.
 
-`OUT UStatus send(UMessage&)`
+[source]
+----
+send(
+  IN message: UMessage&,
+  OUT result: Future<Void, UStatus>
```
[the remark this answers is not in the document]
sophokles73: It is in the Rust library :-)
I wonder if we should actually bother with sync vs async in this spec at all or not leave this up to the language library developers instead.

up-l1/README.adoc
```diff
@@ -50,157 +42,218 @@ We will use the following terminology during this document:
...
-| UProtocol message containing the header (UAttributes) and payload (UPayload)
+|Parameter | Type | Description
 
- * *MUST* be passed by reference
+| message
+| link:../basics/umessage.adoc[UMessage]
+| The message to send.
 
+| result
+| Future<Void, link:../basics/error_model.adoc[UStatus]>
```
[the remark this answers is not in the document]
sophokles73: I wonder why I would get back a UStatus instead of Void if sending the message succeeded? FMPOV this would result in a pretty clunky developer experience which would require me to always explicitly check for the UCode to be 0, or am I mistaken?

up-l1/README.adoc
```diff
@@ -50,157 +42,218 @@ We will use the following terminology during this document:
...
+This API is used to implement the _push_ <<Delivery Method>> for the underlying messaging infrastructure.
+After this method has completed successfully, the given listener will be invoked for each message that matches the given source and sink filter patterns according to the rules defined by the link:../basics/uri.adoc[UUri specification].
 
+[source]
+----
+registerListener(
+  IN sourceFilter: UUri,
+  IN sinkFilter: UUri?,
+  IN listener: UListener&,
+  OUT result: Future<Void, UStatus>
```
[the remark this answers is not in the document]
sophokles73: again, in the Rust library, this returns a Future in order to take advantage of async transport client libraries.

up-l1/README.adoc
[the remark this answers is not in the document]
stevenhartley: Clunky yes but simpler than checking if it is void or UStatus etc...

up-l1/README.adoc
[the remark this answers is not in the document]
stevenhartley: Transport I think we need to be explicit as we said historically that send() returning that the message has been delivered to the "next hop". It is only the transport layer acknowledgement and not the communication layer (ex. rpc or pub/sub) acknowledgement.

up-l1/README.adoc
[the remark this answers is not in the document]
stevenhartley: per previous message, the sending of the message is synchronous, not the delivery to the end destination.

up-l1/README.adoc
stevenhartley: Yes true

up-l1/README.adoc
stevenhartley: I would love that we can do this one a single file (ex. the uri.adoc) and see if this can work end-2-end before we start adding the tags everywhere and then have to remove/change them later.

up-l1/README.adoc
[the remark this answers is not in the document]
sophokles73: IMHO this
```java
try {
  transport.send(myMessage);
} catch (SendException e) {
  log.error(e);
}
```

is much preferable over this
```java
var result = transport.send(myMessage);
if (result.isError()) {
  log.error("failed to send message: {}", result.getCode());
} else {
  // proceed
}
```
This is also true for Rust and I guess every other language that has an exception mechanism. Using that programming language's exception mechanism seems very desirable to me ...

up-l1/README.adoc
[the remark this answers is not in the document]
sophokles73: > Transport I think we need to be explicit as we said historically that send() returning that the message has been delivered to the "next hop". It is only the transport layer acknowledgement and not the communication layer (ex. rpc or pub/sub) acknowledgement.

And I think that this is the important thing to make clear. Whether an implementation implements the delivery to the _next hop_ using a blocking/synchronous call or a non-blocking/asynchronous call does not really make a difference regarding the delivery semantics/guarantees, does it?

up-l1/README.adoc
sophokles73: Yes, that is what I have in mind as well. I have spent some time investigating OpenFastTrace's extension API today. Looks very doable ...

up-l1/README.adoc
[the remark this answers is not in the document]
stevenhartley: Whether it uses blocking or not underneath uTransport, the API itself has to be consistent, otherwise we cannot build a generic uStreamer that behaves the same regardless of the underlining transport implementation.

up-l1/README.adoc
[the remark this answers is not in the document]
PLeVasseur: Perhaps a bit of my Rust bias creeping in here, but we have language-level mechanisms to make discriminating success-vs-failure cases in `Result<T, E>`.

### Modeled with explicit success and failure at the type level:
```rust
fn register_listener(
  sourceFilter: UUri,
  sinkFilter: UUri?,
  listener: UListener&
) -> Result<(), UStatus>;
```
allows us to easily:
```rust
let registration_result = client.register_listener(source_filter, sink_filter, listener);

// use pattern matching
match registration_result {
  Err(ustatus) => { /* handle error case with access to `code` and `message` */ }
  Ok(()) => { /* handle success */ }
}

// use if-let pattern
if let Err(ustatus) = registration_result {
  /* handle just the error condition inside of ustatus */
}
```

### Modeled without explicit success and failure at the type level:
```rust
fn register_listener(
  sourceFilter: UUri,
  sinkFilter: UUri?,
  listener: UListener&
) -> UStatus;
```
we then need to:
```rust
let registration_result = client.register_listener(source_filter, sink_filter, listener);

// unpack the UCode and message to then match on and use both
let (code, message) = (registration_result.get_code(), registration_result.get_message());
match code {
  UCode::OK => { /* handle the success case */ }
   _ => { /* handle the remaining cases of `code` and also use `message` */ }
}

// unable to use if-let, must fall back to an unpack and if condition
let (code, message) = (registration_result.get_code(), registration_result.get_message());
if code != UCode::OK {
  /* handle the cases where not a success using `code` and `message` */
}
```

Could we ask for other language implementors to weigh in here with what this would look like?

up-l1/README.adoc
```diff
@@ -50,157 +42,218 @@ We will use the following terminology during this document:
...
-* *MUST* pass UUris for which `isMicroForm()` return true
-* *SHOULD* pass Resolved UUris (contains both names and numbers)
-
-
-
-=== RegisterListener()
+Communication protocols (a.k.a. uProtocol Transports) define their own Protocol Data Unit (PDU) that comprises of header and payload. Some transports header parameters map already to uProtocol UAttributes. Sending UAttributes in the transport's payload as well as in the header increases overhead for little to no gain. To address this issue,  <<Send()>> allows flexibility for uTransport implementers to define however they see fit what gets mapped into their transport header vs payload. 
 
-Register a UListener to receive message(s) for a given UUri (topic). This API is used to implement the _push_ <<Delivery Method>>.
+For example, if HartleyTransport can map only UPriority to its PDU header, then HartleyTransport would define a PDU payload type that includes UPayload and all the other UAttributes and send that new type in its PDU payload such that the receiver will not loose any metadata.
```
PLeVasseur, line 113:
```suggestion
For example, if HartleyTransport can map only UPriority to its PDU header, then HartleyTransport would define a PDU payload type that includes UPayload and all the other UAttributes and send that new type in its PDU payload such that the receiver will not lose any metadata. 
```

up-l1/README.adoc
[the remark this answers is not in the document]
PLeVasseur: Agreed, we do need consistency for ability to plug into a uStreamer.

Could we have all the methods of a `UTransport` be asynchronous, giving a bit more control over it to the implementor?

For the record, the `up-streamer-rust` implementation in PR uses the `up-rust` `UTransport` definition which has all async methods.

sophokles73:
@stevenhartley @PLeVasseur I have pushed changes and have amended your comments ...

matthewd0123:
Question/comment. Can/should we also align the error messages returned as well? Good that we're aligning the UCodes, but would ask if it's worth it to align the UStatus messages as well between implementations. Benefit would be consistency, but not sure if it's too much micromanagement.

sophokles73:
> Question/comment. Can/should we also align the error messages returned as well? Good that we're aligning the UCodes, but would ask if it's worth it to align the UStatus messages as well between implementations. Benefit would be consistency, but not sure if it's too much micromanagement.

FMPOV relying on error message format/content is an antipattern that leads to code that is hard to maintain and will produce hard to find bugs during runtime. So, no, I do not think that we should standardize error messages ...

up-l1/README.adoc
```diff
@@ -28,179 +28,199 @@ SPDX-FileType: DOCUMENTATION
...
-
- * *MUST* be passed by reference
+| message
+| link:../basics/umessage.adoc[UMessage]
+| The message to send.
 
+| result
+| Future<Void>
+| A succeeded future if the message has been sent successfully. This means that the underlying messaging infrastructure has accepted the message for delivery. It does *not* (necessarily) mean that the message has reached its destination (yet). +
+Otherwise, a failed future which contains a `UCode` indicating the reason for the failure.
```
PLeVasseur, line 92: Does this:
```
Otherwise, a failed future which contains a `UCode` indicating the reason for the failure.
```
mean that the [`send` signature](https://github.com/eclipse-uprotocol/up-spec/pull/134/files#diff-a034e4666a20b9cc4b66d777b10e9b4ffac86a7646a7dc79d13c7d067987436fR76) should read:
```
send(message: UMessage) : Future<Void, UCode>
```
?

sophokles73: In Java, for example, the Future<T> class is parameterized with the type of successful completion only because the Future will also be failed with an Exception. Rust is the only language I know of where any type can be used to represent the erroneous outcome.
FMPOV it should be sufficient to define that the Future will be failed with _something_ that contains the relevant information (the UCode in this case). In the Java case this probably means that the Exception contains a UCode typed field. In Rust it might mean that the error type is UCode. Does that make sense?

PLeVasseur: Thanks for the context. Makes sense.

up-l1/README.adoc
```diff
@@ -87,9 +73,7 @@ Clients use this method to transmit a single message.
 
 [source]
 ----
-send(
-  IN message: UMessage&,
-  OUT result: Future<Void, UStatus>
+send(message: UMessage) : Future<Void>
 )
```
PLeVasseur, line 77: Left over, extra `)`?

up-l1/README.adoc
```diff
@@ -154,20 +130,18 @@ registerListener(
...
 | A succeeded future if the listener has been registered successfully. +
-Otherwise, a failed future containing a `UStatus` indicating the reason for the failure.
+Otherwise, a failed future which contains a `UCode` indicating the reason for the failure.
 |===
 
-[oft-sid="req~utransport-register-listener~1"]
---
 Implementations
 
+* *MAY* fail invocations with a `UCode::UNIMPLEMENTED` if the transport does not support the _push_ <<_delivery_method>>. In that case, the <<_unregisterlistener>> method *MUST* also fail accordingly.
```
PLeVasseur, line 140: Something about the `_` gets mangled or something...?

image

up-l1/README.adoc
```diff
@@ -202,11 +171,15 @@ unregisterListener(
...
-| Future<Void, link:../basics/error_model.adoc[UStatus]>
+| Future<Void>
 | A succeeded future if the listener has been unregistered successfully. +
-Otherwise, a failed future containing a `UStatus` indicating the reason for the failure.
+Otherwise, a failed future which contains a `UCode` indicating the reason for the failure.
 |===
 
+Implementations
+
+* *MAY* fail invocations with a `UCode::UNIMPLEMENTED` if the transport does not support the _push_ <<_delivery_method>>. In that case, the <<_registerlistener>> method *MUST* also fail accordingly.
```
PLeVasseur, line 181: Same thing appears to be happening as [here](https://github.com/eclipse-uprotocol/up-spec/pull/134/commits/03d54831bc214ee82e6fe2ee2895ed387a8ef02a#r1602292080).

up-l1/README.adoc
```diff
@@ -236,18 +205,16 @@ receive(
...
-A failed future if no message can be retrieved from the messaging infrastructure. The `UStatus` will contain a `UCode` indicating the reason for the failure.
+Otherwise, a failed future which contains a `UCode` indicating the reason for the failure.
 
 |===
 
-[oft-sid="req~utransport-register-listener~1"]
---
 Implementations
 
+* *MAY* fail invocations with a `UCode::UNIMPLEMENTED` if the transport does not support the _pull_ <<_delivery_method>>
```
PLeVasseur, line 216: Same thing appears to be happening as [here](https://github.com/eclipse-uprotocol/up-spec/pull/134/commits/03d54831bc214ee82e6fe2ee2895ed387a8ef02a#r1602292080).

up-l1/README.adoc
```diff
@@ -236,18 +205,16 @@ receive(
...
 * *MUST* fail with a `UCode::NOT_FOUND` if there are no matching messages available
---
 
 === Message Delivery
```
PLeVasseur, line 220: Can't quite leave a comment that far down, so:

```
* Transport *MUST* support retransmission of messages that are no able to be delivered to the next-hop
```
=>
```
* Transport *MUST* support retransmission of messages that are not able to be delivered to the next-hop
```

up-l1/README.adoc
```diff
@@ -215,11 +188,7 @@ This method implements the _pull_ <<Delivery Method>> for the underlying messagi
 
 [source]
 ----
-receive(
-  IN sourceFilter: UUri,
-  IN sinkFilter: UUri?,
-  OUT result: Future<UMessage&, UStatus>
-)
+receive(sourceFilter: UUri, sinkFilter: UUri [0..1]) : Future<UMessage>
```
PLeVasseur, line 191: Here's a thought -- I think for _most_ folks, they will actually want to block here.

Maybe we can have a short snippet like --

> If you wish to treat this function as synchronous, you may use your language of choice's mechanism for doing so. For example, Java's `CompleteableFuture::complete()` or Rust's `.await`.

Dunno. Too hand-holdy?

@stevenhartley -- sanity check me on the Java point :laughing:

up-l1/README.adoc
sophokles73: That's weird, the vscode preview correctly renders the links and it is the standard AsciiDoc format. Anyway, I will replace it with the (linked to) section title.

up-l1/README.adoc
sophokles73: > Here's a thought -- I think for most folks, they will actually want to block here.

FMPOV the important thing about `receive` is, that it should fail _immediately_ if no message is available, i.e. it should not wait for a message becoming available.

It is not really about whether a particular implementation actually uses async or sync calls. However, IMHO a UTransport implementation should follow a consistent approach across all methods, i.e. either sync or async. If client code needs to be able to deal with async methods anyway, because all of the other methods are async, then I do not see much value in this particular method being the only one defined as sync.

In general, when it comes to IO operations (and UTransport is as close to IO as it gets), using async non-blocking calls has been proven to make systems more performant, scalable and less resource consuming. This also matches my personal experience with other (communication infrastructure) projects ...

up-l1/README.adoc
sophokles73: That whole section needs to be rewritten FMPOV. I'll give it a shot ...

up-l1/README.adoc
sophokles73: Maybe we should, instead of using the Future<T> return type, just declare the concrete return type (T) and then also declare an Exception being thrown. Would that help to make it easier to understand? We could then also add some general wording to a section called _Implementation Hints_ or similar, where we advise implementers to use non-blocking async calls. WDYT? @PLeVasseur @stevenhartley

sophokles73:
@stevenhartley @PLeVasseur I have pushed changes as discussed in the call.
Would you mind taking another look?

@tamarafischer I have tried to add some wording to the expected outcome of the send() method. WDYT?

up-l1/README.adoc
PLeVasseur: Please view this as non-blocking for merge. It's a minor point that could be discussed later; more a tip to help a developer, but may just as easily fit in with example code we may write later.

up-l1/README.adoc
PLeVasseur: I like your idea of having a tips & tricks section for implementers in which to put some of this advice. I still think having some level of flexibility _even_ at uP-L1 for languages to write it the most reasonable way makes sense, which is where such a section could bring further value IMHO.

PLeVasseur:
> Would you mind taking another look?

Changes to the section we discussed on the call look good to me!

up-l1/README.adoc
```diff
@@ -18,198 +18,314 @@ SPDX-FileType: DOCUMENTATION
...
-| Topic to register the listener for
+| sourceFilter
+| link:../basics/uri.adoc[UUri]
+| The _source_ address pattern that messages need to match.
 
-| <<UListener>>
-| Listener to be registered
+| sinkFilter
+| link:../basics/uri.adoc[UUri]
+| The _sink_ address pattern that messages need to match. If omitted, a message must not contain any sink address in order to match.
```
stevenhartley, line 160:
```suggestion
| The _sink_ address pattern that messages need to match. If omitted, a message **MUST NOT** contain any sink address in order to match.
```

up-l2/rpcclient.adoc
```diff
@@ -41,21 +41,38 @@ SPDX-License-Identifier: Apache-2.0
...
-| `UUri`
-| Method URI that is being invoked
+| options
+| CallOptions
+| Additional meta data to include in the request message.
 
-| `CallOptions`
-| Set of attributes that can be sent as part of a method invocation
+| payload
+| link:../basics/upayload.adoc[UPayload]
```
stevenhartley, line 69: @sophokles73 can you update this to reflect the UPayload change merged in just now? either we define what UPayload is or we pass payload and UPayloadFormat as 2 separate parameters

upclient.adoc
```diff
@@ -20,13 +20,30 @@ SPDX-License-Identifier: Apache-2.0
 
 == Overview
 
-uPClient library is the application facing library that uEs use to communicate to a given implementation of uProtocol (ex. up-android, up-zenoh, etc...). The library contains a transport specific implementation of the link:up-l1/README.adoc[uTransport], and link:up-l2/rpcclient.adoc[RpcClient] interfaces declared in link:languages.adoc[uProtocol Language Specific Library Specifications]. The library also includes implementation specific client-side implementations of the core uServices uDiscovery, uTwin, and uSubscription.
+A uProtocol Client library is the application facing library that uEntities use to interact with other uEntities. The library contains a programminb language and transport protocol specific implementation of the low level link:up-l1/README.adoc[Transport Layer API] and the link:up-l2/rpcclient.adoc[Communication Layer API]. The library also includes implementation specific client-side implementations of the core uServices uDiscovery, uTwin, and uSubscription.
```
stevenhartley, line 23:
```suggestion
A uProtocol Client library is the application facing library that uEntities use to interact with other uEntities. The library contains a programming language and transport protocol specific implementation of the low level link:up-l1/README.adoc[Transport Layer API] and the link:up-l2/rpcclient.adoc[Communication Layer API]. The library also includes implementation specific client-side implementations of the core uServices uDiscovery, uTwin, and uSubscription.
```

upclient.adoc
```diff
@@ -20,13 +20,30 @@ SPDX-License-Identifier: Apache-2.0
...
-* *MUST* Follow the naming convention `up-client-[IMPLEMENTATION]-[LANGUAGE]` 
+Each uProtocol Client library
+[oft-sid="req~up-client-transport-api~1"]
+--
+* *MUST* provide an implementation of the link:up-l1/README.adoc[uProtocol Transport Layer API].
+--
+
+[oft-sid="req~up-client-comm-api~1"]
+--
+* *MUST* provide an implementation of the link:up-l2/rpcclient.adoc[uProtocol Communication Layer API].
```
stevenhartley, line 35: This will be updated when I add the other items and we will refer to the README I gather (but we need to separate dispatchers from the client library, I'll do that on another pull request)

stevenhartley (commented):
Minor changes needed but consider it approved from me once the changes are done. Thanks @sophokles73 !

up-l2/rpcclient.adoc
sophokles73: This will actually need more changes because the CallOptions are gone as well, aren't they? FMPOV this should be done in another PR that deals with the l2 APIs ...

sophokles73:
@stevenhartley I have accepted your proposed changes but in order to merge the PR, GitHub still requires your explicit approval ...

stevenhartley (approved):
LGTM

Issue #112, opened 2024-04-10
Definition of response messages


sophokles73:
This is the continuation of https://github.com/eclipse-uprotocol/up-core-api/issues/77
after `up-core-api` has been merged into `up-spec` as part of #91

stevenhartley:
@sophokles73 , this PR requries an update to USubscription, Udiscovery, and UTwin protos, do we push this till after 1.5.8 or do you have a Pr for it now?

stevenhartley:
Closing as completed

PR #149, opened 2024-05-16
Updated uTwin doc to match uTwin proto file

devkelley:
The method names in the uTwin proto file do not match the interface methods described in the uTwin README:

- Updated README method names to match proto file
- Updated README method definitions to match proto file


up-l3/utwin/v2/README.adoc
```diff
@@ -98,32 +98,27 @@ The Platform itself (which uTwin is part of) is acting as a distributed cache fo
...
+A uE calls this API to retrieve the last uMessages for a given set of topics. What is returned is a list of MessageResponse with the status for message retreival and the event itself if uTwin was able to fetch it. uTwin will also return status for those messages that it was unable to fetch (i.e. due to NOT_FOUND or PERMISSION_DENIED).
 
 === Set last known state
 
-*API: `rpc SetLastEvent(SetLastEventRequest) return (Google.rpc.Status)`*
-
-This function enables an "external" component to set the last event of a given Topic.
-
-WARNING: Usage of this function *MUST* be restricted to uProtocol core components, specifically the uBus
+*API: `rpc SetLastMessage(uprotocol.v1.UMessage) returns (uprotocol.v1.UStatus)`*
```
stevenhartley, line 115: Per #112 we should return nothing, as the UStatus is returned only when there is an error.

up-l3/utwin/v2/README.adoc
```diff
@@ -98,32 +98,27 @@ The Platform itself (which uTwin is part of) is acting as a distributed cache fo
...
 
-The implementation of SetLastEvent MUST enforce permissions associated with the Topic being requested. Either checking permission for each request, or if calling this API is structurally embedded inside uBus
+A call to SetLasMessage (typically from uBus) to update the uTwin internal cache with a message for a given topic (UMessage.source). Return value status.Code is one of:
```
stevenhartley, line 117: uBus makes sense for Android but not other implementations, I think we need to generalize it as "dispatcher"

stevenhartley (changes requested):

up-l3/utwin/v2/README.adoc
devkelley: This is a copy and paste from the current uTwin proto method definition to keep the docs in sync with the proto. When is the change to only return UStatus on error going to happen in the proto file? Or am I missing something here?

up-l3/utwin/v2/README.adoc
devkelley: Updated

up-l3/utwin/v2/README.adoc
```diff
@@ -98,32 +98,27 @@ The Platform itself (which uTwin is part of) is acting as a distributed cache fo
...
 
-*API: `rpc SetLastEvent(SetLastEventRequest) return (Google.rpc.Status)`*
-
-This function enables an "external" component to set the last event of a given Topic.
-
-WARNING: Usage of this function *MUST* be restricted to uProtocol core components, specifically the uBus
+*API: `rpc SetLastMessage(uprotocol.v1.UMessage) returns (uprotocol.v1.UStatus)`*
 
-The implementation of SetLastEvent MUST enforce permissions associated with the Topic being requested. Either checking permission for each request, or if calling this API is structurally embedded inside uBus
+A call to SetLastMessage (typically from a dispatcher) to update the uTwin internal cache with a message for a given topic (UMessage.source). Return value status.Code is one of:
```
ashbeitz, line 117: In the proto file (see https://github.com/eclipse-uprotocol/up-spec/blob/main/up-core-api/uprotocol/core/utwin/v2/utwin.proto), it has "(typically from uBus)", which is different to what you have "(typically from a dispatcher)".  What you have might be correct, but the two should match.

up-l3/utwin/v2/README.adoc
```diff
@@ -98,32 +98,27 @@ The Platform itself (which uTwin is part of) is acting as a distributed cache fo
...
 
 == Interface
 The proposal defines a new Digital Twin service, named uTwin, exposing the following API:
 
 === Access to last known state
 
-*API: `rpc GetLastEvent(Topic) returns (GetLastEventResponse)`*
+*API: `rpc GetLastMessages(uprotocol.v1.UUriBatch) returns (GetLastMessagesResponse)`*
 
-`GetLastEvent()` sends back a CE that was stored in the uTwin instance. uTwin's behavior is similar to a cache, it therefore has no "obligation" to retain the data and in worst case it will simply not provide the last published event.  Applications that make use of uTwin *MUST* therefore account for the possibility that the uTwin cannot deliver the event they are requesting. This may happen in the following cases:
```
ashbeitz, line 111: Wondering if parts of this deleted block may still be applicable to GetLastMessages and should be kept.

Especially the last line of the block on permissions.

ashbeitz (commented):
A few comments.  Please take a look.

up-l3/utwin/v2/README.adoc
stevenhartley: Lets align with existing functionality then.

up-l3/utwin/v2/README.adoc
stevenhartley: Can we keep the deleted text in lieu of removing it?

stevenhartley (commented):
see additional comments

up-l3/utwin/v2/README.adoc
```diff
@@ -98,32 +98,27 @@ The Platform itself (which uTwin is part of) is acting as a distributed cache fo
 
 A Topic is a cache entry that can be refreshed individually. Refreshing cache instances happens automatically as new data propagates through the network of buses
 
-uTwin instances act as a distributed networks of caches that can communicate to retrieve values without requesting a new publication, in case a uTwin instance is lacking a Topic entry in its own cache, by retrieving the last event from the cache (uTwin instance) running on the same device as the publisher of the topic). Using the analogy of a cache, it may also be useful to implement, later, a flush() command, that would remove the last known state for specific Topics, or all Topics of a given uDevice, etc.
+uTwin instances act as a distributed networks of caches that can communicate to retrieve values without requesting a new publication, in case a uTwin instance is lacking a Topic entry in its own cache, by retrieving the last event from the cache (uTwin instance) running on the same device as the publisher of the topic. Using the analogy of a cache, it may also be useful to implement, later, a flush() command, that would remove the last known state for specific Topics, or all Topics of a given uDevice, etc.
```
sophokles73, line 101:
```suggestion
uTwin instances act as a distributed network of caches that can communicate to retrieve values without requesting a new publication, in case a uTwin instance is lacking a Topic entry in its own cache, by retrieving the last event from the cache (uTwin instance) running on the same device as the publisher of the topic. Using the analogy of a cache, it may also be useful to implement, later, a `flush()` command, that would remove the last known state for specific Topics, or all Topics of a given uDevice, etc.
```

up-l3/utwin/v2/README.adoc
```diff
@@ -98,32 +98,27 @@ The Platform itself (which uTwin is part of) is acting as a distributed cache fo
...
 
-`GetLastEvent()` sends back a CE that was stored in the uTwin instance. uTwin's behavior is similar to a cache, it therefore has no "obligation" to retain the data and in worst case it will simply not provide the last published event.  Applications that make use of uTwin *MUST* therefore account for the possibility that the uTwin cannot deliver the event they are requesting. This may happen in the following cases:
-
-* uTwin is out of memory and cannot store new incoming events (the policy that defines what events are prioritized to be stored is implementation-dependent)
-* uTwin has restarted and its internal cache that stores the last events is not persistent
-
-The implementation of `GetLastEvent()` *MUST* enforce permissions associated with the Topic being requested
+A uE calls this API to retrieve the last uMessages for a given set of topics. What is returned is a list of MessageResponse with the status for message retreival and the event itself if uTwin was able to fetch it. uTwin will also return status for those messages that it was unable to fetch (i.e. due to NOT_FOUND or PERMISSION_DENIED).
```
sophokles73, line 111:
```suggestion
A uE calls this API to retrieve the last uMessages for a given set of topics. What is returned is a list of MessageResponse with the status for message retrieval and the event itself if uTwin was able to fetch it. uTwin will also return status for those messages that it was unable to fetch (i.e. due to `NOT_FOUND` or `PERMISSION_DENIED`).
```

sophokles73 (changes requested):
just some typos

up-l3/utwin/v2/README.adoc
devkelley: @stevenhartley Should I just keep the deleted text as is with the new paragraph copied from the proto file? or remove the new paragraph?

up-l3/utwin/v2/README.adoc
devkelley: reverted back to uBus for now to match documentation to protobuf

up-l3/utwin/v2/README.adoc
devkelley: I kept original text but added the protobuf description as the first couple of lines.

stevenhartley (approved):
LGTM

sophokles73 (approved):
LGTM

Issue #129, opened 2024-05-02
uPClient Interface (uP-L2)


stevenhartley:
We have now concluded clear interface and data model definitions for the transport layer (Layer 1) with `UUri`, `UAttributes`, `UPayload`, `UMessage `and `UTransport`.  

The purpose of Layer 2 in the uProtocol stack is to handle the communication patterns for RPC, pub/sub, and notifications however all we had so far is RpcClient for client-side RPC but there is no interface for server-side RPC, pub/sub, and notifications.  uSubscription has been defined at L3 as it defines the protobuf messages that are sent between client and server (inside `UPayload `of a `UMessage`).

After creating numerous implementations of uProtocol across 5 different programming languages, we've noticed that no 2 L2 interfaces for the communication patterns are defined the same as each programmer would like to leverage language specific features like Flows for Kotlin, or certain traits in Rust. 

The purpose of this issue shall be to provide a place to discuss and propose the minimum requirements for what uP-L2 client facing APIs should be extending/clarifying what is in https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l2  and https://github.com/eclipse-uprotocol/up-spec/blob/main/upclient.adoc.  Depending on the feedback for the proposal below, the contents of up-l1 folder and upclient.adoc will be updated.

NOTE: uP-L1 interfaces implements the core uProtocol (across the communication middlewares) and uP-L2 API are the client-facing APIs that wrap the transport functionality into easy to use APIs to do the most common functionality of the protocol (subscribe, publish, notify, invoke a Method, or handle RPC requests). 

# UPClient API
Below lists the high level APIs and their purpose that will be implemented per programming language. The signature of said APIs are left up to the language library implementer to define.

| API Name |  Pattern | Description|
|--------|------------|--------|
| `subscribe()` | pub/sub| Subscribe to a given topic (regardless of the location of the publisher). API must be able to optimally receive the message (ex. passing `UListener` for handling in a callback) |
| **`publish()`** | pub/sub| Publish to a given topic a message |
| **`notify()`** | Notification| Send a notification message to a given topic |
| **`invoke_method()`** |  RPC | Client-side API for RPC to send a request and receive a reply |
| **`register_method_listener()`** | RPC | Server-side API for services to receive the request and send a reply |

PLeVasseur:
I think these APIs are reasonable.

A couple of suggestions (these may already be on your todo list :) )
[...]

PR #151, opened 2024-05-21
Messaging Pattern API Specifications

stevenhartley:
The following change provides the minimal requirements such that language libraries can declare and provide a default  implementations of the communication layer (uP-L2) APIs for pub/sub, notification, and RPC messaging patterns.

#129


up-l2/api.adoc
```diff
@@ -0,0 +1,62 @@
...
+| APIs to invoke service methods and process the outcome of the service method invocation.
+
+|===
+
+
+* link:../languages.adoc[language-specific uProtocol libraries] *MUST* provide a _default_ implementation of the <<messaging-apis>> based on the link:../up-l1/README.adoc[Transport Level API].
+
+* *MUST* follow the link:../basics/error_model.adoc[uProtocol Error Model] for error handling for the respective APIs.
+
+* link:../upclient.adoc[uPClient library] *MAY* provide additional implementations of <<messaging-apis>> which, for example, take advantage of specific features of the underlying communication protocol.
```
PLeVasseur, line 60: I think I found my answer to [this question](https://github.com/eclipse-uprotocol/up-spec/pull/150#discussion_r1608975915) here! So if desired, depending on the particulars of how a language chooses to implement uP-L2 (as an interface or as a class), then it would be possible for e.g. the Zenoh client to have its own way of doing RpcClient that doesn't rest upon the foundations of uP-L1.

Is that correct, @stevenhartley?

gregmedd:
This change might go too far in removing details. A solution somewhere between #150 and this approach is probably best.

I would propose taking an approach that focuses on defining conceptual components (e.g. "There must be an RpcClient interface...") and behaviors ("...that can invoke methods hosted by an RpcServer").

Another example from #150 was in the passage that defines the error codes returned when invoking an RPC method. Instead of explicitly defining an `onError()` callback and the values of a `UStatus` parameter to that, it could be defined as a required behavior:

> The RpcClient interface should be capable of producing the following error states when an RPC method is invoked:
>
> - Request Timeout: the request has exceeded the TTL
> - Invalid Request: the request message or its parameters were invalid
> - Duplicate Request: the request's ID is already in use by an active request
>
> In addition, any errors returned in the form of a response message containing a commstatus other than OK must be surfaced.

up-l2/api.adoc
stevenhartley: Yes that can be true so long as L1 still works for all the patterns as dispatchers (streamer) will only use L1

tamarafischer (commented):
I love the new spec. defines what is needed without forcing a specific API :)
I believe developers will understand

up-l2/README.adoc
```diff
@@ -18,17 +18,17 @@ SPDX-FileType: DOCUMENTATION
...
 |Section | Description
 
 | link:dispatchers/README.adoc[*Dispatchers*]
-| uEs that are responsible for dispatching messages between uEs and between uDevices
+| uEs that are responsible for dispatching/forwarding messages between uEs and between uDevices.
 
-| link:rpcclient.adoc[*RpcClient*]
-| Interface for the generation of client-side boundary objects for uP-L3 services
+| link:api.adoc[*APIs*]
+|  Communication Layer API that implements the pub/sub, notification, and RPC messaging patterns for uE developers (applications and/or services) can use.
```
sophokles73, line 32:
```suggestion
|  Communication Layer API that allows uEntity developers (applications and/or services) to employ the pub/sub, notification and RPC messaging patterns.
```

up-l2/api.adoc
```diff
@@ -0,0 +1,77 @@
+= Communication Layer APIs
+:toc:
+:sectnums:
+
+The key words "*MUST*", "*MUST NOT*", "*REQUIRED*", "*SHALL*", "*SHALL NOT*", "*SHOULD*", "*SHOULD NOT*", "*RECOMMENDED*", "*MAY*", and "*OPTIONAL*" in this document are to be interpreted as described in https://www.rfc-editor.org/info/bcp14[IETF BCP14 (RFC2119 & RFC8174)]
+
+----
+SPDX-FileCopyrightText: 2023 Contributors to the Eclipse Foundation
```
sophokles73, line 8:
```suggestion
SPDX-FileCopyrightText: 2024 Contributors to the Eclipse Foundation
```

up-l2/api.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+* link:../languages.adoc[language-specific uProtocol libraries (Language Libraries)] *MUST* declare the Communication Layer APIs.
+
+*  Language Libraries *MUST* package (group) the <<messaging-apis>> into a folder (package/namespace) called `communication`.
+
+* Language Libraries *MUST* provide a _default_ implementation of the <<messaging-apis>> based on the link:../up-l1/README.adoc[Transport Level APIs].
+
+* *MUST* follow the link:../basics/error_model.adoc[uProtocol Error Model] for error handling for the respective APIs.
+
+* The API *SHOULD* be idempotent, meaning that calling the same API multiple times registering the same handler multiple times will have no effect.
```
sophokles73, line 71: does this _specifically_ and _only_ apply to the register handler functionality, or is this a general requirement?

If it is the former, then we should move this to the corresponding entry in the table above. If it is the latter, we should be more specific regarding the expected behavior.

Implementing idempotency for the publish functionality would be a lot to ask for at this level FMPOV ...

stevenhartley: lol yes. I removed the requirement for now, we can be more specific later as needed

PR #156, opened 2024-05-23
Align uri.proto with spec

sophokles73:
The UUri specification contains all relevant definitions and
requirements for a URI's components. The proto3 file simply defines
the encoding to use when sending UUris over the wire.

All comments containing field definitions/requirements have been
removed in order to prevent diverging semantics.


stevenhartley (approved):
LGTM!

PR #158, opened 2024-05-23
Add RPC response message usage definition

sophokles73:
Addresses #112

### DIFF
--- a/basics/error_model.adoc
+++ b/basics/error_model.adoc
@@ -18,5 +18,19 @@ SPDX-FileType: DOCUMENTATION
 SPDX-License-Identifier: Apache-2.0
 ----
 
+uProtocol supports invoking operations on (remote) service providers by means of _Remote Procedure Calls_ (RPC).
+In order to do so, a client uEntity (service consumer) sends an xref:uattributes.adoc#request-attributes[RPC Request message] to the service provider and waits for a corresponding xref:uattributes.adoc#response-attributes[RPC Response message] which conveys the outcome of the invocation of the operation.
 
-uProtocol will follow Google's https://cloud.google.com/apis/design/errors[protocol-agnostic API error model] declared in up-core-api project in link:../up-core-api/uprotocol/v1/ustatus.proto[`ustatus.proto`] please refer to the proto for details about UStatus and UCode.
+In general, an RPC can succeed or fail from the service consumer's point of view. In the former case, the response message contains the data/information that is the result of successfully processing the input data conveyed in the request message. In the latter case, the response message contains details regarding the reason why processing of the request message has failed.
+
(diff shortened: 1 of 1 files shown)


basics/error_model.adoc
```diff
@@ -18,5 +18,19 @@ SPDX-FileType: DOCUMENTATION
...
+
+uProtocol follows Google's https://cloud.google.com/apis/design/errors[protocol-agnostic API error model]. In particular, a service consumer *MUST* consider an RPC as _failed_ if the `commstatus` property contained in the xref:uattributes.adoc#response-attributes[response message attributes] has a value other than `OK`. In that case the response message's payload *MUST* be a serialized xref:../up-core-api/uprotocol/v1/ustatus.proto[UStatus] containing detail information about the reason for failure.
+
+== Service Interface Design
+
+uProtocol uses https://protobuf.dev/programming-guides/proto3/#services[Protobuf] for defining a service provider's operations. Based on the error model defined above, the _Protobuf Message_ defined/used as the response of an operation *SHOULD* only contain the data/information that represents the outcome of _successful_ execution of the operation.
```
stevenhartley, line 30:
```suggestion
uProtocol uses https://protobuf.dev/programming-guides/proto3/#services[Protobuf] for defining a service provider's operations. Based on the error model defined above, the _Protobuf Message_ defined/used as the response of an operation *SHOULD* only contain the data/information that represents the outcome of _successful_ execution of the operation.
If a response message is not required to be returned by the method that was invoked, the rpc **SHOULD** declare that it returns  `google.protobuf.Empty`  that would mean there is no `UMessage.payload`.  
Example:  `rpc DoSomething(DoSomethingRequest) returns (google.protobuf.Empty) {}`

```

stevenhartley (commented):
Minor enhancement only

basics/error_model.adoc
sophokles73: I have just read through Google's [Protobuf API best practices](https://protobuf.dev/programming-guides/api/#unique-protos) and wonder, if we should instead suggest to define _custom_ empty messages?

stevenhartley: I think it is better that we define an empty per service like below and not a generic `Empty` message but I think that is what you mean:
```
rpc DoSomething(DoSomethingRequest) returns DoSomethingResponse {}

message DoSomethingRequest {
  // A bunch of crap
}

message DoSomethingResponse {}
```

I would need to check what would then be in the `UMessage.payload`, a serialized empty message I guess (doesn't hurt)

sophokles73: Yes, that is what I meant. IMHO this will also result in just an empty byte array, because the response message is defined exactly the same like `google.protobuf.Empty`.
I will update the spec accordingly ...

stevenhartley (approved):
LGTM

sophokles73:
@stevenhartley do you also want me to change/update the core service protos accordingly, or do you want to do that in a separate PR? Technically, they should still be compliant with this updated spec, but IMHO we should change them nevertheless ...

stevenhartley:
> @stevenhartley do you also want me to change/update the core service protos accordingly, or do you want to do that in a separate PR? Technically, they should still be compliant with this updated spec, but IMHO we should change them nevertheless ...

Separate PR is fine.

PR #141, opened 2024-05-14
uProtocol Options Cleanup

stevenhartley:
The following change addresses a number of concerns with the uprotocol_options.proto that is used to declare protocol specific metadata.

#140

PR #161, opened 2024-05-23
Fixing import paths so generated headers work

gregmedd:
The protobuf compiler uses the concept of a "canonical" path when generating source files. This is the base point where it searches for import files. It _expects_ that path will match the base path for searching for headers, too. The end result is that

    import "v1/ustatus.proto"

will generate a C++ header that contains

    #include "v1/ustatus.pb.h"

while we have been explicitly asked to use includes starting with `uprotocol` in up-cpp:

    #include "uprotocol/v1/ustatus.pb.h"

We were able to work around this with some tricks when the import paths were file names only (e.g. `import ustatus.proto`). However, the proto files have moved under the `v1/` directory as of #141.

Based on my reading of the protoc documentation and several stack overflow postings, it seems the intended way for .proto files to be written in this scenario is to import the full canonical path as intended for the output files:

    import "uprotocol/v1/ustatus.proto"

so that protoc can generate the correct paths in output sources in the first place.


gregmedd:
I've combined #162 into this PR.

.github/workflows/build-github-package.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         run: mvn -B package --file up-core-api/pom.xml
```
sophokles73, line 40: I suppose this also needs to be updated? In fact, using the pom.xml file from the current working directory should not require the `--file` param at all.

.github/workflows/build-github-package.yml
```diff
@@ -40,7 +40,7 @@ jobs:
...
   
       - name: Publish maven package to GitHub Packages
-        run: mvn deploy -s $GITHUB_WORKSPACE/settings.xml --file up-core-api/pom.xml
+        run: mvn deploy -s $GITHUB_WORKSPACE/settings.xml --file pom.xml
```
stevenhartley, line 43:
```suggestion
        run: mvn deploy -s $GITHUB_WORKSPACE/settings.xml
```

.github/workflows/build_and_test.yml
```diff
@@ -21,4 +21,4 @@ jobs:
         distribution: 'temurin'
         cache: maven
     - name: Build and Test with Maven
-      run: mvn -B package --file up-core-api/pom.xml
+      run: mvn -B package --file pom.xml
```
stevenhartley, line 24:
```suggestion
      run: mvn -B package
```

.github/workflows/build_deploy.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         gpg-passphrase: GPG_PASSPHRASE
 
     - name: Build and Publish to Apache Maven Central
-      run: mvn clean deploy --file up-core-api/pom.xml
+      run: mvn clean deploy --file pom.xml
```
stevenhartley, line 31:
```suggestion
      run: mvn clean deploy
```

.github/workflows/release.yml
```diff
@@ -34,7 +34,7 @@ jobs:
 
     - name: Stage to Nexus and Release to Maven central
       run: |
-        mvn -B release:clean release:prepare -P release release:perform --file up-core-api/pom.xml
+        mvn -B release:clean release:prepare -P release release:perform --file pom.xml
```
stevenhartley, line 37:
```suggestion
        mvn -B release:clean release:prepare -P release release:perform
```

.github/workflows/release.yml
```diff
@@ -44,7 +44,7 @@ jobs:
 
     - if: cancelled() || failure()
       run: |
-        mvn -B release:rollback --file up-core-api/pom.xml
+        mvn -B release:rollback --file pom.xml
```
stevenhartley, line 47:
```suggestion
        mvn -B release:rollback
```

pom.xml
stevenhartley, line 1: @gregmedd there is a bug in the pom.xml that you copied from my 162 PR. 

Wrong:
`<protoSourceRoot>${project.basedir}/uprotocol</protoSourceRoot>`

Correct:
`<protoSourceRoot>${project.basedir}/up-core-api</protoSourceRoot>`

stevenhartley (changes requested):
Needs a fix

PR #162, opened 2024-05-23
Fix the proto Include paths

stevenhartley:
The generated code needs to be in the `uprotocol` namespace folder structure for clarity however the protos root was the uprotocol folder so generated code was missing the uprotocol folder. This change fixes the protos and cleans up the build to still work and validate the build/changes.

#161


gregmedd:
This has been combined with the existing PR #161

PR #163, opened 2024-05-27
Use dedicated messages in service operations

sophokles73:
Changed operation definitions of uProtocol core services to always
use dedicated request and response messages as recommended by
Protobuf Best Practices.

Removed UStatus from response messages, following recommendations
from uProtocol Error Model specification.

This is a follow up for #112

### DIFF
--- a/up-core-api/uprotocol/core/usubscription/v3/usubscription.proto
+++ b/up-core-api/uprotocol/core/usubscription/v3/usubscription.proto
@@ -204,6 +204,7 @@ message UnsubscribeRequest {
   SubscriberInfo subscriber = 2;
 }
 
+// The (empty) response message representing the successful execution of the Unsubscribe operation.
 message UnsubscribeResponse {}
 
 // Passed to FetchSubscribers such that we can obtain a list of subscribers to
@@ -266,7 +267,6 @@ message FetchSubscriptionsRequest {
 
 // Results from FetchSubscriptions() API
 message FetchSubscriptionsResponse {
-  reserved 3;
   // Repeated list of subscriptions for a subscriber
   repeated Subscription subscriptions = 1;
 
@@ -286,6 +286,8 @@ message NotificationsRequest {
   SubscriberInfo subscriber = 2;
 }
 
+// The (empty) response message representing the successful execution of the
+// RegisterForSubscriptionChanges and UnregisterForSubscriptionChanges operations.
 message NotificationsResponse {}
 
 // Subscription Update Message.
@@ -351,4 +353,5 @@ message ResetRequest {
   }
 }
 
+// The (empty) response message representing the successful execution of the Reset operation.
 message ResetResponse {}
--- a/up-core-api/uprotocol/core/usubscription/v3/usubscription.proto
+++ b/up-core-api/uprotocol/core/usubscription/v3/usubscription.proto
@@ -54,7 +54,7 @@ service uSubscription {
 
   // The consumer no longer wishes to subscribe to a topic so it issues an
   // explicit unsubscribe request.
-  rpc Unsubscribe(UnsubscribeRequest) returns (uprotocol.v1.UStatus) {
+  rpc Unsubscribe(UnsubscribeRequest) returns (UnsubscribeResponse) {
     option (uprotocol.method_id) = 2;
   }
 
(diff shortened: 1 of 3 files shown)


up-core-api/uprotocol/core/usubscription/v3/usubscription.proto
```diff
@@ -284,13 +280,13 @@ message FetchSubscriptionsResponse {
...
   // Topic to register/unregister to receive subscription change notifications
-  v1.UUri topic = 1;
+  uprotocol.v1.UUri topic = 1;
 
   // Subscribers's information
   SubscriberInfo subscriber = 2;
 }
 
-
+message NotificationsResponse {}
```
stevenhartley, line 289: Missing documentation for the response message

up-core-api/uprotocol/core/usubscription/v3/usubscription.proto
```diff
@@ -183,34 +182,35 @@ message SubscriptionRequest {
...
 message UnsubscribeRequest {
   // Topic to be unsubscribed to
-  v1.UUri topic = 1;
+  uprotocol.v1.UUri topic = 1;
 
   // Subscriber identification
   SubscriberInfo subscriber = 2;
 }
 
+message UnsubscribeResponse {}
```
stevenhartley, line 207: Please document

up-core-api/uprotocol/core/usubscription/v3/usubscription.proto
```diff
@@ -354,3 +350,5 @@ message ResetRequest {
     }
   }
 }
+
+message ResetResponse {}
```
stevenhartley, line 354: Missing documentation

up-core-api/uprotocol/core/utwin/v2/utwin.proto
```diff
@@ -66,9 +64,19 @@ message MessageResponse {
   uprotocol.v1.UMessage message = 3;
 }
 
+message GetLastMessagesRequest {
```
stevenhartley, line 67: Please add documentation

up-core-api/uprotocol/core/utwin/v2/utwin.proto
```diff
@@ -66,9 +64,19 @@ message MessageResponse {
...
+  uprotocol.v1.UUriBatch topics = 1;
+}
 
 // Message returned by the rpc GetLastMessages.
 message GetLastMessagesResponse {
   // List of one or more messages and the results for fetching the message per topic
   repeated MessageResponse responses = 2;
 }
+
+message SetLastMessageRequest {
```
stevenhartley, line 78: Add documentation

stevenhartley (changes requested):
Looks good but please document the newly added fields and messages. Also mention why a field is reserved

sophokles73:
> Looks good but please document the newly added fields and messages. Also mention why a field is reserved

I was actually wondering, if we really want/need to reserve the tags for the status fields. FMPOV we would need to do that if we had code that would process messages containing the status field but is then presented with a message that contains a different field under that tag.

Question is: do we have code in the field that expects to find the current message structure as defined on the main branch?

up-core-api/uprotocol/core/usubscription/v3/usubscription.proto
sophokles73: Hmm, what is there to document except for the obvious
> The (empty) response message representing the successful execution of the `Unsubscribe` operation.

This is exactly what we have defined in the Error Model specification. Do you think this is worth adding here again? Or am I missing something?

stevenhartley:
> > Looks good but please document the newly added fields and messages. Also mention why a field is reserved
> 
> I was actually wondering, if we really want/need to reserve the tags for the status fields. FMPOV we would need to do that if we had code that would process messages containing the status field but is then presented with a message that contains a different field under that tag.
> 
> Question is: do we have code in the field that expects to find the current message structure as defined on the main branch?

We have v2 code in the field but not v3 (other than in up-android-core but we can fix that) and the v2 code expects the status. Having said that we don't need to keep the placeholder imho.

up-core-api/uprotocol/core/usubscription/v3/usubscription.proto
stevenhartley: yes the obvious :-)

sophokles73:
@stevenhartley I have pushed changes ...
Please do not forget to squash when merging.

PR #165, opened 2024-05-27
SOME/IP specification cleanup

### LINKED ISSUE #93: Update SOME/IP specification to uProtocol 1.5.x

int0x27:
The following change is a reboot of the SOME/IP specifications in an attempt to drastically simplify to show how we map up-l1 messages to SOME/IP messages and uSubscription messages to SOME/IP-SD messages. the older specifications had a lot of uProtocol version 1.3.6 concepts (cloudevents) that were rather complicating the situation.

(based on PR-117, that can't be merged at the moment)

Fixes #93


up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+
+The following table highlights the mapping of *uAttributes* to *SOME/IP* fields that apply to all event types. For more information about *uAttributes*, please refer to link:../basics/uattributes.adoc[uAttributes Specification].
+
+
+==== Uris
+
+link:../basics/uattributes.adoc[uAttributes Specification] explains that `source` attributes defines the address of whom sent the message, while `sink` defines the destination for the message in link:../basics/uri.adoc[`UUri`] format. `UUri` is translated to *SOME/IP* attributes per <<sink-source-uri-mapping>> below.
+
+.Sink & Source URI Mapping
+[sink-source-uri-mapping, cols="1,1,2"]
```
stevenhartley, line 64:
```suggestion
[#sink-source-uri-mapping, cols="1,1,2"]
```

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+|===
+| *UUri Portion* | *SOME/IP Field* | *Description*
+a| `authority_name` | IP address & port
+a| IP address & port (destination endpoint) of the mDevice.
```
stevenhartley, line 68: I think we need to elaborate more on this portion. The URI can be either a host or IP address. If the authority_name is host, we need to then resolve to an IP address in the transport implementation.

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+
+NOTE: `vsomeip` hides IP endpoint information and uses `Service ID`/`Instance ID` instead.
```
stevenhartley, line 70: I think you need to elaborate more what is meant here. A service (and instance) is deployed to a given ECU device that has an IP address, I don't think it is hidden per say.

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+* Generated *SOME/IP* Responses *MUST* auto-populate the `Request ID` cached from the request message, into the response message and then flush the entry in the cache.
+
+Further details of the usage of IDs for the various message types can be found in the next section.
+
+==== Message Type
+
+<<umessagetype-mapping>> table below maps of uProtocol messages to `[PRS_SOMEIP_00055]` *SOME/IP* message types.
+
+.UMessageType Mapping
+[umessagetype-mapping, cols="1,1,2"]
```
stevenhartley, line 116:
```suggestion
[#umessagetype-mapping, cols="1,1,2"]
```

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+| UMessageType | SOME/IP Type | Details
+
+a| `UMESSAGE_TYPE_PUBLISH` a| `NOTIFICATION` | Publish
+
+a| `UMESSAGE_TYPE_REQUEST` a| `REQUEST` | Requests
+
+a| `UMESSAGE_TYPE_RESPONSE` a| `RESPONSE` or `ERROR` | Responses or error has occurred while attempting to deliver the message or a service
+has thrown an exception
+
+a| `UMESSAGE_TYPE_NOTIFICATION` a| `NOTIFICATION` | Notification
```
stevenhartley, line 127: Per our chat, this is a different type of message that is a request with no response correct?

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+
+  * *MUST* cache the *SOME/IP* `Request ID` as well as the generated `UAttributes` for the request messages so that the response can be translated back to a *SOME/IP* response message
+
+
+==== Communication Status
+
+<<commstatus-mapping>> below provides the mapping of link:../basics/uattributes.adoc[UAttributes] `commstatus` `UCode` codes to `[PRS_SOMEIP_0019]` *SOME/IP* error codes.
+
+.Error Code Mapping
+[commstatus-mapping, width=70%]
```
stevenhartley, line 156:
```suggestion
[#commstatus-mapping, width=70%]
```

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+
+```
+response.priority = request.priority
+response.reqid = request.id
+```
+
+When sending auto-generated *SOME/IP* REQUEST messages:
+
+  * *MUST* cache the message's `Request ID` to correlate with the RESPONSE message.
+  * Underlying *SOME/IP* library *SHOULD* handle `Request ID` updating.
```
stevenhartley, line 144: @int0x27 I don't understand this requirement, can you please explain what you mean here?

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+a| `NOT_FOUND` a| `E_UNKNOWN_SERVICE`
+a| `UNAVAILABLE` a| `E_NOT_READY`
+a| `DATA_LOSS` a| `E_MALFORMED_MESSAGE`
+a| `INTERNAL` a| `E_NOT_REACHABLE`
+a| `UNKNOWN` a| `E_NOT_OK`
+a| `FAILED_PRECONDITION` a| `E_WRONG_PROTOCOL_VERSION` or `E_WRONG_INTERFACE_VERSION`
+|===
+
+
+=== UPayload
```
stevenhartley, line 172: @int0x27 UMessage no longer contains UPayload and there is no more reference vs vale. Now the UMessage contains the data in payload and the UPayloadFormat is stored in UAttributes (change made last week).

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+a| `Major Version` a| `ue_version_major`
+
+a| `EventGroup ID` / `Event ID` a| `resource_id`
+
+NOTE: `1:1` mapping between `EventGroup ID` and `Event ID` is assumed. This may require specific ECU Firmware.
+
+
+|===
+
+NOTE: `UUri.authority_name` *MAY* be translated to/from IPv4 (and/or IPv6) Endpoint Option of the *SOME/IP-SD* message, although in `vsomeip` this is not available in the API (e.g. each discovered Endpoint maps to `Service ID`/`Instance ID`/`Major Version`/`Minor Version`).
```
stevenhartley, line 222: This is where we need to chat, I'm not sure how that would work. How do we know that service_id+instance_id maps to a specific IP address and port, is that static configuration that would be loaded into the uTransport implementation library?

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+a| The message is used to acknowledge a successful subscription request.
+
+* `SubscriptionStatus.code` *SHALL* be set to `OK`
+* `SubscriptionStatus.state` *SHALL* be set to `SUBSCRIBED`
+
+a| `SubscribeEventGroupNack`
+a| `SubscriptionResponse`
+a| The message is used to acknowledge a failed subscription request.
+
+* SubscriptionStatus.code *SHALL* be set to the corresponding error code per the <<commstatus-mapping>> table
```
stevenhartley, line 250: This is changing in https://github.com/eclipse-uprotocol/up-spec/pull/163, it might be better to just state that ,the error codes below will be returned and not through SubscriptionStatus.code (it will be in commstatus of UAttributes)

stevenhartley (changes requested):

up-l1/someip.adoc
PLeVasseur: Also very curious about this point.

I don't know vsomeip all that well, but from what I could see it seems to rely on static configuration of IP addresses and I couldn't figure out reading the vsomeip header how this tied back to an IP address.

If the answer is as you wrote and there is no IP address or logical name contained within the vsomeip header, how do we come up with this?

Must we then read that same vsomeip configuration to be able to correlate service and instance IDs with the IP address?

up-l1/someip.adoc
PLeVasseur: Curious to learn more here too.

PLeVasseur (changes requested):
Hi @int0x27 -- I don't know much about vsomeip yet.

I'm still a bit unclear on how the source and sink UUris map in.

Could I ask for some elaboration on how you see the source and sink UUris lining up with the SOME/IP header?

I'm looking for something akin to what was done for [Zenoh](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/zenoh.adoc#25-uri-mapping) or [MQTT5](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/mqtt_5.adoc#22-uri-mapping-to-mqtt-topics).

I'd essentially like the kind of clarity laid out in those documents on how to map from the uP-L1 registerListener API down into the vsomeip header.

This kind of information will be essential when a uStreamer is fleshed out, as we'll need to be able to say things like:

"Give me all point-to-point (Request, Response, Notification) messages with a source authority of X, so that I can route them to sink authority Y"

and

"Give me Publish messages from topics A, B, and C, since I will need to bring those onto this host for dissemination to uEntities Foo: {A, B} and Bar: {C}"

up-l1/someip.adoc
int0x27: `vsomeip` gets the required endpoint information (via service discovery or static json configuration), but they have decided to leave it private for the implementation, so you have a better abstraction of SOME/IP services/instances/methods available in the network.

e.g. it is possible to have multiple endpoints (hosts) offering the same service, but with different instance or version. vsomeip router aggregates all discovered service/instance information and maps the endpoint to service/instance/major_ver/minor_ver/[method | eventgrup/event]
API requires service and instance as minimum.

The only way I see to provide the mDevice IP:port in `authority_name` is to have it pre-configured (mapped to service/instance pair).

If "host" means any string, it could be created from service/instance pair, e.g. "<service>.<instance>v<version>", but then that same information is present also in ue_id, ue_version_major

up-l1/someip.adoc
int0x27: Although it's not hidden by the protocol, it is hidden by the implementation. You can't work directly with the endpoints, so they have abstracted them by `service/instance/..`

up-l1/someip.adoc
int0x27: `UMESSAGE_TYPE_NOTIFICATION` was previously mapped to `REQUEST_NO_RESPONSE`, which seems wrong.

I can't tell whether to use `UMESSAGE_TYPE_PUBLISH` or `UMESSAGE_TYPE_NOTIFICATION` for incoming SOME/IP events. e.g. if we have a subscriber to the specific `service/instance/event`, it should be a  `UMESSAGE_TYPE_NOTIFICATION`,  if no subscribers are found send it as `UMESSAGE_TYPE_PUBLISH`?

up-l1/someip.adoc
int0x27: I mean let the library do the `Request ID` manipulation.

The transport just needs to `send()` the request and cache the updated `Request ID` (as there are multiple cases, whether to increment Session ID or not).
Sending notifications also has automatic `Request ID` assignment.

// removed some lines about setting Client ID to 0 on events / incrementing the Session ID, etc.

up-l1/someip.adoc
int0x27: I noticed it was referring to a missing entry and tried to figure it out. Maybe it will be better if you rewrite this section.
e.g. why using unspecified payload format (actually that was removed as well), instead of being `UPAYLOAD_FORMAT_SOMEIP`. 
Is there any difference between `UPAYLOAD_FORMAT_RAW` and `UPAYLOAD_FORMAT_SOMEIP`?

up-l1/someip.adoc
int0x27: Some other SOME/IP implementations, require you to specify all the local and remote endpoints in their static configuration along with their service/instance/methods, etc.

For `vsomeip` it can be also static (if service discovery is disabled) or auto-populated from service discovery.
Consider you want to use {service:0x1234, instance:2, method:0x0042, major_ver:1}

- The app registers `on_available(service, instance, is_available)` callback, possibly with ANY_SERVICE, ANY_INSTANCE artguments.
- App calls `app->request_service(0x1234, ANY_INSTANCE, 1, ANY_MINOR)`
- This generates Find Service Entry SOME/IP message `[PRS_SOMEIPSD_00350]`
and all mDevice endpoints must return their 0x1234 service endponits (e.g. hostA provides instance 1, hostB provides instance 2)
- App receives `on_available(0x1234, 1, true)` and `on_available(0x1234, 2, true)` and then it can communicate with the choosen instance
- e.g. send a message to configured instance 2:
```c++
std::shared_ptr<vsomeip::message> rq = vsomeip::runtime::get()->create_request(use_tcp_);
rq->set_service(0x1234);
rq->set_instance(2);
rq->set_method(0x0042);
rq->set_interface_version(1); // major_ver
std::shared_ptr<vsomeip::payload> pl = vsomeip::runtime::get()->create_payload();
// serialize payload for method_id
rq->set_payload(pl);
app_->send(rq);
```
- vsomeip finds the discovered service/instance(2) and sends the request to `HostB` endpoint (app does not care about which host is providing it)
- You can also set `rq->set_instance(1)` and it will be sent to the other `hostA`

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+[sink-source-uri-mapping, cols="1,1,2"]
+|===
+| *UUri Portion* | *SOME/IP Field* | *Description*
+a| `authority_name` | IP address & port
+a| IP address & port (destination endpoint) of the mDevice.
+
+NOTE: `vsomeip` hides IP endpoint information and uses `Service ID`/`Instance ID` instead.
+
+a| `ue_id` a| `Instance ID` \| `Service ID`
+a| `Service ID` *MUST* be encoded in the lower 16 bits of `resource_id`.
```
pranavishere2, line 73: As per my understanding,
`Service ID` *MUST* be encoded in the lower 16 bits of `ue_id` and not 'resource_id'. 
Please confirm

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+If the upper 16 bits are `0x0000` (a reserved instance value), `Instance ID` is assumed to be `1`.
+
+NOTE: Explicitly setting `Instance ID` to `0x0001` *SHOULD* be avoided as it increases the message size.
+
+a| `ue_version_major` a| `Interface Version`
+a| `Interface Version` field of *SOME/IP Header*. `[PRS_SOMEIP_00053]`.
+
+Maps to `Major Version` for the `Service ID`. If set, it *SHOULD* be passed to underlying *SOME/IP* library to override it's default _Service Major Version_
+
+a| `resource_id` a| `Method ID`/`Event ID` | Identifier of the method/event placed in the
```
pranavishere2, line 85: As per my understanding,
uProtocol resource_id is mapped to method_id in someip header for RPC usecase and
uProtocol resource_id is mapped to event ID in someip header for Pub Sub usecase
Question: How are we going to map event group id in vsomeip ? If we se event_id same as event_groupID then we may have a problem. We faced this issue wirking with chimes team few weeks ago where they have usecase of different event_id and event_group ID for their publish-subscribe usecase.

up-l1/someip.adoc
```diff
@@ -0,0 +1,265 @@
...
+
+a| `ue_id` a| `Instance ID` \| `Service ID`
+a| `Service ID` *MUST* be encoded in the lower 16 bits of `resource_id`.
+
+`Instance ID` *MAY* be encoded in the upper 16 bits of the 32 bit `ue_id`.
+If the upper 16 bits are `0x0000` (a reserved instance value), `Instance ID` is assumed to be `1`.
+
+NOTE: Explicitly setting `Instance ID` to `0x0001` *SHOULD* be avoided as it increases the message size.
+
+a| `ue_version_major` a| `Interface Version`
```
pranavishere2, line 80: As per my understanding,
uProtocol "ue_version_major" is mapped to Service Major Version in someip. What about minor version?
In someip world, minor version is also required at the time of offer service at server side and request service at client side. We will have a problem when mE would try to offer a service with different minor version and Streamer will need to request the same service with that specific minor version. 
I think, we need a mapping of ue_minor_version to map it to vsomeip Service Minor version as well.
Please confirm

up-l1/someip.adoc
int0x27: My opinion is still in favor of using `Event ID` for both `PUBLISH` and `SubscriptionRequest.topic.resource_id`.
This needs additional configuration for the transport (mapping Service/Instance/Event to a set of eventgroup IDs), but it allows a lot more flexibility in supporting ECUs.
The other option: forcing ECUs to have `EventID=EventGroupID` is a lot more restrictive for the HW and not OSS friendly. This can be assumed as "default" option - if no mapping is provided per event, assume `EventID=EventGroupID`

up-l1/someip.adoc
int0x27: `minor version` is needed indeed , if omitted vsomeip assumes `vsomeip::DEFAULT_VERSION=0x00000000`.
The same applies for `major version` (assumed `0x00`, if not provided)
As minor is not available in uProtocol, I suppose it should be a transport configuration (especially for offered services).
For requested services we can get away with `vsomeip::ANY_MINOR`

The transport *MUST* know in advance which services/instances/major/minor and events it *MAY* interact with and pre-register them in the vsomeip router. It could be possible to request the event in e.g. `on_available` handler, but that may block the dispatcher thread for too long and you'll get vsomeip warnings.

up-l1/someip.adoc
```diff
@@ -24,164 +26,207 @@ SPDX-License-Identifier: Apache-2.0
...
-`UMessage.attributes` also contains link:../basics/upayloadformat.adoc[UPayloadFormat] used to give a hint of the payload format (protobuf serialized, SOME/IP format, TEXT, RAW, etc...).
 
-The *SOME/IP* specification however does not have an equivalent field for `UPayloadFormat` as it is assumed that the payload is serialized in the format that the other end knows how to deserialize (i.e. it is fixed per topic). As such, when converting between *uProtocol* and *SOME/IP*, the `UPayloadFormat` field *SHOULD* be ignored (left at the default of `UMESSAGE_TYPE_UNSPECIFIED`).
+==== Communication Status
 
+<<commstatus-error-mapping>> below provides the mapping of link:../basics/uattributes.adoc[UAttributes] `commstatus` `UCode` codes to SOME/IP error codes `[PRS_SOMEIP_0019]`.
 
-=== uTransport
+.UCode to SOME/IP Error Code Mapping
+[[commstatus-error-mapping]]
```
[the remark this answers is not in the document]
int0x27: I get warnings from asciidoc while usint `[#id]`, so used alternative notation (for me it works in generated html, vscode and in github rendering)
```
$ asciidoc someip.adoc
asciidoc: ERROR: someip.adoc: line 217: illegal style name: #commstatus-error-mapping
```

up-l1/someip.adoc
```diff
@@ -24,164 +26,207 @@ SPDX-License-Identifier: Apache-2.0
...
-    * `Event ID` is an alias for `Method ID`, but used for events. `[PRS_SOMEIP_00245]`
-    * `Instance ID` is defined in *SOME/IP-SD Service Entry Type*, specified by AUTOSAR `[PRS_SOMEIPSD_00268]`.
-    * `Instance ID` is not present in *SOME/IP Header*, but is needed for _Service Discovery_ and events and *MUST* be provided for *SOME/IP* libraries (e.g. `vsomeip`). `[PRS_SOMEIP_00162]`
-    * Additional mappings *MAY* be required by the underlying *SOME/IP* library, e.g. Service Major/Minor version. Such values *MAY* be pre-configured
+* A `NOTIFICATION` *MAY* use the default priority.
+* A `RESPONSE` *SHOULD* use the `REQUEST` priority.
 
+| ttl | Should be checked for incoming `UMESSAGE_TYPE_REQUEST` messages.
+
+* SOME/IP `REQUEST` *SHOULD* not be sent if incoming UMessage has `ttl > 0` and timestamp from `id` + `ttl` has already expired un *UE* host.
```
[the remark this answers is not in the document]
int0x27: OK, I can simplify it, what about "UMessage has expired." sounds better for ttl

up-l1/someip.adoc
```diff
@@ -24,164 +26,207 @@ SPDX-License-Identifier: Apache-2.0
...
+* If present `ttl` *SHOULD* be used to limit the time a `Request` is kept in the cache for correlation with a `RESPONSE`.
+* If needed `ttl=0` *MAY* be used to indicate `REQUEST_NO_RETURN` (currently not allowed)
+
+NOTE: Unix timestamps between hosts are very likely to be different (even with `ntp`, especially for cloud events), so the transport *MAY* assume the local timestamp of receiving the request, especially for `ttl < 100ms`.
```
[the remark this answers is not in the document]
int0x27: About the NOTE?

I mean relying on UUID timestamp to be in sync on different hosts even in the vehicle network is a bit optimistic. Especially if requests have very small `ttl`, it may randomly expire them.

My idea addresses this issue: For local (in vehicle) uMessage requests with small enough `ttl` we may assume the time between generating the request UUID and receiving it in uE is almost identical. So in that edge case `ttl` should be used as response timeout. Consider the measured vsomeip request -> response is ~17ms.
This means in the ideal case tts < 20ms will either expire or not being send at all depending on sender and receiver clock deviation and network latency.

For cloud events it makes much more sense to check timestamp from UUID, but probably they should also use a bigger ttl as well.

So this is an optional solution the problem, if you want I'll remove it.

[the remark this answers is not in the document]
stevenhartley: Still do not understand. Are you saying we should not check ttl for in-vehicle use cases? We can't control app developers doing the wrong thing (putting a timeout of 1ms for something that should take 1000ms, etc...)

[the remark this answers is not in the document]
int0x27: It really depends if UMessage Request may be delayed in in-vehicle use cases.
If we assume just network latency for delivering the request, it simplifies timeout handling in some/ip bridge.
Then `ttl` can be used as RESPONSE timeout.

Otherwise `ttl` is used for both REQUEST validation and the remaining time as RESPONSE timeout.
Including the UUID timestamp variability between hosts it looks that small `ttl` values will cause undefined behavior, e.g:
  - sending the REQUEST, timing out for RESPONSE
  - sending the REQUEST and RESPONSE in time
  - not sending REQUEST at all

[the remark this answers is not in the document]
stevenhartley: The purpose of the ttl in the request really was for how long should the request be active for before we give up and declare a failure (and generate a `DEADLINE_EXCEEDED` error). So I think we are saying the same thing actually.

stevenhartley:
@int0x27 I think we are OK to merge, next week we can add specific mappings of the uTransport API given the work that Peter has done on up-transport-vsomeip-rust.

int0x27:
OK, just rebased on main, squashed some commits

stevenhartley (approved):
LGTM for now, we'll add to it more for uTransport specifics next week

PR #166, opened 2024-05-28
Remove obsolete proto imports

sophokles73:


stevenhartley (approved):
Thanks Kai for the quick fix!

PR #167, opened 2024-05-29
Restrict authority to lowercase unreserved characters only

sophokles73:


stevenhartley (approved):
LGTM

sophokles73:
@evshary does this work for you as well?

basics/uri.adoc
```diff
@@ -133,16 +133,19 @@ URI = "up" ":" hier-part
...
-[.specitem,oft-sid="dsn~uri-host-only~1",oft-needs="impl,utest"]
---
-The _authority_ component of a URI *MUST* consist of a _host_ part only, i.e. it *MUST NOT* contain neither _userinfo_ nor _port_ information.
+[.specitem,oft-sid="dsn~uri-host-only~2",oft-needs="impl,utest"]
 --
+The _authority_ component of a URI *MUST* consist of characters according to the following rule:
 [source,abnf]
 ----
-authority = host
+lc-unreserved = *( %x61-7A / DIGIT / "-" / "." / "_" / "~" )
```
evshary, line 141: So we don't allow URL encoding now, right?
For example: can we use `//%78%79%7A/...`  instead of `//xyz/...`?

evshary (approved):
I like this change, which makes our life easier.
Only one question to be confirmed, otherwise it looks perfect to me.

basics/uri.adoc
sophokles73: No, there is no more percent encoding. The reason why we wanted to remove support for it is, that in order to properly match filter URIs (for invoking listeners), we would always need to _normalize_ all URIs, which might be quite expensive. Is that an issue?

evshary: No. Actually, it's better for me. :laughing:

PR #182, opened 2024-06-12
removed zero-length segments from mqtt 5 spec

devkelley:
Certain messaging brokers handle zero-length segments differently for topics, as such we will simply remove the zero length segments if the sink uri is 'None'

### DIFF
--- a/up-l1/mqtt_5.adoc
+++ b/up-l1/mqtt_5.adoc
@@ -129,9 +129,9 @@ The MQTT topic a message is published on utilizes the source and sink UUri field
 
 `{client_identifier}/{source.authority_name}/{source.ue_id}/{source.ue_version_major}/{source.resource_id}/{sink.authority_name}/{sink.ue_id}/{sink.ue_version_major}/{sink.resource_id}`
 
-If the no sink uuri is provided, then the topic sections using sink uuri segments are empty, like so:
+If the no sink uuri is provided, then the topic sections using sink uuri segments are removed, like so:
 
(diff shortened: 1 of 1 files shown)


stevenhartley (approved):
LGTM, thanks!

up-l1/mqtt_5.adoc
```diff
@@ -179,8 +179,8 @@ Shows how source and sink filters can be used to construct wildcard topics for l
 |===
 | goal | source filter | sink filter | listener topic
 
-| Subscribe to all publish messages from devices | empty | None | d/\+/+/\+/+////
-| Subscribe to all messages sent from a UAuthority | //vehicle_1.veh.com/FFFF/FF/FFFF | empty | d/vehicle_1.veh.com/\+/+/\+/+/\+/+/\+/+
+| Subscribe to all publish messages from devices | empty | None | d/\+/+/\+/+
+| Subscribe to all message types but publish messages sent from a UAuthority | //vehicle_1.veh.com/FFFF/FF/FFFF | empty | d/vehicle_1.veh.com/\+/+/\+/+/\+/+/\+/+
```
r-vanooyen, line 183: what is the difference between a sinkFilter of "none" and sinkFilter of "empty"?

if this "empty" sinkFilter is a sinkFilter with
```
authorityName = "*"
uEid=0xffff
majorVersion = 0xff
resourceId = 0xffff
```
could we rename it with "any" or "wildcard"?

PR #184, opened 2024-06-17
Use inclusive range definitions only

### LINKED ISSUE #181: Use inclusive ranges for integer values
In the [uattributes spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc), a mix of inclusive (`[X, Y]`) and exclusive (`(X, Y)`) range formats are used. Whenever the values for a range are integers, only inclusive ranges should be used to express the range to avoid confusion. Exclusive ranges only add meaning when discussing floating point values.

For example, one requirement reads:

> The UUri's `resource_id` **MUST** be set to a value in the range `[0x8000, 0xFFFF)`

The range, as written, can be read as "any value starting at 0x8000 up to, but not including, 0xFFFF" - which requires mental subtraction to find the real maximum value of `0xFFFE`. If only inclusive ranges are used instead, this reads more cleanly:

> The UUri's `resource_id` **MUST** be set to a value in the range `[0x8000, 0xFFFE]`

In this form, it would be read as "any value from 0x8000 to 0xFFFE, inclusive". No additional interpretation is required to understand the minimum and maximum allowable values.

sophokles73:
The ranges of valid resource identifiers are now defined by means of
inclusive lower and upper boundaries.

Fixes #181

### DIFF
--- a/basics/uattributes.adoc
+++ b/basics/uattributes.adoc
@@ -102,7 +102,7 @@ The following table defines attributes that are used for publish messages in add
 |yes
 a|A link:uri.adoc[UUri] representing the topic that this message is published to.
 
-* The UUri's `resource_id` *MUST* be set to a value in range `[0x8000, 0xFFFF)`.
+* The UUri's `resource_id` *MUST* be set to a value in range `[0x8000, 0xFFFE]`.
 |===
 
 [#notification-attributes]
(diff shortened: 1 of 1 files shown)


stevenhartley (approved):
LGTM thanks

gregmedd (approved):
Looks good 👍

Issue #185, opened 2024-06-17
unused import warrning in uattributes.proto


stevenhartley:
Once I committed #173 I caused an issue with a warning that appears that ustatus proto is included in uattributes.proto even though it is not used so this issue will be used to remove the warning.

Issue #173, opened 2024-06-04
Split UCode from UStatus


stevenhartley:
Mechatronics folks would like to use UCode in their uService definitions and not be dependent on UStatus that includes strings. To avoid the uservice protos from having to include UStatus, we would like to pull UCode out from UStatus (this should not affect generated code but we will check to be sure).

halimragab:
The main purpose of this is simplification, where embedded controllers can import UCode only and don't have to think or worry about UStatus (which includes fields that are not typically used in embedded software).

PR #186, opened 2024-06-17
Fix warning for unused ustatus.proto

stevenhartley:
Once I committed #173 I caused an issue with a warning that appears that ustatus proto is included in uattributes.proto even though it is not used so this issue will be used to remove the warning.

#185


sophokles73 (approved):
LGTM

PR #188, opened 2024-06-19
Add the table for listeners to distinguish different message types.

evshary:
As @PLeVasseur suggested [here](https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/pull/40#discussion_r1635545740), add the table mapping for listeners and message type.
This should be helpful for the implementation.

### DIFF
--- a/basics/uattributes.adoc
+++ b/basics/uattributes.adoc
@@ -226,3 +226,39 @@ a|The link:uri.adoc[UUri] that the service consumer expects to receive this resp
 == Validation
 
 Each link:../languages.adoc[uProtocol Language Library] *MUST* provide means to assert the consistency of a message's attributes according to its type as defined in the sections above. The concrete implementation should follow common practices of the particular programming language.
+
+
+== Listeners for different message types
+
+In uTransport listener API, we only have the `resource_id`` of the source and sink.
+However, sometimes it's necessary to distinguish the message types which should be listened to.
+
+Here is the table to map the various combination of source `resource_id` and sink `resource_id` to different message types.
+
+[%autowidth]
+|===
+| src resource_id | sink resource_id | Publish | Notification | Request | Response
+
+| [8000-FFFE] |   None   |    V    |              |         |          
+| [8000-FFFE] |     0    |         |      V       |         |          
+|      0      | [1-7FFF] |         |              |    V    |          
+|   [1-7FFF]  |     0    |         |              |         |    V     
+|     FFFF    |     0    |         |      V       |         |    V     
+|     FFFF    | [1-7FFF] |         |              |    V    |          
+|      0      |   FFFF   |         |              |    V    |          
+|   [1-7FFF]  |   FFFF   |         |              |         |    V     
+| [8000-FFFE] |   FFFF   |         |      V       |         |          
+|     FFFF    |   FFFF   |         |      V       |    V    |    V     
+|===
+
+These can be reorganized as:
+
+[%autowidth]
+|===
+| Message Type | Possible resource_id combinations {src_resource_id, sink_resource_id}
+
+| Publish      | {[8000-FFFE], None}
+| Notification | {[8000-FFFE], 0}, {[8000-FFFE], FFFF}, {FFFF, 0}, {FFFF, FFFF}
+| Request      | {0, [1-7FFF]}, {0, FFFF}, {FFFF, [1-7FFF]}, {FFFF, FFFF}
+| Response     | {[1-7FFF], 0}, {[1-7FFF], FFFF}, (FFFF, 0), {FFFF, FFFF}
+|===
--- a/basics/uattributes.adoc
+++ b/basics/uattributes.adoc
@@ -226,39 +226,3 @@ a|The link:uri.adoc[UUri] that the service consumer expects to receive this resp
 == Validation
 
 Each link:../languages.adoc[uProtocol Language Library] *MUST* provide means to assert the consistency of a message's attributes according to its type as defined in the sections above. The concrete implementation should follow common practices of the particular programming language.
-
-
-== Listeners for different message types
-
-In uTransport listener API, we only have the `resource_id`` of the source and sink.
-However, sometimes it's necessary to distinguish the message types which should be listened to.
-
-Here is the table to map the various combination of source `resource_id` and sink `resource_id` to different message types.
-
-[%autowidth]
-|===
-| src resource_id | sink resource_id | Publish | Notification | Request | Response
-
-| [8000-FFFE] |   None   |    V    |              |         |          
-| [8000-FFFE] |     0    |         |      V       |         |          
-|      0      | [1-7FFF] |         |              |    V    |          
-|   [1-7FFF]  |     0    |         |              |         |    V     
-|     FFFF    |     0    |         |      V       |         |    V     
-|     FFFF    | [1-7FFF] |         |              |    V    |          
-|      0      |   FFFF   |         |              |    V    |          
-|   [1-7FFF]  |   FFFF   |         |              |         |    V     
-| [8000-FFFE] |   FFFF   |         |      V       |         |          
-|     FFFF    |   FFFF   |         |      V       |    V    |    V     
-|===
-
-These can be reorganized as:
-
-[%autowidth]
-|===
-| Message Type | Possible resource_id combinations {src_resource_id, sink_resource_id}
-
-| Publish      | {[8000-FFFE], None}
-| Notification | {[8000-FFFE], 0}, {[8000-FFFE], FFFF}, {FFFF, 0}, {FFFF, FFFF}
-| Request      | {0, [1-7FFF]}, {0, FFFF}, {FFFF, [1-7FFF]}, {FFFF, FFFF}
-| Response     | {[1-7FFF], 0}, {[1-7FFF], FFFF}, (FFFF, 0), {FFFF, FFFF}
-|===
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -261,25 +261,25 @@ end
 deactivate T
 ----
 
-In RegisterListener API, we only have the `resource_id`` of the source and sink.
+In RegisterListener API, we only have the `resource_id` of the source and sink.
 However, sometimes it's necessary to distinguish the message types which should be listened to.
 
 Here is the table to map the various combination of source `resource_id` and sink `resource_id` to different message types.
 
 [%autowidth]
 |===
-| src resource_id | sink resource_id | Publish | Notification | Request | Response
+| source `resource_id` | sink `resource_id` | Publish | Notification | Request | Response | Use Case
 
-| [8000-FFFE] |   None   |    V    |              |         |          
-| [8000-FFFE] |     0    |         |      V       |         |          
-|      0      | [1-7FFF] |         |              |    V    |          
-|   [1-7FFF]  |     0    |         |              |         |    V     
-|     FFFF    |     0    |         |      V       |         |    V     
-|     FFFF    | [1-7FFF] |         |              |    V    |          
-|      0      |   FFFF   |         |              |    V    |          
-|   [1-7FFF]  |   FFFF   |         |              |         |    V     
-| [8000-FFFE] |   FFFF   |         |      V       |         |          
-|     FFFF    |   FFFF   |         |      V       |    V    |    V     
+| [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message 
+| [8000-FFFE] |     0    |         |      V       |         |         | A uE listens for a specific notification message
+|      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request 
+|   [1-7FFF]  |     0    |         |              |         |    V    | A uE listens for a specific response
+|     FFFF    |     0    |         |      V       |         |    V    | A uE listens for all notifications and responses 
+|     FFFF    | [1-7FFF] |         |              |    V    |         | Same as {0, [1-7FFF]}
+|      0      |   FFFF   |         |              |    V    |         | Same as {0, [1-7FFF]}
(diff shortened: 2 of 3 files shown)


PLeVasseur:
Hey @evshary -- is this generally useful to include in e.g. the [uAttributes](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc) spec?

That's what I had originally thought

gregmedd:
I agree that this seems like it belongs in the UAttributes (or even better, UUri) spec. It's not zenoh specific.

It would also be really helpful to have a similar table of the allowable values for both source and sink when sending a message.

up-l1/zenoh.adoc
```diff
@@ -129,3 +129,39 @@ Take some examples:
...
+
+[%autowidth]
+|===
+| Message Type | Possible resource_id combinations {src_resource_id, sink_resource_id}
+
+| Publish      | {[8000-FFFF), None}
+| Notification | {[8000-FFFF), 0}, {[8000-FFFF), FFFF]}, {FFFF, 0}, {FFFF, FFFF}
+| Request      | {0, (0-8000)}, {0, FFFF}, {FFFF, (0-8000)}, {FFFF, FFFF}
+| Response     | {(0-8000), 0}, {(0-8000), FFFF}, (FFFF, 0), {FFFF, FFFF}
+|===
```
sophokles73, line 167: this information is not specific to the Zenoh transport, is it? In that case, I would recommend to add it to a more generic document like [uattributes.adoc](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc)?

evshary:
@PLeVasseur @gregmedd @sophokles73 Thanks for your suggestion. I thought it was the only issue Zenoh faced, but it seems like we can view it more generically. I've moved it to `uattributes.adoc` now.

basics/uattributes.adoc
```diff
@@ -226,3 +226,39 @@ a|The link:uri.adoc[UUri] that the service consumer expects to receive this resp
 == Validation
 
 Each link:../languages.adoc[uProtocol Language Library] *MUST* provide means to assert the consistency of a message's attributes according to its type as defined in the sections above. The concrete implementation should follow common practices of the particular programming language.
+
+
+== Listeners for different message types
```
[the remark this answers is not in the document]
evshary: You're right. `up-l1/README.adoc` is a better place
Regarding the first table, I still think it's necessary, otherwise, it's not easy for newcomers to understand how we get the second table.

up-l1/README.adoc
```diff
@@ -261,6 +261,39 @@ end
 deactivate T
 ----
 
+In RegisterListener API, we only have the `resource_id`` of the source and sink.
```
stevenhartley, line 264:
```suggestion
In RegisterListener API, we only have the `resource_id` of the source and sink.
```

up-l1/README.adoc
```diff
@@ -261,6 +261,39 @@ end
...
+However, sometimes it's necessary to distinguish the message types which should be listened to.
+
+Here is the table to map the various combination of source `resource_id` and sink `resource_id` to different message types.
+
+[%autowidth]
+|===
+| src resource_id | sink resource_id | Publish | Notification | Request | Response
```
stevenhartley, line 271:
```suggestion
| source `resource_id` | sink `resource_id` | Publish | Notification | Request | Response
```

up-l1/README.adoc
```diff
@@ -261,6 +261,39 @@ end
...
+
+| [8000-FFFE] |   None   |    V    |              |         |          
+| [8000-FFFE] |     0    |         |      V       |         |          
+|      0      | [1-7FFF] |         |              |    V    |
```
stevenhartley, line 275: I don't think this is a valid usecase for source and sink matching

up-l1/README.adoc
```diff
@@ -261,6 +261,39 @@ end
...
+|===
+| src resource_id | sink resource_id | Publish | Notification | Request | Response
+
+| [8000-FFFE] |   None   |    V    |              |         |          
+| [8000-FFFE] |     0    |         |      V       |         |          
+|      0      | [1-7FFF] |         |              |    V    |          
+|   [1-7FFF]  |     0    |         |              |         |    V     
+|     FFFF    |     0    |         |      V       |         |    V     
+|     FFFF    | [1-7FFF] |         |              |    V    |          
+|      0      |   FFFF   |         |              |    V    |
```
stevenhartley, line 279: Another usecase that I think is not valid

stevenhartley (changes requested):
Made some comments but what I would also like to see @evshary is another column of the first table that explains the use case for said combination, by that I mean:

| src resource_id | sink resource_id | Publish | Notification | Request | Response| Use Case|
| ---------------- | ----------------- | --------- |--------- |--------- |--------- |--------- |
| [8000-FFFE] |   None   |    V    |              |     |    |  A uE listens for a specific published message|
| [8000-FFFE] |     0    |         |      V       |     |    |  A uE listens for a response to a specific notification |
|   [1-7FFF]  |     0    |         |              |         |    V  |  A uE listens for a response for a specific request |
|     FFFF    |     0    |         |      V       |         |    V  |  A uE listens for all notifications or responses |
|     FFFF    | [1-7FFF] |         |              |    V    |  | A service listens for requests coming in from anyone to a specific request |
|   [1-7FFF]  |   FFFF   |         |              |         |    V   |   | Streamer listens for all responses that came from anyone to anyone |
| [8000-FFFE] |   FFFF   |         |      V       |         |    | Streamer listens for any notification from anyone
|     FFFF    |   FFFF   |         |      V       |    V    |    V     |  Streamer listens for anything that is sent to this device

evshary:
@stevenhartley The idea of the table is to distinguish message type from different source and sink `resource_id` combinations. You gave a good suggestion to add use cases here, which I didn't think of before.

I've updated the table now. I think the `FFFF` of the `resource_id` indicates that we don't care about the value.
For the case `{0, FFFF}`, it should have the same meaning as `{0, [1-7FFF]}`, while `{0, [8000-FFFE]}` is not valid.
However, I agree that users might never use the way to listen to the `resource_id`. The same concept can be applied to index 6 to index 9. We can leave these combinations here (although users never use them) or perhaps we can forbid these combinations to make the table simpler. WDYT?

| index | source `resource_id` | sink `resource_id` | Publish | Notification | Request | Response | Use Case
| - | - | - | - | - | - |- | -
| 1 | [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message 
| 2 | [8000-FFFE] |     0    |         |      V       |         |         | A uE listens for a specific notification message
| 3 |      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request 
| 4 |   [1-7FFF]  |     0    |         |              |         |    V    | A uE listens for a specific response
| 5 |     FFFF    |     0    |         |      V       |         |    V    | A uE listens for all notifications and responses 
| 6 |     FFFF    | [1-7FFF] |         |              |    V    |         | Same as {0, [1-7FFF]}
| 7 |      0      |   FFFF   |         |              |    V    |         | Same as {0, [1-7FFF]}
| 8 |   [1-7FFF]  |   FFFF   |         |              |         |    V    | Same as {[1-7FFF], 0}
| 9 | [8000-FFFE] |   FFFF   |         |      V       |         |         | Same as {[8000-FFFE], 0}
| 10 |     FFFF    |   FFFF   |         |      V       |    V    |    V    | uStreamer listens for all notifications, requests, and responses

up-l1/README.adoc
evshary: I left the comments below. Please let me know your thoughts.

stevenhartley:
IMHO the table is just getting too complicated. We should only list the known examples that uEs (and/or the streamer) should use and then get rid of any of the combinations that don't make any sense.

evshary:
Then the table can be simplified as 

| index | source `resource_id` | sink `resource_id` | Publish | Notification | Request | Response | Use Case
| - | - | - | - | - | - |- | -
| 1 | [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message 
| 2 | [8000-FFFE] |     0    |         |      V       |         |         | A uE listens for a specific notification message
| 3 |      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request 
| 4 |   [1-7FFF]  |     0    |         |              |         |    V    | A uE listens for a specific response
| 5 |     FFFF    |     0    |         |      V       |         |    V    | A uE listens for all notifications and responses 
| 6 |     FFFF    |   FFFF   |         |      V       |    V    |    V    | uStreamer listens for all notifications, requests, and responses

These combinations only make sense to users, and we can forbid users from using other combinations explicitly.
If you agree with this, then I'll update the table.

stevenhartley:
> Then the table can be simplified as
> 
> index	source `resource_id`	sink `resource_id`	Publish	Notification	Request	Response	Use Case
> 1	[8000-FFFE]	None	V				A uE listens for a specific published message
> 2	[8000-FFFE]	0		V			A uE listens for a specific notification message
> 3	0	[1-7FFF]			V		A uE listens for a specific request
> 4	[1-7FFF]	0				V	A uE listens for a specific response
> 5	FFFF	0		V		V	A uE listens for all notifications and responses
> 6	FFFF	FFFF		V	V	V	uStreamer listens for all notifications, requests, and responses
> These combinations only make sense to users, and we can forbid users from using other combinations explicitly. If you agree with this, then I'll update the table.

@evshary I'm ok with the latest changes. Please update the PR and I will approve it

up-l1/README.adoc
```diff
@@ -261,6 +261,38 @@ end
...
+
+| [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message 
+| [8000-FFFE] |     0    |         |      V       |         |         | A uE listens for a specific notification message
+|      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request 
+|   [1-7FFF]  |     0    |         |              |         |    V    | A uE listens for a specific response
+|     FFFF    |     0    |         |      V       |         |    V    | A uE listens for all notifications and responses 
+|     FFFF    |   FFFF   |         |      V       |    V    |    V    | uStreamer listens for all notifications, requests, and responses
+|===
+
+Note that other combinations not in the table *MUST* never be used in RegisterListener API.
```
stevenhartley, line 281: Sorry @evshary but I'm reminded of what I actually implemented in up-java. The [InMemoryRpcClient.java](https://github.com/eclipse-uprotocol/up-java/blob/main/src/main/java/org/eclipse/uprotocol/communication/InMemoryRpcClient.java) will register a listener for source=FFFF, sink=0 so that it can get all responses. Additionally, [InMemoryRpcServer.java](https://github.com/eclipse-uprotocol/up-java/blob/main/src/main/java/org/eclipse/uprotocol/communication/InMemoryRpcServer.java#L100) will register a listener with source=FFFF and sink=[1-7FFFF] to be able to receive requests from any client for a specific method so we need to add those options to the table...

up-l1/README.adoc
```diff
@@ -261,6 +261,38 @@ end
...
+[%autowidth]
+|===
+| source `resource_id` | sink `resource_id` | Publish | Notification | Request | Response | Use Case
+
+| [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message 
+| [8000-FFFE] |     0    |         |      V       |         |         | A uE listens for a specific notification message
+|      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request 
+|   [1-7FFF]  |     0    |         |              |         |    V    | A uE listens for a specific response
+|     FFFF    |     0    |         |      V       |         |    V    | A uE listens for all notifications and responses 
+|     FFFF    |   FFFF   |         |      V       |    V    |    V    | uStreamer listens for all notifications, requests, and responses
```
stevenhartley, line 278: I'm not sure this is valid because it will them mean the streamer will register for listeners for itself. @PLeVasseur what are you passing to the transports now in up-streamer-rust?

stevenhartley (changes requested):

up-l1/README.adoc
PLeVasseur: Should be okay.

This is for the resource IDs only, which is a much more narrow scope. The uStreamer still has the ability to discriminate based on the authority_name.

Here's a [snippet of code](https://github.com/eclipse-uprotocol/up-streamer-rust/blob/d4ab17fbc745705290e29e5251fa8e9fcbffe943/up-streamer/src/ustreamer.rs#L158-L159) showing how we currently use `FFFF`, `FFFF` for source and sink resource IDs. You'd have to read up a bit to see the definitions of the functions for creating the UUris [here](https://github.com/eclipse-uprotocol/up-streamer-rust/blob/d4ab17fbc745705290e29e5251fa8e9fcbffe943/up-streamer/src/ustreamer.rs#L31-L49).

up-l1/README.adoc
evshary: For the case `source=FFFF, sink=0`, it's covered at index 5.
However, I'm not sure about `source=FFFF, sink=[1-7FFF]`. Could we use `source=0, sink=[1-7FFF]` to cover it?
I mean we can set Authority to any, but specify the `resource_id` must be 0. It doesn't make sense for any other `resource_id`.

```
UUri {
  authority_name: "*",         // any authority
  ue_id: 0x0000_FFFF           // any instance, any service
  ue_version_major: 0xFF,  // any
  resource_id: 0
}
```

I'm fine to add it back, but in that case, should we also add the index 7-9 back? I removed them because they can be covered by other cases.
image

evshary: Hi @stevenhartley, I forgot to tag you in the last comment. Let me know if you have any other ideas. Thank you!

stevenhartley: I think it is Ok for now. We will need to cleanup this API to make the API more explicit as the use of the API is still confusing (something for 1.7 or 2.0 later).

stevenhartley (approved):
LGTM

Issue #193, opened 2024-07-09
Clarify UTransport registerListener() & unregisterListener() when passed the same parameters


stevenhartley:
The uP-L1 uTransport specification does not explicitly state what implementations should do when you call the methods multiple times passing the same parameters. We can either make the API idempotent whereby the API simply returns `UCode.OK` or we make the API return a failure of `uCode.ALREADY_EXISTS`. This issue will be used to propose a change to the spec to pick the former and not the latter.

PR #194, opened 2024-07-09
Add idempotent requirement to UTransport APIs.

stevenhartley:
Calling the UTransport register and unregister API multiple times with the same parameters return the same value as if you called it the first time in lieu of returning  `UCode.ALREADY_EXISTS`.

#193

### DIFF
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -237,6 +237,7 @@ UTransport implementations
 * *MUST* support registering more than one listener for any given address patterns
 * *MUST* support registering the same listener for multiple address patterns
 * *MUST* document the maximum supported number of listeners per address pattern.
+* *MUST* be idempotent, multiple calls to said API with the same parameters *MUST* have the same effect as a single call.
 
 .Registering a Listener
 [mermaid]
@@ -294,6 +295,7 @@ UTransport implementations
 
 * *MUST* fail with a `UCode.UNIMPLEMENTED` if the transport does not support the _push_ <<delivery-method>>. In that case, the <<register-listener>> method *MUST* also fail accordingly.
 * *MUST* fail with a `UCode.NOT_FOUND`, if no such listener had been registered before
+* *MUST* be idempotent, multiple calls to said API with the same parameters *MUST* have the same effect as a single call.
 
 .Unregistering a Listener
 [mermaid]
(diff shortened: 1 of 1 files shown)


up-l1/README.adoc
```diff
@@ -237,6 +237,7 @@ UTransport implementations
 * *MUST* support registering more than one listener for any given address patterns
 * *MUST* support registering the same listener for multiple address patterns
 * *MUST* document the maximum supported number of listeners per address pattern.
+* *MUST* be idempotent, multiple calls to said API with the same parameters *MUST* have the same effect as a single call.
```
PLeVasseur, line 240: Practically speaking, does this mean just that if there were some sort of error other than ALREADY_EXISTS, we'd return that error and if we would have returned ALREADY_EXISTS, return OK?

PLeVasseur (commented):
Just wanna clarify the implementation

up-l1/README.adoc
```diff
@@ -294,6 +295,7 @@ UTransport implementations
 
 * *MUST* fail with a `UCode.UNIMPLEMENTED` if the transport does not support the _push_ <<delivery-method>>. In that case, the <<register-listener>> method *MUST* also fail accordingly.
 * *MUST* fail with a `UCode.NOT_FOUND`, if no such listener had been registered before
+* *MUST* be idempotent, multiple calls to said API with the same parameters *MUST* have the same effect as a single call.
```
dddj698, line 298: The following call would return NOT_FOUND if initial call success?

stevenhartley: OMG right (so dumb) fixed it now thanks @dddj698 !

up-l1/README.adoc
sophokles73: If we want all calls to be idempotent then IMHO `unregister` should never return `NOT_FOUND`, or should it?

PLeVasseur: That makes sense to me @sophokles73. If it succeeds or is not found we still just return `OK`.

stevenhartley: I think there is value in the unregisterListener returning NOT_FOUND if you try and unregister something that is not there (that is an error) but registering the same listener twice just has the same affect as doing it once.

up-l1/README.adoc
stevenhartley: Removed the unregisterListener idempotent requirement

sophokles73 (approved):
LGTM

Issue #192, opened 2024-06-26
MQTT 5 uTransport topic structure needs to be revisited for d2c & c2d connections


devkelley:
This issue was raised in the `uprotocol-mqtt` slack channel:

"a critical flaw in our D2C and C2D topic definitions that there will be too many topics across the millions of vehicles and hundreds of topics per-vehicle if we events are published to the specific topics following: {client_identifier}/{source.authority_name}/{source.ue_id}/{source.ue_version_major}/{source.resource_id}/{sink.authority_name}/{sink.ue_id}/{sink.ue_version_major}/{sink.resource_id}
To simplify the communication (and to drastically reduce the # of topics the broker needed to manage), we used a single D2C and C2D topic that was more like:  {client_identifier}/{sink.authority_name}/ and the sink authority was added to published events sent to the vehicles (so we can ensure all events made it to the vehicles)."

Desired outcome: A MQTT topic structure that allows for authorization controls at the broker level, conforms to the segment limitations of cloud brokers, and reduces the number of unique topics for each vehicle.

Please use this issue to discuss use cases and proposals.

stevenhartley:
Hi @jjj-vtm, What I recall the concern was not about the number of subscriptions as you pointed out we subscribe to a lot of wildcard sections of the topic, but the amount of individual publish topics. I don't understand the problem so I will ask them to comment on this issue now.

jjj-vtm:
Hi @stevenhartley, ok what I can think of and this is a real problem is when for example every ue would instantiate a UTransport (MQTT) on its own and subscribe to the wild card topic. This would lead to many connections and many unnecessary messages. 

Maybe it is already considered in the uProtocol but a Gateway component should solve this issue. For MQTT one would have an internal MQTT broker where each ue subscribes to the very specific topic it wants to listen too and the Gateway would be an UStreamer (MQTT <--> MQTT) which subscribes to the wildcard topic on the cloud broker an forwards all messages to the internal broker. With this topology one would only need a single MQTT connection to the cloud broker and it also has the benefit that one might think of more sophisticated authorization checks on the gateway.

stevenhartley:
That is what we have but we are questioning if the topic structure can be the same across the brokers so that the utransport implementations can remain the same.

stevenhartley:
Below is the information that was placed in #201 for the solution to this issue:

[...]

Issue #169, opened 2024-05-30
Add OFT (Open Fast Trace) metadata to Specifications


stevenhartley:
The following issue shall be used to insert OFT tags to the specifications using uri.adoc as a reference.

PR #204, opened 2024-07-18
Updates to uattributes and MQTT5 specifications

stevenhartley:
The following changes adds D2D usecase for the MQTT5 transport, fixes the uE-2-uE use case, and adds OFT tags to uatributes and MQTT5 specs.

#192, #169

### DIFF
--- a/basics/uattributes.adoc
+++ b/basics/uattributes.adoc
@@ -60,7 +60,7 @@ The following table defines attributes that are used for all message types:
 |link:uuid.adoc[UUID]
 |yes
 a|
-[.specitem,oft-sid="req~up-attributes-id~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-id~1",oft-needs="impl,utest"]
 --
 Every message *MUST* have a unique message identifier. The timestamp contained in the UUID is used as the message's creation time.
 --
@@ -70,7 +70,7 @@ Every message *MUST* have a unique message identifier. The timestamp contained i
 |no
 a|
 The QoS level that this message should be processed/delivered with. 
-[.specitem,oft-sid="req~up-attributes-priority~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-priority~1",oft-needs="impl,utest"]
 --
 A consumer *MUST* assume a message to have the link:qos.adoc#default-priority[default priority] if not set explicitly.
 --
@@ -80,11 +80,11 @@ A consumer *MUST* assume a message to have the link:qos.adoc#default-priority[de
 |no
 a|
 The amount of time (in milliseconds) after which this message is no longer delivered/processed.
-[.specitem,oft-sid="req~up-attributes-ttl-timeout~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-ttl-timeout~1",oft-needs="impl,utest"]
 --
 A consumer *MUST NOT* process an expired message. A message is expired if the TTL is > `0` and the consumer's current time is _after_ the message's creation time _plus_ its TTL.
 --
-[.specitem,oft-sid="req~up-attributes-ttl~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-ttl~1",oft-needs="impl,utest"]
 --
 A consumer *MUST* not consider a message expired if it has a TTL value of `0`.
 --
@@ -98,7 +98,7 @@ stem:[t_{current} > t_{id} + ttl]
 a|
 A tracing identifier to use for correlating messages across the system. Intended to be compatible with https://github.com/cloudevents/spec/blob/main/cloudevents/extensions/distributed-tracing.md[CloudEvents distributed tracing extension].
 
-[.specitem,oft-sid="req~up-attributes-traceparent~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-traceparent~1",oft-needs="impl,utest"]
 --
 * *MAY* contain traceparent attribute.
 --
@@ -108,7 +108,7 @@ A tracing identifier to use for correlating messages across the system. Intended
 |link:upayloadformat.adoc[UPayloadFormat]
 |no
 a|The format of the payload (data).
-[.specitem,oft-sid="req~up-attributes-payload-format~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-payload-format~1",oft-needs="impl,utest"]
 --
 * *MAY* contain payload format attribute.
 --
@@ -127,7 +127,7 @@ The following table defines attributes that are used for publish messages in add
 |`type`
 |yes
 a| 
-[.specitem,oft-sid="req~up-attributes-publish-type~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-publish-type~1",oft-needs="impl,utest"]
 --
 *MUST* be set to `UMESSAGE_TYPE_PUBLISH`.
 --
@@ -136,7 +136,7 @@ a|
 |yes
 a|A link:uri.adoc[UUri] representing the topic that this message is published to.
 
-[.specitem,oft-sid="req~up-attributes-publish-source~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-publish-source~1",oft-needs="impl,utest"]
 --
 * The UUri's `resource_id` *MUST* be set to a value in range `[0x8000, 0xFFFE]`.
 --
@@ -146,7 +146,7 @@ a|A link:uri.adoc[UUri] representing the topic that this message is published to
 a|A link:uri.adoc[UUri] representing the receiver of this published message. This information is 
 only required when we are sending the published event between devices through the streamer, publishers are not required to populate this information.
 
-[.specitem,oft-sid="req~up-attributes-publish-sink~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-publish-sink~1",oft-needs="impl,utest"]
 --
 * *MAY* be present in the message. If present, the UUri *MUST* only contain the `authority_name` field.
 --
@@ -165,7 +165,7 @@ The following table defines attributes that are used for notification messages i
 |`type`
 |yes
 a|
-[.specitem,oft-sid="req~up-attributes-notification-type~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-notification-type~1",oft-needs="impl,utest"]
 --
 *MUST* be set to `UMESSAGE_TYPE_NOTIFICATION`.
 --
@@ -174,7 +174,7 @@ a|
 a|
 A link:uri.adoc[UUri] representing the component that this notification originates from.
 
-[.specitem,oft-sid="req~up-attributes-notification-source~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-notification-source~1",oft-needs="impl,utest"]
 --
 * The UUri's `resource_id` *MUST* be set to a value in range `[0x8000, 0xFFFE]`.
 --
@@ -183,7 +183,7 @@ A link:uri.adoc[UUri] representing the component that this notification originat
 |yes
 a|A link:uri.adoc[UUri] representing the receiver of this notification.
 
-[.specitem,oft-sid="req~up-attributes-notification-sink~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-notification-sink~1",oft-needs="impl,utest"]
 --
 * The UUri's `resource_id` *MUST* be set to `0`.
 --
@@ -202,7 +202,7 @@ The following table defines attributes that are used for RPC request messages in
 |`type`
 |yes
 a|
-[.specitem,oft-sid="req~up-attributes-request-type~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-request-type~1",oft-needs="impl,utest"]
 --
 *MUST* be set to `UMESSAGE_TYPE_REQUEST`.
 --
@@ -211,7 +211,7 @@ a|
 |yes
 a|The link:uri.adoc[UUri] that the service consumer expects to receive the response message at.
 
-[.specitem,oft-sid="req~up-attributes-request-source~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-request-source~1",oft-needs="impl,utest"]
 --
 * The UUri's `resource_id` *MUST* be set to `0`.
 --
@@ -220,7 +220,7 @@ a|The link:uri.adoc[UUri] that the service consumer expects to receive the respo
 |yes
 a|A link:uri.adoc[UUri] identifying the service provider's method to invoke.
 
-[.specitem,oft-sid="req~up-attributes-request-sink~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~up-attributes-request-sink~1",oft-needs="impl,utest"]
 --
 * The UUri's `resource_id` *MUST* be set to a value in range `[1, 0x7FFF]`.
 --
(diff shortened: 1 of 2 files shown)


tamarafischer (approved):
I liked the idea and find it useful.
I am guessing that this kind of patter can be used for any broker - all the interface would have to define are the wildcard characters and we can easily build the topics.

basics/uattributes.adoc
```diff
@@ -59,29 +59,57 @@ The following table defines attributes that are used for all message types:
 |`id`
 |link:uuid.adoc[UUID]
 |yes
-|A unique message identifier. The timestamp contained in the UUID is used as the message's creation time.
+a|
+[.specitem,oft-sid="req~up-attributes-id~1",oft-needs="impl,utest"]
```
sophokles73, line 63: I wonder if this really is the right place to specify this requirement. It seems quite unclear, where and at what time this would need to be verified. Maybe it would be better to define this in the Transport layer API spec? There we could require that all messages that are sent via a transport must have valid attributes according to the UAttributes spec and we could also define what needs to be done with messages received that are not compliant, e.g. ignore them.

I guess that would also reduce the number of spec items significantly and make it much easier to implement and test the requirements.

WDYT?

up-l1/mqtt_5.adoc
```diff
@@ -24,165 +24,215 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
 
 For more information, please refer to https://mqtt.org/
 
-This document will discuss the uTransport implementation on MQTT 5.
+This document will discuss the uTransport implementation on MQTT 5 for two well known use cases:
+1. *Device-2-Device (D2D) Communication:* This is when devices (through the streamer) was to send messages to other devices through a cloud gateway (broker). 
+2. *uE-2-uE Communication:* When MQTT is used as a local software bus for uEs to talk to each other through a local broker.
 
-== Specifications
+The reason fo the different configurations is that cloud MQTT brokers often have limitations on the number of topics that can be subscribed to as well as the topic segment lengths so we need to ensure that message routing is as efficient as possible.
```
sophokles73, line 31:
```suggestion
The reason for the different configurations is that cloud MQTT brokers often have limitations on the number of topics that can be subscribed to as well as the topic segment lengths so we need to ensure that message routing is as efficient as possible.
```

up-l1/mqtt_5.adoc
```diff
@@ -24,165 +24,215 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+== MQTT5 Header
 
-MQTT 5 supports custom header fields, and we leverage this to map the UAttributes values into the MQTT header.
+MQTT 5 supports custom header fields, and we leverage this to map the `UAttributes` values into the `UserProperty` MQTT header in string format.
 
-[cols="1,1,1,1"]
+[.specitem,oft-sid="req~up-transport-mqtt5-attributes~1",oft-needs="impl,utest"]
+--
+* `UserProperty` Keys *MUST* map to the UAttributes protobuf field numbers.
+--
```
sophokles73, line 40: FMPOV there is no need for this as you are explicitly listing all attributes in the following table, along with the keys to use.

However, what we probably should define is that only non-empty attribute values should be mapped to MQTT header fields.

up-l1/mqtt_5.adoc
```diff
@@ -24,165 +24,215 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+=== uE-2-uE Communication
 
-The client_identifier that is used for the topic is based on the kind of MQTT uTransport client that is sending the message. An MQTT uTransport client can be one of the following:
+When a message is sent from one uE to another uE, the topic is constructed using the source and sink UUri fields. The topic is constructed as follows using the string representation of a `UUri`:
 
-- Cloud MQTT uTransport client: A client that runs in the cloud.
-- Remote MQTT uTransport client: A client that runs on a device and is connected to a cloud MQTT broker.
+[.specitem,oft-sid="req~up-transport-mqtt5-ue2ue-topic~1",oft-needs="impl,utest"]
+--
+`{UAttributes.source}/{UAttributes.sink}/`
```
sophokles73, line 186: what is the purpose of the trailing `/` character?

sophokles73: also, what you have specified means that the topic always starts with a `/` character, which is unusual from my experience

up-l1/mqtt_5.adoc
```diff
@@ -24,165 +24,215 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
 |===
-| source | sink | topic
+| Type| source URI | sink URI | MQTT5 Topic
 
-| //vehicle_1.veh.com/AB34/1/12CD | //vehicle_2.veh.com/43BA/1/DC21 | {cli_identifier}/vehicle_1.veh.com/AB34/1/12CD/vehicle_2.veh.com/43BA/1/DC21
-| //vehicle_1.veh.com/AB34/1/12CD | None | {cli_identifier}/vehicle_1.veh.com/AB34/1/12CD
+| *Request* | `//device1/AB34/1/0` | `//device1/43BA/1/2` | `device1/AB34/1/0/device1/43BA/1/2`
+| *Response* | `//device1/43BA/1/2` | `//device1/AB34/1/0` | `device1/43BA/1/2/device1/AB34/1/0`
+| *Publish* | `//device1/AB34/1/8000` | None | `device1/AB34/1/8000`
+| *Notification* | `//device1/43BA/1/8001` | `//device1/AB34/1/0` | `device1/43BA/1/8001/device1/AB34/1/0`
```
sophokles73, line 204: these examples do not match the topic pattern defined above

sophokles73 (commented):
In general, you should consider adding a `oft-title` attribute to the spec items which contains a short, human readable summary of the spec item. This will make the generated tracing reports much easier to read/understand ...

up-l1/mqtt_5.adoc
stevenhartley: Removed the trailing '/'. Topic won't start with '/' as every message will always have a source but not always a sink.

up-l1/mqtt_5.adoc
sophokles73: with a source of `//device1/43BA/1/8001` and a sink of `//device1/AB34/1/0` and a pattern of `$(source}/${sink}`, the topic should be `//device1/43BA/1/8001///device1/AB34/1/0`, shouldn't it? However, in the table, you are using `device1/43BA/1/8001/device1/AB34/1/0`.

Even considering that the string representation of a UUri without an authority name actually is not `//device1/43BA/1/8001` but `/device1/43BA/1/8001`, applying the pattern would yield `/device1/AB34/1/8000//device1/AB34/1/0` ...

basics/uattributes.adoc
sophokles73: Sure, but may I at least suggest to turn them into design spec items (using the `dsn` type instead of `req`) as these are not really user level requirements but are the outcome of having thought about how to make it possible ...

up-l1/mqtt_5.adoc
```diff
@@ -24,165 +24,215 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
-=== UAttributes Mapping
+== MQTT5 Header
 
-MQTT 5 supports custom header fields, and we leverage this to map the UAttributes values into the MQTT header.
+MQTT 5 supports custom header fields, and we leverage this to map the `UAttributes` values into the `UserProperty` MQTT header in string format.
 
-[cols="1,1,1,1"]
+[.specitem,oft-sid="req~up-transport-mqtt5-attributes-non-empty~1",oft-needs="impl,utest"]
+--
+* Only non-empty attribute values *MUST* be mapped to MQTT header `UserProperty` keys.
```
sophokles73, line 39:
```suggestion
* All non-empty attribute values *MUST* be mapped to MQTT header *User Properties* using the keys defined below.
* Empty attribute values *MUST NOT* be mapped.
```

basics/uattributes.adoc
stevenhartley: I guess it depends on what your definitions of feature vs user vs design vs architecture requirements are.  I got the following response from chatGPT:

1. *Feature Requirements:* These are specific functionalities or capabilities that a product or system should have. Feature requirements describe what the product should do or the tasks it should perform to meet user needs or business objectives. They focus on the desired outcomes or features that users expect from the product.

2. *User Requirements:* User requirements capture the needs, expectations, and goals of the end-users or stakeholders who will interact with the product. They define the user's perspective, including their desired functionalities, usability, user experience, and any specific constraints or preferences they may have.

3. *Design Requirements:* Design requirements focus on the visual and interactive aspects of the product. They specify how the product should look, feel, and behave from a user interface (UI) and user experience (UX) perspective. Design requirements may include guidelines for layout, color schemes, typography, navigation, and other visual elements that contribute to the overall design aesthetics and usability.

4. *Architecture Requirements:* Architecture requirements pertain to the underlying structure, components, and technical aspects of the product's system architecture. They define the high-level design principles, performance criteria, scalability, security, integration requirements, and other technical considerations necessary for the product's development and deployment. Architecture requirements guide the development team in designing a robust and scalable system that meets the desired technical specifications.

If we stick with design meaning UI/UX above, we should change the various requirements to `arch` . 

Having said that we do need to agree on definitions of feature vs user vs design vs architecture, my vote is using the definitions above.

up-l1/mqtt_5.adoc
stevenhartley: OK I see your point. We cannot simply append the strings so I will go back to listing each attribute independently.

basics/uattributes.adoc
sophokles73: OFT uses [this definition](https://github.com/itsallcode/openfasttrace/blob/main/doc/user_guide.md#specification-item-artifact-type) where, according to my understanding, `dsn` does not refer to UX requirements ...

up-l1/mqtt_5.adoc
stevenhartley: done

up-l1/mqtt_5.adoc
```diff
@@ -24,165 +24,216 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
...
+
+[.specitem,oft-sid="req~up-transport-mqtt5-d2d-topic~1",oft-needs="impl,utest"]
+--
+`{source.authority_name}/{sink.authority_name}`
+--
+
+==== Examples
+
+===== Registering Listener to Receive All Messages 
+`UTransport::registerListener(ANY, getSource())` where getSource(0 returns the local device's UUri, this translates into the MQTT subscribe to topic `+/{my_source_authority_name}`
```
PLeVasseur, line 233:
```suggestion
`UTransport::registerListener(ANY, getSource())` where getSource() returns the local device's UUri, this translates into the MQTT subscribe to topic `+/{my_source_authority_name}`
```

basics/uattributes.adoc
```diff
@@ -96,13 +124,31 @@ The following table defines attributes that are used for publish messages in add
...
 
+[.specitem,oft-sid="req~up-attributes-publish-source~1",oft-needs="impl,utest"]
+--
 * The UUri's `resource_id` *MUST* be set to a value in range `[0x8000, 0xFFFE]`.
+--
+
+|`sink`
+|no
+a|A link:uri.adoc[UUri] representing the receiver of this published message. This information is 
+only required when we are sending the published event between devices through the streamer, publishers are not required to populate this information.
```
PLeVasseur, line 145:
```suggestion
only required when we are sending the published event between devices through a streamer. A uEntity sending a Publish message is not required to populate this information.
```

Clarification ideas to separate a typical uE sending a publish message vs a streamer sending a publish message

basics/uattributes.adoc
```diff
@@ -96,13 +124,31 @@ The following table defines attributes that are used for publish messages in add
...
 |yes
 a|A link:uri.adoc[UUri] representing the topic that this message is published to.
 
+[.specitem,oft-sid="req~up-attributes-publish-source~1",oft-needs="impl,utest"]
+--
 * The UUri's `resource_id` *MUST* be set to a value in range `[0x8000, 0xFFFE]`.
+--
+
+|`sink`
+|no
```
PLeVasseur, line 143:
```suggestion
|no (see Description)
```

Highlight that it's _usually_ no, but some details need be understood.

basics/uattributes.adoc
```diff
@@ -59,29 +59,57 @@ The following table defines attributes that are used for all message types:
...
+
 
 |`payload format`
 |link:upayloadformat.adoc[UPayloadFormat]
 |no
-|The format of the payload (data).
+a|The format of the payload (data).
+[.specitem,oft-sid="req~up-attributes-payload-format~1",oft-needs="impl,utest"]
+--
+* *MAY* contain payload format attribute.
```
PLeVasseur, line 111: I would have assumed this is necessary, even if just to write that it's unspecified.

Also according to the .proto it's [not optional](https://github.com/eclipse-uprotocol/up-spec/blob/8801cc68be1356df6e861e56aa824fbdad853d54/up-core-api/uprotocol/v1/uattributes.proto#L66).

up-l1/mqtt_5.adoc
```diff
@@ -24,165 +24,216 @@ MQTT is an OASIS standard messaging protocol for the Internet of Things (IoT). I
 
 For more information, please refer to https://mqtt.org/
 
-This document will discuss the uTransport implementation on MQTT 5.
+This document will discuss the uTransport implementation on MQTT 5 for two well known use cases:
+1. *Device-2-Device (D2D) Communication:* This is when devices (through the streamer) was to send messages to other devices through a cloud gateway (broker).
```
PLeVasseur, line 28: From my reading then a uStreamer would only need put the MQTT5 transport into a special mode and do nothing else special.

Is that correct?

PLeVasseur (commented):
Overall makes sense. Had a couple of suggestions and a question.

basics/uattributes.adoc
sophokles73: It is optional, all message/enum typed fields in proto3 are per se optional, i.e. they do not need to be marked with the `optional` modifier.
For an enum, if no value is set, the enum's default value is assumed.

basics/uattributes.adoc
```diff
@@ -59,29 +59,57 @@ The following table defines attributes that are used for all message types:
...
 |`priority`
 |link:qos.adoc[UPriority]
 |no
-|The QoS level that this message should be processed/delivered with. If not specified explicitly, the link:qos.adoc#default-priority[default priority level] is assumed.
-
+a|
+The QoS level that this message should be processed/delivered with. If not specified explicitly, the link:qos.adoc#default-priority[default priority level] is assumed.
+[.specitem,oft-sid="req~up-attributes-priority~1",oft-needs="impl,utest"]
+--
+* *MUST* have a priority attribute.
```
sophokles73, line 75: just two lines above, we say that it does not need to have one, aren't we? I would also suggest to require the attribute to have a value instead of the message having the attribute. FMPOV a message always has a _priority_ attribute, the attribute just doesn't always have (to have) a value set explicitly, and thus assume a default value.

I would suggest something like:
```
[.specitem,oft-sid="req~up-attributes-priority~1",oft-needs="impl,utest"]
--
A consumer *MUST* assume a message to have the link:qos.adoc#default-priority[default priority] if not set explicitly.
--

File: basics/uattributes.adoc [hunk 13]
```diff
@@ -59,29 +59,57 @@ The following table defines attributes that are used for all message types:
...
+a|
+The amount of time (in milliseconds) after which this message is no longer delivered/processed.
+[.specitem,oft-sid="req~up-attributes-ttl~1",oft-needs="impl,utest"]
+--
+* *MUST* have a TTL attribute.
+--
+[.specitem,oft-sid="req~up-attributes-ttl-timeout~1",oft-needs="impl,utest"]
+--
+* The message *MUST* expire if the current time is greater than the sum of the message's creation time and the TTL value.
+--
```
sophokles73, line 89:
```suggestion
The amount of time (in milliseconds) after which this message is no longer delivered/processed.
[.specitem,oft-sid="req~up-attributes-ttl-timeout~1",oft-needs="impl,utest"]
--
A consumer *MUST NOT* process an expired message. A message is expired if the TTL is > `0` and the consumer's current time is _after_ the message's creation time _plus_ its TTL.
--
[.specitem,oft-sid="req~up-attributes-ttl~1",oft-needs="impl,utest"]
--
A consumer *MUST* not consider a message expired if it has a TTL value of `0`.
--

```

basics/uattributes.adoc
stevenhartley: changed

basics/uattributes.adoc
stevenhartley: done

stevenhartley (commented):
All feedback has been applied

sophokles73 (approved):
LGTM

devkelley (approved):

PR #207, opened 2024-07-22
Use relative paths in gitignore

sophokles73:
Changed all absolute paths in gitignore so that the paths are
also being ignored when including up-spec as a git submodule from
other repos (as we do from up-rust).


stevenhartley (approved):
LGTM

PR #210, opened 2024-07-31
Update the table of the resource ID combination for register_listener.

evshary:
As [the issue](https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/issues/60) mentioned, users tend to use source ID `0xFFFF` instead of `0` while registering to all possible requests. Since it often blocks users and costs lots of time to debug, I think it might be better to allow using `0xFFFF` in this case.

@stevenhartley @AnotherDaniel @neelam-kushwah Please let me know your thoughts.

### DIFF
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -275,8 +275,7 @@ The following table are the valid combinations, which maps to the corresponding 
 | [8000-FFFE] |     0    |         |      V       |         |         | A uE listens for a specific notification message
 |      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request
 |   [1-7FFF]  |     0    |         |              |         |    V    | A uE listens for a specific response
-|     FFFF    |   None   |    V    |              |         |         | A uE listens for all published messages
-|     FFFF    | [1-7FFF] |         |              |    V    |         | A uE listens for all requests
+|     FFFF    | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request of 1-7FFF
 |     FFFF    |     0    |         |      V       |         |    V    | A uE listens for all notifications and responses
 |     FFFF    |   FFFF   |         |      V       |    V    |    V    | uStreamer listens for all notifications, requests, and responses
 |===
@@ -290,7 +289,7 @@ We can get the message types by reorganizing the table above to the following on
 |===
 | Message Type | Possible resource_id combinations {src_resource_id, sink_resource_id}
 
-| Publish      | {[8000-FFFE], None}, {FFFF, None}
+| Publish      | {[8000-FFFE], None}
 | Notification | {[8000-FFFE], 0}, {FFFF, 0}, {FFFF, FFFF}
 | Request      | {0, [1-7FFF]}, {FFFF, [1-7FFF]}, {FFFF, FFFF}
 | Response     | {[1-7FFF], 0}, (FFFF, 0), {FFFF, FFFF}
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -271,11 +271,13 @@ The following table are the valid combinations, which maps to the corresponding 
 |===
 | source `resource_id` | sink `resource_id` | Publish | Notification | Request | Response | Use Case
 
-| [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message 
+| [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message
 | [8000-FFFE] |     0    |         |      V       |         |         | A uE listens for a specific notification message
-|      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request 
+|      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request
 |   [1-7FFF]  |     0    |         |              |         |    V    | A uE listens for a specific response
-|     FFFF    |     0    |         |      V       |         |    V    | A uE listens for all notifications and responses 
+|     FFFF    |   None   |    V    |              |         |         | A uE listens for all published messages
+|     FFFF    | [1-7FFF] |         |              |    V    |         | A uE listens for all requests
+|     FFFF    |     0    |         |      V       |         |    V    | A uE listens for all notifications and responses
 |     FFFF    |   FFFF   |         |      V       |    V    |    V    | uStreamer listens for all notifications, requests, and responses
 |===
 
@@ -288,9 +290,9 @@ We can get the message types by reorganizing the table above to the following on
 |===
 | Message Type | Possible resource_id combinations {src_resource_id, sink_resource_id}
 
-| Publish      | {[8000-FFFE], None}
+| Publish      | {[8000-FFFE], None}, {FFFF, None}
 | Notification | {[8000-FFFE], 0}, {FFFF, 0}, {FFFF, FFFF}
-| Request      | {0, [1-7FFF]}, {FFFF, FFFF}
+| Request      | {0, [1-7FFF]}, {FFFF, [1-7FFF]}, {FFFF, FFFF}
 | Response     | {[1-7FFF], 0}, (FFFF, 0), {FFFF, FFFF}
 |===
 


gregmedd:
To be honest, I don’t find that those tables lend much clarity on the matter at all, and adding more to them doesn’t help. Recommendations on allowable ranges for UUri fields only had meaning within the context of message attributes in outgoing messages where 1) wildcards are not allowed (since it’s not a filter) and 2) a message type is present to determine which set of rules apply.

When talking about filters for `registerListener()`, the is no meaningful way to reliably derive what message type that filter will match to solely by inspecting the UUris. We just had a major refactor around this in up-cpp - once wildcard fields and overlapping value ranges (e.g. publish and notify resources), the only thing that can be determined is the number of filters set (source only or source and sink).

If there is a need to provide a hint about the expected message type that will correspond to a listener registration, then that type should be an additional parameter passed from layer 2 to layer 1; the layer 2 module knows what to expect and could easily pass that information along.

evshary:
@gregmedd Thank you for the precious feedback.
Yes, if we have an additional parameter from L2, it helps a lot. However, in the current L1 API design, we don't have such argument. That's why we need to figure out the message type based on the source and sink resource ID.
I couldn't get your idea why it makes only sense if wildcard is not allowed. Could you elaborate more?

From Zenoh's perspective, I need a way to distinguish the message type and then apply it to different Zenoh mechanisms. Even if the UUri is wildcard, I still have to tell the intention of the message type from the combination of UUri. Otherwise, I couldn't map it to Zenoh implementation. Do you have any thoughts that we could have another way to get the message type?

up-l1/README.adoc
```diff
@@ -271,11 +271,13 @@ The following table are the valid combinations, which maps to the corresponding
...
 
-| [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message 
+| [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message
 | [8000-FFFE] |     0    |         |      V       |         |         | A uE listens for a specific notification message
-|      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request 
+|      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request
 |   [1-7FFF]  |     0    |         |              |         |    V    | A uE listens for a specific response
-|     FFFF    |     0    |         |      V       |         |    V    | A uE listens for all notifications and responses 
+|     FFFF    |   None   |    V    |              |         |         | A uE listens for all published messages
+|     FFFF    | [1-7FFF] |         |              |    V    |         | A uE listens for all requests
```
stevenhartley, line 279:
```suggestion
|     FFFF    | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request of 1-7FFF
```

up-l1/README.adoc
```diff
@@ -271,11 +271,13 @@ The following table are the valid combinations, which maps to the corresponding
...
 | source `resource_id` | sink `resource_id` | Publish | Notification | Request | Response | Use Case
 
-| [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message 
+| [8000-FFFE] |   None   |    V    |              |         |         | A uE listens for a specific published message
 | [8000-FFFE] |     0    |         |      V       |         |         | A uE listens for a specific notification message
-|      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request 
+|      0      | [1-7FFF] |         |              |    V    |         | A uE listens for a specific request
 |   [1-7FFF]  |     0    |         |              |         |    V    | A uE listens for a specific response
-|     FFFF    |     0    |         |      V       |         |    V    | A uE listens for all notifications and responses 
+|     FFFF    |   None   |    V    |              |         |         | A uE listens for all published messages
```
stevenhartley, line 278: Don't think this is valid to listen to all published messages, that would mean everything is sent to the uE

up-l1/README.adoc
```diff
@@ -288,9 +290,9 @@ We can get the message types by reorganizing the table above to the following on
 |===
 | Message Type | Possible resource_id combinations {src_resource_id, sink_resource_id}
 
-| Publish      | {[8000-FFFE], None}
+| Publish      | {[8000-FFFE], None}, {FFFF, None}
```
stevenhartley, line 293: No, we do not listen for all published events. This cannot be supported

up-l1/README.adoc
```diff
@@ -288,9 +290,9 @@ We can get the message types by reorganizing the table above to the following on
...
 | Notification | {[8000-FFFE], 0}, {FFFF, 0}, {FFFF, FFFF}
-| Request      | {0, [1-7FFF]}, {FFFF, FFFF}
+| Request      | {0, [1-7FFF]}, {FFFF, [1-7FFF]}, {FFFF, FFFF}
```
stevenhartley, line 295: I think the {ANY, ANY} for request is not valid

stevenhartley (changes requested):
please see comments

up-l1/README.adoc
evshary: So we don't need this in uStreamer case? If not, I will remove this.
Tag @PLeVasseur for more comments

up-l1/README.adoc
evshary: I thought this was also used in ustreamer. @PLeVasseur would you mind helping us double check?

up-l1/README.adoc
stevenhartley: Not for publish no.

PLeVasseur: Hi @evshary -- sorry I missed this. It's true. We do not listen for all Publish messages.

We have to use the uSubscription's notification of which topics should be forwarded out / listened for in order to select them one-by-one by a specific topic UUri.

This is because there is no destination, so if we listened for (and forwarded) all Publish messages, we could find all devices getting flooded by messages they did have an subscribing uEs on.

up-l1/README.adoc
PLeVasseur: This line is not used in the uStreamer.

We don't care what the resource IDs are for either the sink or source on Request/Response/Notification messages are when we are routing them through.

So that means the really important line for the uStreamer is that last line:
> uStreamer listens for all notifications, requests, and responses

See where we add a "forwarding listener" [here](https://github.com/eclipse-uprotocol/up-streamer-rust/blob/d4ab17fbc745705290e29e5251fa8e9fcbffe943/up-streamer/src/ustreamer.rs#L158-L159) and then the definitions of these functions [here](https://github.com/eclipse-uprotocol/up-streamer-rust/blob/d4ab17fbc745705290e29e5251fa8e9fcbffe943/up-streamer/src/ustreamer.rs#L31-L49).

up-l1/README.adoc
evshary: Done

up-l1/README.adoc
evshary: Thanks for both of your feedback. I removed it.

up-l1/README.adoc
evshary: Removed

up-l1/README.adoc
evshary: If I understand correctly, then we still need `{ANY, ANY}` in the uStreamer case, right?

PLeVasseur: We need `{ANY, ANY}` for being able to get all notification, request, and response. We don't have a need to get specifically only Requests with `{ANY, ANY}`.

evshary: oh, let me make it clear. The table here is the reorganized one from the table above. That means if we have `{ANY, ANY}`, the indicated message type will be notification, request and response.
image

Therefore in this case, we should get the request if the combination of resource ID is `{ANY, ANY}`. I guess the table might be a little hard to understand. Feel free to suggest a better form (I don't have a better idea...)

PLeVasseur: Ah okay, I understand what you mean now.

So yes -- source and sink UUri filters for resource IDs of {FFFF, FFFF} could mean any of Notification, Request, or Response.

PLeVasseur (approved):
LGTM

stevenhartley (changes requested):
@evshary I want to make sure we do not merge this. We have been talking about how we are sandwitching zenoh (a L2 protocol) into the L1 interface and then mapping back to L2 stuff under L1 implementation and that is why there is the need for the table. @gregmedd will be filing another issue to talk about what I mean but for now lets park this PR. Thank you!

stevenhartley:
@evshary, I'm closing the issue and this is why:

A lot of this confusion comes about if we want a transport to handle different uMessage types differently and the UMessage type is not passed to the registerListener() API. Given that we will not do this anymore (for now) with zenoh, I would like to put to rest this topic and when we decide to add this optimization, we can consider changing the UTransport APIs to add the UMessageType in lieu of these complicated mapping tables. 

If you disagree, please feel free to re-open the issue and we can discuss next week.

Issue #214, opened 2024-08-01
Error model for setting comstatus when error is a MUST not a SHOULD


stevenhartley:
in https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/error_model.adoc when the service provider fails to process a request, it **MUST** set `commstatus` with the `UCode` value and it **SHOULD** put the `UStatus` (to get the message and any other details) in the `UMessage` payload.

stevenhartley:
done

PR #215, opened 2024-08-01
Clarification for populating comstatus for RPC request failure

stevenhartley:
When the service fails to successfully process the request message, it must populate comstatus vs should.

#214


sophokles73 (approved):
Great clarification :+1: 

LGTM

stevenhartley:
I'll add the OFT tags later (just remembered about that). We probably should try and figure out how to associate the design tags with what client language library needs to implement vs transports, etc...

Issue #217, opened 2024-08-02
uE Instance ID of UUri when set to 0 is ambiguous


stevenhartley:
Some languages like up-cpp has interpreted the spec that it must always set the instance id to be 1 however the intention was that a value of 0 (don't care) is also used to indicate when there is only a single instance. the purpose of not forcing the "1" for a single instance was to avoid sending additional bytes for the URI.

The specifications need to be clarified to indicate that setting instance_id (upper 16 bits of ue_id) to 0 (don't care) is used to also indicate a single instance. It is covered slightly in the someip specs but needs some more TLC.

gregmedd:
@stevenhartley - I think this is a spec bug that needs to be fixed before we make code changes. We discussed elsewhere, but I've copied my recommendation into the bug filed against up-cpp: https://github.com/eclipse-uprotocol/up-cpp/issues/259#issuecomment-2276430356

In short, I recommend that we update the spec to define `0xFFFF` as the wildcard filter value for instance ID. It solves several problems, allows for the wildcard value to be reserved, improves the SOME/IP instance ID mapping, and removes ambiguity around the usage of instance ID `0` in several situations.

PLeVasseur:
FWIW, this would very seamlessly slot into how vsomeip [defines ANY_INSTANCE and ANY_SERVICE](https://github.com/COVESA/vsomeip/blob/5b4d2a64f11e23240728a7722dc6ac37a07b0752/interface/vsomeip/constants.hpp#L27-L28). :+1:

PR #222, opened 2024-08-02
URI ue instance id wildcard 0xFFFF in lieu of 0x0000

stevenhartley:
The following change adds some verbiage that states that it is valid to not set the instance_id that indicates that we do not care. This is often used when referring to services that do not have more than one instance.

This change is to help clarify the requirements after looking at the C++ UUri validators.

also minor fixes to the specifications w.r.t pattern matching for resource_id

#217

### DIFF
--- a/basics/uri.adoc
+++ b/basics/uri.adoc
@@ -277,10 +277,10 @@ impl UUri {
 A UUri can be used to define a _pattern_ that other UUris can then be matched against.
 For that purpose, a UUri
 
-* *MAY* have its `authority_name` set to the `*` (`U+002A`, Asterisk) character in order to match _any_ (including _no_) authority,
-* *MAY* have the _service ID_ part of its `ue_id` set to `0xFFFF` in order to match _any_ type,
-* *MAY* have the _service instance ID_ part of its `ue_id` set to `0x0000` in order to match _any_ instance,
-* *MAY* have its `ue_version_major` set to `0xFF` in order to match _any_ version,
+* *MAY* have its `authority_name` set to the `*` (`U+002A`, Asterisk) character in order to match _any_ (including _no_) authority.
+* *MAY* have the _service ID_ part of its `ue_id` set to `0xFFFF` in order to match _any_ service type.
+* *MAY* have the _service instance ID_ part of its `ue_id` set to `0xFFFF` in order to match _any_ service instance. `0x0000` is the default instance ID used when there is only a single instance of a service.
+* *MAY* have its `ue_version_major` set to `0xFF` in order to match _any_ version.
 * *MAY* have its `resource_id` set to `0xFFFF` in order to match _any_ resource.
 
 A _candidate_ UUri _matches_ a particular _pattern_ UUri if all of the candidate UUri's properties match the pattern UUri's corresponding properties according to the rules defined by the predicates below.
@@ -302,7 +302,7 @@ def: matches_entity_type(candidate : UUri) : Boolean =
   self.service_type() = candidate.service_type()
 
 def: matches_entity_instance(candidate : UUri) : Boolean =
-  self.service_instance() = 0x0000_0000
+  self.service_instance() = 0xFFFF_0000
   or
   self.service_instance() = candidate.service_instance()
 
@@ -312,9 +312,9 @@ def: matches_entity_version(candidate : UUri) : Boolean =
   self.ue_version_major = candidate.ue_version_major
 
 def: matches_resource(candidate : UUri) : Boolean =
-  self.instance_id = 0xFFFF
+  self.resource_id = 0xFFFF
   or
-  self.instance_id = candidate.instance_id
+  self.resource_id = candidate.resource_id
 
 def: matches(candidate : UUri) : Boolean =
   self.matches_authority(candidate)
@@ -341,7 +341,7 @@ The following pattern UUri
 ----
 UUri {
   authority_name: "192.168.1.100",
-  ue_id: 0x0000_FFFF                // any instance, any service
+  ue_id: 0xFFFF_FFFF                // any instance, any service
   ue_version_major: 0xFF,           // any
   resource_id: 0xFFFF               // any
 }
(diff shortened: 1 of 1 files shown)


sophokles73 (approved):
LGTM

basics/uri.adoc
```diff
@@ -102,7 +102,7 @@ uEntities *MAY* assume both the Service and Application roles.
 A UUri's `ue_id` property value determines the type and instance of the service being referred to:
 
 * The value's _least_ significant 16 bits contain the _service ID_, representing the service interface type.
-* The value's _most_ significant 16 bits contain the _service instance ID_.
+* The value's _most_ significant 16 bits contain the _service instance ID_ where a value of `0x0000` refers to _any_ (or don't care).
```
PLeVasseur, line 105: IIRC you chatted with Greg about this on Friday/Saturday and he had a different idea which you thumbs-upped.

PLeVasseur (changes requested):
Marking this so we remember to revisit based on Greg's proposal.

basics/uri.adoc
stevenhartley: Thanks Pete, I'll push an update with justification shortly

stevenhartley:
@gregmedd @PLeVasseur I've updated the _any_ service instance id to be 0xFFFF like we talked about so it is clearer. Now the  wildcard value and default instance id are two separate values and it is simpler to code/understand. Thanks everyone for your feedback!

basics/uri.adoc
```diff
@@ -277,10 +277,10 @@ impl UUri {
 A UUri can be used to define a _pattern_ that other UUris can then be matched against.
 For that purpose, a UUri
 
-* *MAY* have its `authority_name` set to the `*` (`U+002A`, Asterisk) character in order to match _any_ (including _no_) authority,
-* *MAY* have the _service ID_ part of its `ue_id` set to `0xFFFF` in order to match _any_ type,
-* *MAY* have the _service instance ID_ part of its `ue_id` set to `0x0000` in order to match _any_ instance,
-* *MAY* have its `ue_version_major` set to `0xFF` in order to match _any_ version,
+* *MAY* have its `authority_name` set to the `*` (`U+002A`, Asterisk) character in order to match _any_ (including _no_) authority.
+* *MAY* have the _service ID_ part of its `ue_id` set to `0xFFFF` in order to match _any_ service type.
+* *MAY* have the _service instance ID_ part of its `ue_id` set to `0xFFFF` in order to match _any_ service instance. `0x0000` is the default instance ID used when there is only a single instance of a service.
```
PLeVasseur, line 282: :+1:

basics/uri.adoc
```diff
@@ -341,7 +341,7 @@ The following pattern UUri
 ----
 UUri {
   authority_name: "192.168.1.100",
-  ue_id: 0x0000_FFFF                // any instance, any service
+  ue_id: 0xFFFF_FFFF                // any instance, any service
```
PLeVasseur, line 344: :+1:

basics/uri.adoc
```diff
@@ -371,7 +371,7 @@ The following pattern UUri
 ----
 UUri {
   authority_name: "*",       // any
-  ue_id: 0x0000_0000,        // any instance of service 0x0000
+  ue_id: 0xFFFF_0000,        // any instance of service 0x0000
```
PLeVasseur, line 374: :+1:

gregmedd (approved):
LGTM - thanks for making this change!

basics/uri.adoc
```diff
@@ -302,7 +302,7 @@ def: matches_entity_type(candidate : UUri) : Boolean =
   self.service_type() = candidate.service_type()
 
 def: matches_entity_instance(candidate : UUri) : Boolean =
-  self.service_instance() = 0x0000_0000
+  self.service_instance() = 0xFFFF_0000
```
PLeVasseur, line 305: :+1:

basics/uri.adoc
```diff
@@ -417,7 +417,7 @@ But not these:
 [example]
 ----
 //vcu.other.device/0/3/1     // non-local authority
-/1/3/3                       // wrong resource
+/0/3/3                       // wrong resource
```
PLeVasseur, line 420: Are there any reserved ue_ids? Is zero kosher to use? If so, that's fine, change just caught my eye.

stevenhartley: @PLeVasseur I'm formulating the registry topic shortly (for the lower 16 bits). For Instance id, we use 0 to mean "default/not set" and 0xFFFF to mean wildcard but any other value is permitted.

sophokles73:
@stevenhartley I am generally ok with the change to the entity instance wildcard pattern. However, introducing this semantic change _after_ I had approved the PR which was originally only about adjusting some wording seems a little off. I would appreciate if we could create a new PR for additional changes that are outside of the original scope of a PR in the future.

sophokles73:
@PLeVasseur

My understanding is that this change means that URIs used for configuring a streamer may need to be updated. Listening for messages from a particular entity type will now require addition of the instance ID wildcard:

```
up://origin/A1/1/D3A2
```
will need to be changed to
```
up://origin/FFFF00A1/1/D3A2
```

to receive the same messages, right?

PLeVasseur:
> @PLeVasseur
> 
> My understanding is that this change means that URIs used for configuring a streamer may need to be updated. Listening for messages from a particular entity type will now require addition of the instance ID wildcard:
> 
> ```
> up://origin/A1/1/D3A2
> ```
> 
> will need to be changed to
> 
> ```
> up://origin/FFFF00A1/1/D3A2
> ```
> 
> to receive the same messages, right?

We will have to update [these lines](https://github.com/eclipse-uprotocol/up-streamer-rust/blob/06af0faab4267d2583a60c8733c6431fae4a5318/up-streamer/src/ustreamer.rs#L31-L49).

Opened issue [here](https://github.com/eclipse-uprotocol/up-streamer-rust/issues/42)

basics/uri.adoc
sophokles73: @stevenhartley 
Given that the semantics have been changed, the OFT spec item's revision needs to also be changed: `[.specitem,oft-sid="dsn~uri-pattern-matching~2",oft-needs="impl,utest"]`

This is necessary in order for OpenFastTrace to be able to detect that existing implementations no longer cover this specification item.

Issue #218, opened 2024-08-02
uTransport::send() APIs MUST have requirements for validating source address in UMessages


stevenhartley:
When you send a message, the source address has to be the same as what was passed when you created the transport from the UriProvider, otherwise you can spoof other uEs and devices. 

Add requirements to up-l1/README.adoc to mandate the validation (saw that TCK is spoofing the source address which is a no-no

PLeVasseur:
A uStreamer _is_ kinda spoofing from one transport onto the next, so we need to allow some wiggle room here for certain cases like the uStreamer, right?

PR #224, opened 2024-08-06
Adding OFT metadata and uTransport address spoofing check.

stevenhartley:
The transport needs to verify that the caller (client) source address is the same as what is in the passed `UMessage` of the send() API to avoid any uE from spoofing another uE.  The only exception for this rule is if the uE is the streamer who is simply forwarding events from one transport to the next and in that case the streamers identity needs to be verified upon creation of the transport.

The PR also adds in missing OFT metadata.

#218

### DIFF
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -122,6 +122,7 @@ On the other hand, unsuccessful completion of this method does not necessarily m
 
 NOTE: The above strategy for retrying failed attempts to send a message results in https://www.cloudcomputingpatterns.org/at_least_once_delivery/[at-least-once delivery]. Recipient(s) of these messages should therefore be https://www.cloudcomputingpatterns.org/idempotent_processor/[Idempotent Processors].
 
+*Requirements for UTransport Implementations:*
 
 [.specitem,oft-sid="dsn~utransport-send-preserve-data~1",oft-needs="impl,utest"]
 --
@@ -186,6 +187,8 @@ receive(sourceFilter: UUri, sinkFilter: UUri [0..1]) : UMessage
 
 This method implements the _pull_ <<delivery-method, delivery method>> on top of the underlying communication protocol.
 
+*Requirements for UTransport Implementations:*
+
 [.specitem,oft-sid="req~utransport-receive-error-unimplemented~1",oft-needs="impl,utest"]
 --
 * *MUST* return `UCode.UNIMPLEMENTED` if the transport does not support the _pull_ <<delivery-method, delivery method>>
@@ -246,6 +249,8 @@ registerListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListene
 
 This API is used to implement the _push_ <<delivery-method, delivery method>> on top of the underlying communication protocol.
 After this method has completed successfully, the given listener will be invoked for each message that matches the given source and sink filter patterns according to the rules defined by the link:../basics/uri.adoc[UUri specification].
+
+*Requirements for UTransport Implementations:*
 
 [.specitem,oft-sid="req~utransport-registerlistener-error-unimplemented~1",oft-needs="impl,utest"]
 --
@@ -477,13 +482,5 @@ uProtocol defines bindings to the following communication protocols:
 * link:mqtt_5.adoc[*MQTT*]
 * link:http.adoc[*HTTP*]
 * link:someip.adoc[*SOME/IP*]
-
-
-If a communication protocol binding uses link:https://cloudevents.io/[CloudEvents]:
-
-[.specitem,oft-sid="dsn~utransport-cloudevents~1",oft-needs="impl,utest"]
---
-* *MUST* adhere to link:cloudevents.adoc[*UMessage mapping to CloudEvents*]
---
 
 
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -129,12 +129,12 @@ NOTE: The above strategy for retrying failed attempts to send a message results 
 * *MUST* preserve all of the message's meta data and payload during transmission
 --
 
-[.specitem,oft-sid="req~utransport-send-error-invalid-parameter~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-send-error-invalid-parameter~1",oft-needs="impl,utest"]
 --
 * *MUST* fail with a `UCode.INVALID_ARGUMENT` if the passed UMessage failed validation
 --
 
-[.specitem,oft-sid="req~utransport-send-error-permission-denied~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="req~utransport-send-error-permission-denied~1",oft-needs="dsn"]
 --
 * *MUST* fail with a `UCode.PERMISSION_DENIED` if a non-streamer client tries to send a message with `UAttributes.source` that is different than the source address associated with the transport, this is to avoid address spoofing.
 --
@@ -189,12 +189,12 @@ This method implements the _pull_ <<delivery-method, delivery method>> on top of
 
 *Requirements for UTransport Implementations:*
 
-[.specitem,oft-sid="req~utransport-receive-error-unimplemented~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-receive-error-unimplemented~1",oft-needs="impl,utest"]
 --
 * *MUST* return `UCode.UNIMPLEMENTED` if the transport does not support the _pull_ <<delivery-method, delivery method>>
 --
 
-[.specitem,oft-sid="req~utransport-receive-error-notfound~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-receive-error-notfound~1",oft-needs="impl,utest"]
 --
 * *MUST* return a `UCode.NOT_FOUND` if there are no matching messages available
 --
@@ -252,12 +252,12 @@ After this method has completed successfully, the given listener will be invoked
 
 *Requirements for UTransport Implementations:*
 
-[.specitem,oft-sid="req~utransport-registerlistener-error-unimplemented~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-registerlistener-error-unimplemented~1",oft-needs="impl,utest"]
 --
 * *MUST* fail with a `UCode.UNIMPLEMENTED` if the transport does not support the _push_ <<delivery-method, delivery method>>. In that case, the <<unregister-listener, unregisterListener>> method *MUST* also fail accordingly.
 --
 
-[.specitem,oft-sid="req~utransport-registerlistener-error-resource-exhausted~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-registerlistener-error-resource-exhausted~1",oft-needs="impl,utest"]
 --
 * *MUST* fail with a `UCode.RESOURCE_EXHAUSTED`, if the maximum number of listeners is reached
 --
@@ -372,17 +372,17 @@ unregisterListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListe
 | The listener to be unregistered.
 |===
 
-[.specitem,oft-sid="req~utransport-unregisterlistener-error-unimplemented~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-unregisterlistener-error-unimplemented~1",oft-needs="impl,utest"]
 --
 * *MUST* fail with a `UCode.UNIMPLEMENTED` if the transport does not support the _push_ <<delivery-method>>. In that case, the <<register-listener>> method *MUST* also fail accordingly.
 --
 
-[.specitem,oft-sid="req~utransport-unregisterlistener-error-notfound~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-unregisterlistener-error-notfound~1",oft-needs="impl,utest"]
 --
 * *MUST* fail with a `UCode.NOT_FOUND`, if no such listener had been registered before
 --
 
-[.specitem,oft-sid="req~utransport-unregisterlistener-error-invalid-parameter~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-unregisterlistener-error-invalid-parameter~1",oft-needs="impl,utest"]
 --
 * *MUST* fail with a `UCode.INVALID_ARGUMENT` if the passed UUri failed validation
 --
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -123,7 +123,7 @@ On the other hand, unsuccessful completion of this method does not necessarily m
 NOTE: The above strategy for retrying failed attempts to send a message results in https://www.cloudcomputingpatterns.org/at_least_once_delivery/[at-least-once delivery]. Recipient(s) of these messages should therefore be https://www.cloudcomputingpatterns.org/idempotent_processor/[Idempotent Processors].
 
 
-[.specitem,oft-sid="des~utransport-send-preserve-data~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-send-preserve-data~1",oft-needs="impl,utest"]
 --
 * *MUST* preserve all of the message's meta data and payload during transmission
 --
@@ -262,17 +262,17 @@ After this method has completed successfully, the given listener will be invoked
 * *MUST* fail with a `UCode.INVALID_ARGUMENT` if the passed UUri failed validation
 --
 
-[.specitem,oft-sid="req~utransport-registerlistener-number-of-listeners~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-registerlistener-number-of-listeners~1",oft-needs="impl,utest"]
 --
 * *MUST* support registering more than one listener for any given address patterns
 --
   
-[.specitem,oft-sid="req~utransport-registerlistener-listener-reuse~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="dsn~utransport-registerlistener-listener-reuse~1",oft-needs="impl,utest"]
 --
 * *MUST* support registering the same listener for multiple address patterns
 --
 
-[.specitem,oft-sid="dsn~utransport-registerlistener-max-listeners~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="req~utransport-registerlistener-max-listeners~1",oft-needs="uman"]
 --
 * *MUST* document the maximum supported number of listeners per address pattern.
 --
@@ -415,7 +415,7 @@ A `UTransport` implementation can use the `LocalUriProvider` to determine the uE
 
 A uEntity invokes this method to get its own authority.
 
-[.specitem,oft-sid="req~utransport-localuriprovider-getauthority~1",oft-needs="impl,utest"]
+[.specitem,oft-sid="req~utransport-localuriprovider-getauthority~1",oft-needs="dsn"]
 --
 Implementations *MAY* use any appropriate mechanism to determine the local authority during runtime, e.g. by means of a configuration file, environment variables or a central registry.
 --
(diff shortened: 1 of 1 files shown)


up-l1/README.adoc
```diff
@@ -122,9 +122,21 @@ On the other hand, unsuccessful completion of this method does not necessarily m
...
+--
+
+[.specitem,oft-sid="req~utransport-send-error-invalid-parameter~1",oft-needs="impl,utest"]
+--
+* *MUST* fail with a `UCode.INVALID_ARGUMENT` if the passed UMessage failed validation
+--
+
+[.specitem,oft-sid="req~utransport-send-error-permission-denied~1",oft-needs="impl,utest"]
+--
+* *MUST* fail with a `UCode.PERMISSION_DENIED` if a non-streamer client tries to send a message with `UAttributes.source` that is different than the source address associated with the transport, this is to avoid address spoofing.
```
PLeVasseur, line 138:
```suggestion
* *MUST* fail with a `UCode.PERMISSION_DENIED` if a non-streamer client tries to send a message with `UAttributes.source` that is different than the source address associated with the transport. Done to avoid address spoofing.
```

How do we propose to activate / deactivate this logic within each transport? If all we do is have an alternative constructor uStreamer constructor, it's possible for someone to simply use that constructor and spoof.

Is there some kind of authentication or verification that should be done? I'm not well-versed in how to allow this for the uStreamer, but prevent for all other users of a transport on a device.

PLeVasseur (commented):
IMHO it'd be good to outline _possible_ methods of preventing transport spoofing if not a uStreamer user of a transport. WDYT?

up-l1/README.adoc
```diff
@@ -122,9 +122,21 @@ On the other hand, unsuccessful completion of this method does not necessarily m
 
 NOTE: The above strategy for retrying failed attempts to send a message results in https://www.cloudcomputingpatterns.org/at_least_once_delivery/[at-least-once delivery]. Recipient(s) of these messages should therefore be https://www.cloudcomputingpatterns.org/idempotent_processor/[Idempotent Processors].
 
-UTransport implementations
```
sophokles73, line 125: FMPOV this line needs to remain in the document. Otherwise, the following list makes no sense ...

up-l1/README.adoc
sophokles73: FMPOV this taps into the domain of authentication/authorization. A uEntity and/or uStreamer will need to be run in the context of an authenticated identity with some authorities granted by the system owner. This might be as simple as a JWT that has been issued by an identity provider. A uStreamer in this context would then have a specific authority that is being checked by the transport during runtime and which allows the streamer to send messages _on behalf of_ other uEntities.

up-l1/README.adoc
```diff
@@ -122,9 +122,21 @@ On the other hand, unsuccessful completion of this method does not necessarily m
...
 
+[.specitem,oft-sid="des~utransport-send-preserve-data~1",oft-needs="impl,utest"]
```
sophokles73, line 126: the _design_ spec item type uses the `dsn` abbreviation

```suggestion
[.specitem,oft-sid="dsn~utransport-send-preserve-data~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -122,9 +122,21 @@ On the other hand, unsuccessful completion of this method does not necessarily m
...
+--
 * *MUST* preserve all of the message's meta data and payload during transmission
+--
+
+[.specitem,oft-sid="req~utransport-send-error-invalid-parameter~1",oft-needs="impl,utest"]
```
sophokles73, line 131: I would also place this in the _design_ realm ...

```suggestion
[.specitem,oft-sid="dsn~utransport-send-error-invalid-parameter~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -122,9 +122,21 @@ On the other hand, unsuccessful completion of this method does not necessarily m
...
+--
+* *MUST* fail with a `UCode.INVALID_ARGUMENT` if the passed UMessage failed validation
+--
+
+[.specitem,oft-sid="req~utransport-send-error-permission-denied~1",oft-needs="impl,utest"]
```
sophokles73, line 136: FMPOV it is not obvious, how this should be implemented. Thus, this requirement will probably need some implementation specific design first.

```suggestion
[.specitem,oft-sid="req~utransport-send-error-permission-denied~1",oft-needs="dsn"]
```

up-l1/README.adoc
```diff
@@ -174,10 +186,15 @@ receive(sourceFilter: UUri, sinkFilter: UUri [0..1]) : UMessage
 
 This method implements the _pull_ <<delivery-method, delivery method>> on top of the underlying communication protocol.
 
-UTransport implementations
+[.specitem,oft-sid="req~utransport-receive-error-unimplemented~1",oft-needs="impl,utest"]
```
sophokles73, line 189:
```suggestion
[.specitem,oft-sid="dsn~utransport-receive-error-unimplemented~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -174,10 +186,15 @@ receive(sourceFilter: UUri, sinkFilter: UUri [0..1]) : UMessage
...
+--
+* *MUST* return `UCode.UNIMPLEMENTED` if the transport does not support the _pull_ <<delivery-method, delivery method>>
+--
 
-* *MUST* fail with a `UCode.UNIMPLEMENTED` if the transport does not support the _pull_ <<delivery-method, delivery method>>
-* *MUST* fail with a `UCode.NOT_FOUND` if there are no matching messages available
+[.specitem,oft-sid="req~utransport-receive-error-notfound~1",oft-needs="impl,utest"]
```
sophokles73, line 194:
```suggestion
[.specitem,oft-sid="dsn~utransport-receive-error-notfound~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -230,14 +247,41 @@ registerListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListene
 This API is used to implement the _push_ <<delivery-method, delivery method>> on top of the underlying communication protocol.
 After this method has completed successfully, the given listener will be invoked for each message that matches the given source and sink filter patterns according to the rules defined by the link:../basics/uri.adoc[UUri specification].
 
-UTransport implementations
-
+[.specitem,oft-sid="req~utransport-registerlistener-error-unimplemented~1",oft-needs="impl,utest"]
```
sophokles73, line 250:
```suggestion
[.specitem,oft-sid="dsn~utransport-registerlistener-error-unimplemented~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -230,14 +247,41 @@ registerListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListene
...
+--
 * *MUST* fail with a `UCode.UNIMPLEMENTED` if the transport does not support the _push_ <<delivery-method, delivery method>>. In that case, the <<unregister-listener, unregisterListener>> method *MUST* also fail accordingly.
+--
+
+[.specitem,oft-sid="req~utransport-registerlistener-error-resource-exhausted~1",oft-needs="impl,utest"]
```
sophokles73, line 255:
```suggestion
[.specitem,oft-sid="dsn~utransport-registerlistener-error-resource-exhausted~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -230,14 +247,41 @@ registerListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListene
...
+--
 * *MUST* fail with a `UCode.RESOURCE_EXHAUSTED`, if the maximum number of listeners is reached
+--
+
+[.specitem,oft-sid="req~utransport-registerlistener-error-invalid-parameter~1",oft-needs="impl,utest"]
+--
+* *MUST* fail with a `UCode.INVALID_ARGUMENT` if the passed UUri failed validation
+--
```
sophokles73, line 263: FMPOV this is already covered by validation of the message

up-l1/README.adoc
```diff
@@ -230,14 +247,41 @@ registerListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListene
...
+
+[.specitem,oft-sid="req~utransport-registerlistener-number-of-listeners~1",oft-needs="impl,utest"]
```
sophokles73, line 265:
```suggestion
[.specitem,oft-sid="dsn~utransport-registerlistener-number-of-listeners~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -230,14 +247,41 @@ registerListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListene
...
+--
 * *MUST* support registering more than one listener for any given address patterns
+--
+  
+[.specitem,oft-sid="req~utransport-registerlistener-listener-reuse~1",oft-needs="impl,utest"]
```
sophokles73, line 270:
```suggestion
[.specitem,oft-sid="dsn~utransport-registerlistener-listener-reuse~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -323,10 +367,20 @@ unregisterListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListe
 | The listener to be unregistered.
 |===
 
-UTransport implementations
-
+[.specitem,oft-sid="req~utransport-unregisterlistener-error-unimplemented~1",oft-needs="impl,utest"]
```
sophokles73, line 370:
```suggestion
[.specitem,oft-sid="dsn~utransport-unregisterlistener-error-unimplemented~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -323,10 +367,20 @@ unregisterListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListe
...
+--
 * *MUST* fail with a `UCode.UNIMPLEMENTED` if the transport does not support the _push_ <<delivery-method>>. In that case, the <<register-listener>> method *MUST* also fail accordingly.
+--
+
+[.specitem,oft-sid="req~utransport-unregisterlistener-error-notfound~1",oft-needs="impl,utest"]
```
sophokles73, line 375:
```suggestion
[.specitem,oft-sid="dsn~utransport-unregisterlistener-error-notfound~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -323,10 +367,20 @@ unregisterListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListe
...
+--
 * *MUST* fail with a `UCode.NOT_FOUND`, if no such listener had been registered before
+--
+
+[.specitem,oft-sid="req~utransport-unregisterlistener-error-invalid-parameter~1",oft-needs="impl,utest"]
```
sophokles73, line 380:
```suggestion
[.specitem,oft-sid="dsn~utransport-unregisterlistener-error-invalid-parameter~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -361,31 +415,52 @@ A `UTransport` implementation can use the `LocalUriProvider` to determine the uE
 
 A uEntity invokes this method to get its own authority.
 
+[.specitem,oft-sid="req~utransport-localuriprovider-getauthority~1",oft-needs="impl,utest"]
```
sophokles73, line 418: this might require some transport specific design

```suggestion
[.specitem,oft-sid="req~utransport-localuriprovider-getauthority~1",oft-needs="dsn"]
```

up-l1/README.adoc
```diff
@@ -361,31 +415,52 @@ A `UTransport` implementation can use the `LocalUriProvider` to determine the uE
...
+--
 Implementations *MAY* use any appropriate mechanism to determine the local authority during runtime, e.g. by means of a configuration file, environment variables or a central registry.
+--
 
 === GetSource
 
 A uEntity invokes this method to get the address that it expects incoming Notification or RPC Response messages to be sent to.
 
+[.specitem,oft-sid="req~utransport-localuriprovider-getsource-uri-segments~1",oft-needs="impl,utest"]
```
sophokles73, line 427:
```suggestion
[.specitem,oft-sid="dsn~utransport-localuriprovider-getsource-uri-segments~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -361,31 +415,52 @@ A `UTransport` implementation can use the `LocalUriProvider` to determine the uE
...
+--
 The address returned by an implementation *MUST* consist of the uEntity's (fixed) _authority_, _identifier_ and _major version_ and _resource ID_ `0x0000`.
+--
 
+[.specitem,oft-sid="req~utransport-localuriprovider-getsource-runtime~1",oft-needs="impl,utest"]
```
sophokles73, line 432:
```suggestion
[.specitem,oft-sid="req~utransport-localuriprovider-getsource-runtime~1",oft-needs="dsn"]
```

up-l1/README.adoc
```diff
@@ -361,31 +415,52 @@ A `UTransport` implementation can use the `LocalUriProvider` to determine the uE
...
+--
 Implementations *MAY* use any appropriate mechanism to determine these values during runtime, e.g. by means of a configuration file, environment variables or a central registry.
+--
 
 === GetResource
 
 A uEntity invokes this method to get a resource specific address to publish messages to or that it expects incoming RPC Request messages to be sent to.
 
+[.specitem,oft-sid="req~utransport-localuriprovider-getresource~1",oft-needs="impl,utest"]
```
sophokles73, line 441:
```suggestion
[.specitem,oft-sid="dsn~utransport-localuriprovider-getresource~1",oft-needs="impl,utest"]
```

up-l1/README.adoc
```diff
@@ -361,31 +415,52 @@ A `UTransport` implementation can use the `LocalUriProvider` to determine the uE
...
+--
 The address returned by an implementation *MUST* consist of the uEntity's (fixed) _authority_, _identifier_ and _major version_ and the passed in _resource ID_.
-
+--
+  
+[.specitem,oft-sid="req~utransport-localuriprovider-getresource-runtime~1",oft-needs="impl,utest"]
```
sophokles73, line 446:
```suggestion
[.specitem,oft-sid="req~utransport-localuriprovider-getresource-runtime~1",oft-needs="dsn"]
```

up-l1/README.adoc
```diff
@@ -361,31 +415,52 @@ A `UTransport` implementation can use the `LocalUriProvider` to determine the uE
...
 == Message Delivery
 
-Transport API implementations
-
+[.specitem,oft-sid="dsn~utransport-delivery-methods~1",oft-needs="impl,utest"]
+--
 * *MUST* support at least one of _push_ or _pull_ delivery methods and *MAY* support both
+--
+  
+[.specitem,oft-sid="req~utransport-delivery-methods-docs~1",oft-needs="impl"]
```
sophokles73, line 459:
```suggestion
[.specitem,oft-sid="req~utransport-delivery-methods-docs~1",oft-needs="uman"]
```

up-l1/README.adoc
```diff
@@ -230,14 +247,41 @@ registerListener(sourceFilter: UUri, sinkFilter: UUri [0..1], listener: UListene
...
+--
 * *MUST* support registering more than one listener for any given address patterns
+--
+  
+[.specitem,oft-sid="req~utransport-registerlistener-listener-reuse~1",oft-needs="impl,utest"]
+--
 * *MUST* support registering the same listener for multiple address patterns
+--
+
+[.specitem,oft-sid="dsn~utransport-registerlistener-max-listeners~1",oft-needs="impl,utest"]
```
sophokles73, line 275:
```suggestion
[.specitem,oft-sid="req~utransport-registerlistener-max-listeners~1",oft-needs="uman"]
```

up-l1/README.adoc
```diff
@@ -403,11 +478,12 @@ uProtocol defines bindings to the following communication protocols:
...
 * link:someip.adoc[*SOME/IP*]
 
-Each uProtocol Client *MUST* employ exactly one of these bindings for implementing the Transport Layer API.
-
-Additional bindings *MAY* be defined in future versions of uProtocol.
 
-A binding *MAY* employ link:https://cloudevents.io/[CloudEvents] as a means to map uProtocol messages to the communication protocol's PDU. In order to provide for consistency across implementations, such bindings *MUST* adhere to link:cloudevents.adoc[*UMessage mapping to CloudEvents*]
+If a communication protocol binding uses link:https://cloudevents.io/[CloudEvents]:
 
+[.specitem,oft-sid="dsn~utransport-cloudevents~1",oft-needs="impl,utest"]
```
sophokles73, line 484: a transport that does not use CloudEvents will not be able to cover this spec item

up-l1/README.adoc
stevenhartley: I'll add the ":" then to the end

up-l1/README.adoc
stevenhartley: Arn't error values returned user requirements for the API and not design/architecture considerations? that is why I made some `req` and others `dsn`

up-l1/README.adoc
stevenhartley: We define what the UMessage validators are suppose to do so why do we need more implementation specific design first?

up-l1/README.adoc
stevenhartley: Agree with you @sophokles73 but trying to see how I can word this into the transport spec as we need to ensure that address spoofing cannot happen by clients. I wonder if we bake this into the LocalUriAuthority definitions perhaps?

up-l1/README.adoc
stevenhartley: Also API return codes are user requirements no?

up-l1/README.adoc
stevenhartley: same as other comment

up-l1/README.adoc
stevenhartley: registerListener() API does not get passed a UMessage so we still need to validate the passed URIs used in the filters.

up-l1/README.adoc
stevenhartley: Yes and that is why there is a "If" at the beginning of line 482

up-l1/README.adoc
sophokles73: IMHO something like
> In case of an error, the API must provide specific information to client code regarding the reason of the error.

would be a (user) requirement. If we already explicitly define the error codes and under which circumstances we want to return them, then IMHO we have already made a design decision as to how the API looks like. Which is a reasonable thing to do here because we want to make sure that every language library implements this the same way.

up-l1/README.adoc
sophokles73: I don't think I understand ...

up-l1/README.adoc
sophokles73: This is about the spec item for the _permission denied_ case, not the _invalid message_ ...

up-l1/README.adoc
sophokles73: yes, my bad, I thought it was in the _send_ section ...

up-l1/README.adoc
stevenhartley: Given that we have moved away from CloudEvents, I would prefer removing the requirements entirely as non of the transports will use it in version >1.5.

up-l1/README.adoc
PLeVasseur: So where did we land on this? Leave open-ended on how to achieve the no-spoofing requirement, give some suggestions, or mandate something?

stevenhartley: @PLeVasseur If it is OK with you, I would like to add that in a subsequent PR unless you think it needs to be addressed in this PR now.

up-l1/README.adoc
stevenhartley: I added it back

up-l1/README.adoc
PLeVasseur: If it's a follow-up, then I'm good.

PLeVasseur (approved):
LGTM

Reminder to self on needing a follow-up to add recommendations on preventing uTransport spoofing (aside from uStreamer).

up-l1/README.adoc
stevenhartley: @PLeVasseur I created #228 to track this open item

Issue #212, opened 2024-08-01
Specify re-connect handling for different protocols 


jjj-vtm:
Hi, 

currently there is no specification on the reconnect handling for mqtt or zenoh protocol. I think it would make sense to include it in the specifications since loosing the connection from a vehicle to a broker is quite a common case. I would suggest to use a backoff retry strategy with an increasing backoff time. For example 500ms, 1s, 2s, 4s, 8s and then stay at 8s. 

For the different protocols the following things must be taken into account for the implementation. 
MQTT: If the connection was established with clean sessions = true than after successful reconnection the listener topics have to be resubscribed. If there was a previous session indicated via sessionPresent flag, then no re-subscription should occur.
Zenoh: To my knowledge there is no session option so re-subscription must occur.

stevenhartley:
Hi @jjj-vtm can you please push a pull request with the above proposal to mqtt_5.adoc and zenoh.adoc. @evshary tagging you for input on the subject.

evshary:
In my understanding, Zenoh should take care of the re-subscription internally. Zenoh also provides the reconnection mechanism in the configuration.
https://github.com/eclipse-zenoh/zenoh/blob/5d09cf7cb7af1c01611fa7eaeed3b9592baa9374/DEFAULT_CONFIG.json5#L46

Loop @Mallets here in case I'm wrong.

jjj-vtm:
Spec adaption was merged

PR #226, opened 2024-08-07
Specify re-connect handling for different protocols #212

jjj-vtm:


up-l1/mqtt_5.adoc
```diff
@@ -235,5 +235,10 @@ When MQTT5 (broker) is used for D2D communication, the topic shall consist of on
 ===== Sending a Message
 `UTransport::send(UMessage)` translates to the MQTT publish to topic `{UMessage.source.authority_name}/{UMessage.sink.authority_name}`
 
+== MQTT5 Connection handling
 
+=== Establishing a connection
+Establishing a connection to an MQTT broker *SHOULD* be done with `cleanSession = true` and `sessionExpiry = 0`. With those settings the broker will not buffer any messages while a the client is offline and will loose any subscriptions on a client disconnect.
```
stevenhartley, line 241: Please add the OFT tags like above for your requirements.

stevenhartley, line 241: We actually might want to buffer messages in the broker for the scenario so vehicles can receive messages when they wakeup so I'm not sure if we want to make this a SHOULD. I think the transport needs to have a configuration option for this setting.

up-l1/mqtt_5.adoc
```diff
@@ -235,5 +235,10 @@ When MQTT5 (broker) is used for D2D communication, the topic shall consist of on
 ===== Sending a Message
 `UTransport::send(UMessage)` translates to the MQTT publish to topic `{UMessage.source.authority_name}/{UMessage.sink.authority_name}`
 
+== MQTT5 Connection handling
 
+=== Establishing a connection
+Establishing a connection to an MQTT broker *SHOULD* be done with `cleanSession = true` and `sessionExpiry = 0`. With those settings the broker will not buffer any messages while a the client is offline and will loose any subscriptions on a client disconnect. 
 
+=== Reconnection
+If a connection is lost the UTransport *MUST* try to reconnect to the MQTT broker. The client *SHOULD* implement a backoff retry strategy to avoid overloading the broker. If the client connected with `cleanSession = true` then after a successful reconnect the UTransport *MUST* re-register all the previosly registred UListener.
```
stevenhartley, line 244: Add OFT traces and make each requirement a new line.  Please define the backoff strategy or the requirement is of type `req` and it needs `dsn` like this:

```
[.specitem,oft-sid="req~up-transport-mqtt5-reconnect-backoff~1",oft-needs="dsn"]
--
*  *SHOULD* implement a backoff retry strategy to avoid overloading the broker
--

```

stevenhartley (changes requested):

up-l1/mqtt_5.adoc
jjj-vtm: Ok, I will rewrite it in a form that we want the transport to have the options. IMHO going with "stateless" ie. cleanSession = true and sessionExpiry=0 is the best way for scaling because it opens up some possible deployment topologies but if you have state ie. sessions you are basically tied to a clustered MQTT deployment.

jjj-vtm: Done via: 

[](https://github.com/jjj-vtm/up-spec/blob/d984eb005d26b15437f95337ada3c953c7687c09/up-l1/mqtt_5.adoc?plain=1#L240C1-L244C1)

up-l1/mqtt_5.adoc
```diff
@@ -235,5 +235,43 @@ When MQTT5 (broker) is used for D2D communication, the topic shall consist of on
...
+Using a UTransport from the cloud side to connect to an MQTT broker we recommend to use `cleanSession=true` and `sessionExpiry=0` because we assume that there will always be at least one UTransport connected from the cloud which can handle messages so there is no need for buffering. 
+
+Establishing an mqtt5 UTransport from a device there is no general recommendation. Using `cleanSession=true` and `sessionExpiry=0` might be best for scaling since it reduces the load on the MQTT broker. Doing this one needs to take special care for offline devices. Using `cleanSession=true` and `sessionExpiry=t` might stress the broker but the broker helps when dealing with offline devices.
+
+--
+
+=== Reconnection
+[.specitem,oft-sid="req~up-transport-mqtt5-reconnection~1",oft-needs="impl,utest"]
+--
+If a connection is lost the UTransport *MUST* try to reconnect to the MQTT broker.  If on successfull reconnect the broker sends the `sessionPresent=false` flag then the UTransport *MUST* re-register all the previosly registred UListener.
```
sophokles73, line 264:
```suggestion
If a connection is lost, the UTransport *MUST* try to reconnect to the MQTT broker.  If on successfull reconnect the broker sends the `sessionPresent=false` flag, the UTransport *MUST* re-subscribe to all previously subscribed topics. 
```

up-l1/mqtt_5.adoc
```diff
@@ -235,5 +235,43 @@ When MQTT5 (broker) is used for D2D communication, the topic shall consist of on
...
+
+=== Reconnection
+[.specitem,oft-sid="req~up-transport-mqtt5-reconnection~1",oft-needs="impl,utest"]
+--
+If a connection is lost the UTransport *MUST* try to reconnect to the MQTT broker.  If on successfull reconnect the broker sends the `sessionPresent=false` flag then the UTransport *MUST* re-register all the previosly registred UListener. 
+--
+
+[.specitem,oft-sid="req~up-transport-mqtt5-reconnection-backoff~1",oft-needs="impl,utest"]
+--
+The client *MUST* implement a backoff strategy in order not to overload the MQTT broker. It *MUST* be possible to inject this strategy into the UTransport. The UTransport *MUST* implement the default strategy where the n+1-th reconnect is done after the n-th after a pause given by the following table:
```
sophokles73, line 269: > It *MUST* be possible to inject this strategy into the UTransport.

What is that supposed to mean? Should it be possible to enable/disable the backoff, or parameterize it? If so, how? Unless we are more specific in this point I would propose to simply require the transport to implement the standard exponential backoff algorithm and use always use it ...

up-l1/mqtt_5.adoc
jjj-vtm: Addressed here: https://github.com/jjj-vtm/up-spec/blob/d984eb005d26b15437f95337ada3c953c7687c09/up-l1/mqtt_5.adoc?plain=1#L245C1-L258C1

up-l1/mqtt_5.adoc
jjj-vtm: Idea was to allow custom backoff strategies. I agree that just specifying the standard exponential backoff is good enough.

up-l1/mqtt_5.adoc
jjj-vtm: OFT added
https://github.com/jjj-vtm/up-spec/blob/d55ebba8cecf79e74ba916f265b3b3ee0ff0d1c3/up-l1/mqtt_5.adoc?plain=1#L261C1-L268C3

up-l1/mqtt_5.adoc
jjj-vtm: https://github.com/eclipse-uprotocol/up-spec/commit/08f440905b2457fbcecb7a72f805131b1d7b9780

sophokles73 (approved):
LGTM

stevenhartley (approved):
LGTM

Issue #220, opened 2024-08-02
Cleanup SubscriberInfo in usubscription.proto that is not used in many cases


stevenhartley:
The `SubscriberInfo` message contained the URI of the subscriber and `repeated google.protobuf.Any` of details  that we never used. 
The SubscriberInfo is also passed to various methods that never uses the information and/or is not required. 

We will remove the unused attribute (details) and mark the SubscriberInfo as deprecated (not required) for those APIs that it is not used/needed.

PR #227, opened 2024-08-07
Remove unused SubscriberInfo from some USubscription messages

stevenhartley:
The SubscriberInfo was a wrapper for the subscribers URI however it passes details that was never used (and complicates the implementations) and is passed in some scenarios that is redundant (we know who is calling the APIs). The SubscriberInfo is also passed to registerForNotifications however subscribers don't call this API so it is confusing for developers.

#220


sophokles73 (approved):
LGTM

but I guess @AnotherDaniel has more insight into the needs of the API ...

mishap4 (approved):

PR #230, opened 2024-08-13
Change OFT spec item revision

sophokles73:
The revision number of the specification item for the Pattern Matching
requirements has been increased to reflect the semantic change to
the service instance ID wildcard value.

This addresses #217


PLeVasseur (approved):
LGTM

PR #234, opened 2024-08-16
Use Zenoh for L1 Transport Only

stevenhartley:
Zenoh is a technology that supports uProtocol L1 and L2 functionality however we have sandwiched it into the L1 uTransport interface and then expected it to be optimized and do some L2 stuff below L1 (see the confusion). The following change will state that implementations will ONLY use the pub/sub of zenoh such that it can simply send() and receive UMessages and we leave the L2 functionality (ex. RPC corelation) up to the uProtocol language libraries.

### DIFF
--- a/up-l1/zenoh.adoc
+++ b/up-l1/zenoh.adoc
@@ -65,18 +65,8 @@ Now the version is always 0x01.
 
 ==== Message Type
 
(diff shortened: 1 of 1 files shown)


gregmedd (approved):
Based on discussion in #229 and on slack, this looks correct to me

PR #235, opened 2024-08-17
Enable generic service generation in C++

gregmedd:
See https://protobuf.dev/reference/cpp/cpp-generated/#service

### DIFF
--- a/up-core-api/uprotocol/core/udiscovery/v3/udiscovery.proto
+++ b/up-core-api/uprotocol/core/udiscovery/v3/udiscovery.proto
@@ -22,6 +22,9 @@ option java_package = "org.eclipse.uprotocol.core.udiscovery.v3";
 option java_outer_classname = "UDiscoveryProto";
 option java_multiple_files = true;
 
+// Enables generation of generic service attributes in C++
+option cc_generic_services = true;
+
 // Platform USE Discovery Service Interface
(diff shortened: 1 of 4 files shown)


gregmedd:
@stevenhartley - should this target main, or do you want me to create a release / bugfix branch off of the alpha.3 tag?

stevenhartley:
> @stevenhartley - should this target main, or do you want me to create a release / bugfix branch off of the alpha.3 tag?

Not sure what you mean by this issue, all we need is in the generated protoc code.

gregmedd:
> Not sure what you mean by this issue, all we need is in the generated protoc code.

The generated C++ code from protoc does not include any of the service information by default. It must be enabled by adding the `cc_generic_services = true" setting to .proto files containing service descriptions.

The only alternative offered for C++ service code generation in the protobuf documentation is to implement a custom code generator plugin.

Since a change to the .proto files is required, and the uProtocol libraries are using 1.6.0-alpha3 right now, I wanted to know if we should branch off of the alpha3 tag to make this a small bug fix (maybe tagged 1.6.0-alpha3.1) or if this should be merged to main.

If this is merged to main, up-cpp would have to consume up-spec main until the next tagged release.

stevenhartley (approved):
LGTM

PR #243, opened 2024-10-31
Update the Zenoh spec to match Zenoh 1.0

evshary:
There are some updates in Zenoh spec:

* Fix Zenoh version to 1.0.0
* Fix some out-dated terms, like `up-client` => `up-transport`
* Update the user attachment mechanism


sophokles73 (approved):
LGTM

PR #244, opened 2024-10-31
Adding Topology Diagram

stevenhartley:
The following commit adds the topology drawio diagram found in the roadmap document.


PLeVasseur (approved):
:+1:

sophokles73 (commented):
FMPOV it would be even better to add this as an SVG that has the diagram's xml embedded. This can be created from with in draw.io by means of exporting the diagram as SVG and then ticking the _embed copy of diagram_ checkbox.

stevenhartley:
@sophokles73 done.

sophokles73 (approved):
LGTM

Issue #237, opened 2024-08-21
Payload format for SOME/IP transport should not be UNSPECIFIED


sophokles73:
SOME/IP has no concept of payload encoding and/or content types. The [SOME/IP Transport spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/someip.adoc#221-common-umessage-attributes) therefore advises to leave a UMessage's `payload_format` as `UNDEFINED` when mapping a SOME/IP message to a UMessage, assuming that the _receiver will know_ how to interpret the data.

Using the `UNSPECIFIED` payload format for this purpose, however, seems inappropriate because [uattributes.proto](https://github.com/eclipse-uprotocol/up-spec/blob/da5ca97d3a7541d2fcd52ed010bc3bcca92e46cb/up-core-api/uprotocol/v1/uattributes.proto#L128) defines `UPAYLOAD_FORMAT_PROTOBUF_WRAPPED_IN_ANY` to be assumed as the payload format if `UNSPECIFIED` (not set). This means that any uE that is not aware of the fact that the message originates from a SOME/IP service, will try to process the message payload as a protobuf wrapped in an `Any`.

PLeVasseur:
Hey Kai. So is the proposal to leave the `payload_format` empty instead to truly say "we have no idea"?

Or to modify the meaning of `UPAYLOAD_FORMAT_UNSPECIFIED` to truly mean "we have no idea"?

IMHO having `UPAYLOAD_FORMAT_UNSPECIFIED` mean "yeah... maybe it's `UPAYLOAD_FORMAT_PROTOBUF_WRAPPED_IN_ANY` actually?" seems a bit unintuitive. So I'd vote for the latter idea.

What are your thoughts?

sophokles73:
The reason for defining `UPAYLOAD_FORMAT_UNSPECIFIED` as meaning `UPAYLOAD_FORMAT_PROTOBUF_WRAPPED_IN_ANY` was that we wanted to be able to simply omit the attribute in the majority of messages, because the most prevalent payload format is expected to be `UPAYLOAD_FORMAT_PROTOBUF_WRAPPED_IN_ANY`. FMPOV this still makes sense.

[...]

PR #247, opened 2024-11-08
Remove ambiguity for UNSPECIFIED UPayloadFormat

stevenhartley:
Previously (version 1.3.6 and earlier of the specification) we did not have an enum for the payload format (it was a string) and sending the string was costly to the cloud even though we always knew it was WRAPPED_IN_ANY. To avoid sending the payload format we omitted the attribute and then  "assumed" that if it was not present it would be WRAPPED_IN_ANY.  Now that we have an enum defined, it is better to remove the ambiguity of this attribute and make it explicit, this means that if someone sets it to UNSPECIFIED it really means unspecified (we don't know or are not setting it to anything). Additionally, for SOME/IP there is no way to pass this attribute (UPayloadFormat) so we have to set UPayloadFormat as UNSPECIFIED and then leave it up to the client to know the format from the source.

#237

### DIFF
--- a/basics/upayloadformat.adoc
+++ b/basics/upayloadformat.adoc
@@ -28,7 +28,7 @@ The following table provides an overview of the payload formats supported by uPr
 
 | `UPAYLOAD_FORMAT_UNSPECIFIED`
 | N/A
-| Payload format was not specified so we cannot assume a format and automatically deserialize the payload, the client needs to know what payload format is,
+| The payload format is unknown. Clients will need to have gained knowledge about the concrete encoding and semantics of the payload by means of an out-of-band mechanism.
 
 
 |`UPAYLOAD_FORMAT_PROTOBUF_WRAPPED_IN_ANY`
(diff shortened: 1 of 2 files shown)


sophokles73 (changes requested):
Making this change will also require a change to the corresponding specitem revision in the uattributes/umessage specification where we define that the payload format should be assumed to be WRAPPED_IN_ANY if not specified. This will make sure that OFT discovers the discrepancy between spec and implementation...

stevenhartley:
> Making this change will also require a change to the corresponding specitem revision in the uattributes/umessage specification where we define that the payload format should be assumed to be WRAPPED_IN_ANY if not specified. This will make sure that OFT discovers the discrepancy between spec and implementation...

There is no OFT requirement for this in upayload.adoc or in uattributes.adoc, it will need to be added on a subsequent PR

sophokles73 (approved):
I see, I will create a PR for adding some wording and OFT spec item then ...

LGTM

PR #248, opened 2024-11-11
Support UNSPECIFIED payload format in CloudEvents mapping

sophokles73:
This is a follow up of
https://github.com/eclipse-uprotocol/up-spec/pull/247


sophokles73:
@stevenhartley would you mind taking a look?

stevenhartley (changes requested):
We cannot get rid of datacontenttype. If we want to add a new one called pformat that we could but datacontenttype is defined in CE specs and it is needed for backwards compatibility with older versions of the protocol

sophokles73:
> We cannot get rid of datacontenttype. If we want to add a new one called pformat that we could but datacontenttype is defined in CE specs and it is needed for backwards compatibility with older versions of the protocol

What _older_ version of which protocol are you referring to? An old version of uProtocol? If so, what exactly are the expectations regarding the use of `datacontenttype` in that old version?

CloudEvents itself does not actually define `datacontenttype` to contain a value, based on my understanding of the spec ...

stevenhartley:
> > We cannot get rid of datacontenttype. If we want to add a new one called pformat that we could but datacontenttype is defined in CE specs and it is needed for backwards compatibility with older versions of the protocol
> 
> What _older_ version of which protocol are you referring to? An old version of uProtocol? If so, what exactly are the expectations regarding the use of `datacontenttype` in that old version?
> 
> CloudEvents itself does not actually define `datacontenttype` to contain a value, based on my understanding of the spec ...

Older versions of the protocol would reject anything that was not `application/x-protobuf` (what was in production) because AFAIK the PR I made to remove this attribute (and assume it is this type all the time) did not make it into the production code in the cloud.  

Cloudevent specifications does define the attribute as optional [here](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#datacontenttype). 

Having said that, I am simply asking that we do not remove the attribute definition and in fact leave the "if it is missing it is assumed to be `application/x-protobuf`). 

If we want to add a new attribute that sends an int in lie of a string, that is ok, but don't change the behavior of `datacontenttype`

stevenhartley:
> 2\. unset

I think we leave it as is meaning we keep `datacontenttype` and if it is not present we assume it is UPAYLOAD_PROTOBUF_WRAPPED_IN_ANY.

stevenhartley (commented):
@sophokles73 ,
i was thinking, since we are re-writing the cloudevent, why don't we get rid of the string names for the various non-mandatory CE types (i.e. everything other than id, type, source)? Since it is a key-value pair where the key is a string, `pformat` could be replaced with `12` and `priority` replaced with `5`, thus reducing even further the packet size. If we don't do this now, we will have to go to v2 later for the message types...

sophokles73:
> @sophokles73 , i was thinking, since we are re-writing the cloudevent, why don't we get rid of the string names for the various non-mandatory CE types (i.e. everything other than id, type, source)? Since it is a key-value pair where the key is a string, `pformat` could be replaced with `12` and `priority` replaced with `5`, thus reducing even further the packet size. If we don't do this now, we will have to go to v2 later for the message types...

I am a little concerned that having a mixture of _real_ key names and plain integers (IDs from the proto) will lead to a lot of confusion and many questions. Also, the `traceparent` attribute is actually a [common registered attribute in CloudEvents](https://github.com/cloudevents/spec/blob/main/cloudevents/extensions/distributed-tracing.md#traceparent) and using the numerical identifier instead would somehow be counterproductive, wouldn't it? FMPOV we should therefore stick with the key names instead of the IDs. WDYT?

stevenhartley (approved):

stevenhartley:
LGTM

Issue #221, opened 2024-08-02
Remove non-client facing APIs from uTwin proto and specification


stevenhartley:
Currently the uTwin specification includes both a Get() and Set() API however uEs only call/use the Get() API, the Set() was something specific to our Android uBus implementation and clients are never suppose to call Set() API. 

This PR shall be used to clean up the uTwin specs and proto.

PR #252, opened 2024-11-25
Remove Set API from uTwin

stevenhartley:
The set API is not part of the public APIs that clients will use so it needs to be removed.

#221


sophokles73 (approved):
LGTM

PR #255, opened 2024-12-18
Improve Language Library overview diagram

sophokles73:
Updated the overview diagram to adhere more closely to
UML syntax and semantics.


stevenhartley (commented):
Thanks for the cleanup, about the transport impl, why the extra dotted line box called TransportImpl inside of the TransportImpl package?

sophokles73:
> Thanks for the cleanup, about the transport impl, why the extra dotted line box called TransportImpl inside of the TransportImpl package?

The _TransportLibrary_ is a _template package_ and the dotted box indicates the parameter that the template uses. A concrete implementation then _binds_ a concrete type to the parameter, e.g. the MQTT transport library binds its `Mqtt5Transport` ...
See https://www.uml-diagrams.org/template.html

PR #256, opened 2025-01-10
Fix OFT spec item identifier

sophokles73:
A malformed OFT spec item identifier causees the OFT check to fail


PLeVasseur (approved):
I know you requested @stevenhartley, but I'll give you a thumbs up :+1:

stevenhartley (approved):
LGTM

PR #257, opened 2025-01-23
Make use of existing MQTT 5 header properties

sophokles73:
Adapted the mapping of UAttribute fields to make use of standard
properties defined in MQTT 5.

Also removed the distinct spec item identifiers defined for each field.
The mapping of the fields will most likely happen in very coherent
code blocks (functions) so it should be sufficient to trace the
requirements at that level.

I wonder if we should also adapt the mapping of the message ID to use hex encoding instead of the hyphenated string representation. WDYT? @PLeVasseur @ValMobBIllich

### DIFF
--- a/up-l1/mqtt_5.adoc
+++ b/up-l1/mqtt_5.adoc
@@ -50,121 +50,85 @@ An MQTT 5 PUBLISH packet that is used to convey a uProtocol message *MUST* conta
 | N/A
 | User Property [key: _uP_]
 a| The uProtocol version being used
-[.specitem,oft-sid="dsn~up-transport-mqtt5-attribute-version~1",oft-needs="impl,utest"]
---
+
 * *MUST* be set to the (major) version of the uProtocol specification that the message mapping adheres to. At the time of writing, the version is `1`.
 * Message consumers *MAY* use this property to determine that the (PUBLISH) packet contains a uProtocol message.
---
 
 | _id_
 | User Property [key: _1_]
 a| The unique identifier of the message
 
-[.specitem,oft-sid="dsn~up-transport-mqtt5-attribute-id~1",oft-needs="impl,utest"]
---
 * *MUST* be set to the https://www.rfc-editor.org/rfc/rfc4122.html#section-3[hyphenated string representation] of the UUID.
---
 
 | _type_
 | User Property [key: _2_]
 a| The message type
 
-[.specitem,oft-sid="dsn~up-transport-mqtt5-attribute-type~1",oft-needs="impl,utest"]
---
-* *MUST* be set to the value of the `uprotocol.ce_name` option defined for the
+* *MUST* be set to the value of the _uprotocol.ce_name_ option defined for the
 link:../up-core-api/uprotocol/uattributes.proto[UMessageType enum].
---
 
 | _source_
 | User Property [key: _3_]
 a| The origin (address) of the message
 
-[.specitem,oft-sid="dsn~up-transport-mqtt5-attribute-source~1",oft-needs="impl,utest"]
---
 * *MUST* be set to the link:../basics/uri.adoc#uri-definition[string serialization of the UUri]
---
 
 | _sink_
 | User Property [key: _4_]
 a| The destination (address) of the message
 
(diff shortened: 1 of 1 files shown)


sophokles73:
@ValMobBIllich you might also want to take a look ...

up-l1/mqtt_5.adoc
```diff
@@ -50,129 +50,93 @@ An MQTT 5 PUBLISH packet that is used to convey a uProtocol message *MUST* conta
...
 | _reqId_
-| User Property [key: _9_] 
+| _Correlation Data_
 a| The identifier that a service consumer can use to correlate an RPC response message with its RPC request
 
-[.specitem,oft-sid="dsn~up-transport-mqtt5-attribute-req-id~1",oft-needs="impl,utest"]
---
-* *MUST* be set to the https://www.rfc-editor.org/rfc/rfc4122.html#section-3[hyphenated string representation] of the UUID.
---
+* *MUST* be set to the UUID's 16 byte representation in big endian order.
```
stevenhartley, line 111: Why remove the reference link to the spec that defines the UUID format that it has to follow?

stevenhartley (commented):
small comment

up-l1/mqtt_5.adoc
sophokles73: The link was referring to the definition of the _hyphenated string representation_, which we are no longer using. The representation of the UUID's 16 bytes are, well, 16 bytes, not much formatting to do there ...

stevenhartley: duh of course.

stevenhartley (approved):
LGTM

up-l1/mqtt_5.adoc
sophokles73: Any thoughts regarding the encoding of the message ID as indicated in the PR's description?

up-l1/mqtt_5.adoc
```diff
@@ -50,129 +50,93 @@ An MQTT 5 PUBLISH packet that is used to convey a uProtocol message *MUST* conta
...
-a| The format for the data stored in the UMessage
+| _Content Type_
+a| The type of data contained in the message's payload
 
-[.specitem,oft-sid="dsn~up-transport-mqtt5-attribute-payload-format~1",oft-needs="impl,utest"]
---
 * *MUST* be set to the link:../up-core-api/uprotocol/v1/uattributes.proto[UPayloadFormat enum] integer value's decimal string representation.
---
+
+Note that the enum's integer value is used instead of the _uprotocol.mime_type_ option's value in order to reduce the overall size of the PUBLISH packet.
```
PLeVasseur, line 131: Seems reasonable 👍

up-l1/mqtt_5.adoc
```diff
@@ -50,129 +50,93 @@ An MQTT 5 PUBLISH packet that is used to convey a uProtocol message *MUST* conta
...
-[.specitem,oft-sid="dsn~up-transport-mqtt5-attribute-traceparent~1",oft-needs="impl,utest"]
---
 * *MUST* be set to the traceparent value.
---
 
 | _payload_format_
-| User Property [key: _12_] 
-a| The format for the data stored in the UMessage
+| _Content Type_
+a| The type of data contained in the message's payload
```
PLeVasseur, line 127: Good idea :)

up-l1/mqtt_5.adoc
```diff
@@ -50,129 +50,93 @@ An MQTT 5 PUBLISH packet that is used to convey a uProtocol message *MUST* conta
...
 a| A UCode indicating an error that has occurred during the delivery of either an RPC Request or Response message
 
-[.specitem,oft-sid="dsn~up-transport-mqtt5-attribute-comm-status~1",oft-needs="impl,utest"]
---
 * *MUST* be set to the link:../up-core-api/uprotocol/v1/ustatus.proto[UCode enum] integer value's decimal string representation.
---
 
 | _reqId_
-| User Property [key: _9_] 
+| _Correlation Data_
```
PLeVasseur, line 108: Interesting. Makes sense that it already exists!

PLeVasseur (approved):
LGTM. Thanks @sophokles73 for spotting where we can use existing header properties :)

PR #263, opened 2025-02-25
Fix requirements for preventing sender address spoofing

### LINKED ISSUE #262: UTransport spec erroneously prevents clients from publishing events
The UTransport's [send operation spec](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#21-send) currently requires implementers to fail the sending if the message's source address does not match the address associated with the transport.

However, the address associated with the transport is usually something like
`up://my_authority/ABCD/1/0`
where the resource ID is _always_ zero, i.e. contains the generic _inbound message_ resource ID. Based on that, a uEntity would not be able to publish an event to topic `up://my_authority/ABCD/1/9000` because the UUri does not match the UUri associated with the transport.

FMPOV we need to rephrase the spec to exclude the resource ID from the matching.
WDYT? @stevenhartley @PLeVasseur @AnotherDaniel

sophokles73:
Fixes #262

### DIFF
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -134,9 +134,11 @@ NOTE: The above strategy for retrying failed attempts to send a message results 
 * *MUST* fail with a `UCode.INVALID_ARGUMENT` if the passed UMessage failed validation
 --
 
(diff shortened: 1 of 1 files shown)


stevenhartley (approved):
LGTM!

PR #269, opened 2025-03-07
[#268] Retain sub-second granularity TTL value

### LINKED ISSUE #268: MQTT 5 spec does not retain original TTL
The [MQTT 5 transport spec](https://github.com/eclipse-uprotocol/up-spec/blob/9f7a3248cddd2df53401dc0180047bb1ac8bbf8d/up-l1/mqtt_5.adoc?plain=1#L93) currently fails to retain a UMessage's TTL value with sub-second granularity.

This is in violation of the [L1 API spec](https://github.com/eclipse-uprotocol/up-spec/blob/9f7a3248cddd2df53401dc0180047bb1ac8bbf8d/up-l1/README.adoc?plain=1#L138C1-L143C3) which requires

> UTransport implementations
>
> * *MUST* preserve all of the message's meta data and payload during transmission

sophokles73:
The mapping of a UMessage's TTL value to an MQTT 5 message's
Message Expiry Interval property using the ceil function has resulted
in the receiving side not being able to retain the original TTL value
if the TTL had sub-second granularity.

This has been fixed by requiring implementors to also add a User
Property containing the original TTL value in such cases.

Fixes #268

### DIFF
--- a/up-l1/mqtt_5.adoc
+++ b/up-l1/mqtt_5.adoc
@@ -1,6 +1,7 @@
 = MQTT 5 Transport Protocol
 :toc: preamble
 :sectnums:
+:stem: latexmath
 
 The key words "*MUST*", "*MUST NOT*", "*REQUIRED*", "*SHALL*", "*SHALL NOT*", "*SHOULD*", "*SHOULD NOT*", "*RECOMMENDED*", "*MAY*", and "*OPTIONAL*" in this document are to be interpreted as described in https://www.rfc-editor.org/info/bcp14[IETF BCP14 (RFC2119 & RFC8174)]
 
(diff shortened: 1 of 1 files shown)


up-l1/mqtt_5.adoc
```diff
@@ -87,10 +88,21 @@ a| The message's priority as defined in link:../basics/qos.adoc[QoS doc]
...
+a| The amount of time after which this message must no be delivered/processed anymore
     
-* *MUST* be set to _ceil(ttl/1000)_.
+The _Message Expiry Interval_ property *MUST* be set to
+
+stem:[expiry := ceil(ttl/1000)].
+
+The User Property *MUST* be set to the value of _ttl_ if, and only if
+
+stem:[ttl \bmod 1000 > 0].
```
PLeVasseur, line 102: For some reason this is not rendering for me properly when I got to `View file`. Any ideas?

up-l1/mqtt_5.adoc
```diff
@@ -87,10 +88,21 @@ a| The message's priority as defined in link:../basics/qos.adoc[QoS doc]
...
+
+This is necessary to retain the original value for the receiving end, because the MQTT message expiry interval has second granularity only.
```
PLeVasseur, line 104: Is the idea here to save a bit on bandwidth between cloud and device? I can see that not being sub-second granularity and could potentially add up over millions of messages.

PLeVasseur (approved):
Thanks @sophokles73! It was not rendering a portion for me correctly, but I'm not sure if that's just an issue on my end. Is it working for you when you `View file`?

up-l1/mqtt_5.adoc
sophokles73: Not on GitHub, but it renders properly in VSCode's Asciidoc preview.

up-l1/mqtt_5.adoc
sophokles73: Not really, maybe that is true, however. I just thought we should only include it if we need it ...

PR #271, opened 2025-03-12
Add OFT specification items

sophokles73:
Revisited basic specification documents and added and/or updated OpenFastTrace specification items where missing/appropriate.


sophokles73:
> Thanks @sophokles73! Support adding more OFT to the spec.
> 
> I'll be honest that I haven't had to learn OFT much yet myself, so I often skim over these PRs. Is there any tooling we can include in CI from OFT which would check that they are filled out or at least formatted correctly?

Good question. Let me see if I can find anything in the OpenFastTrace docs.

AnotherDaniel (approved):
LGTM

sophokles73:
> > Thanks @sophokles73! Support adding more OFT to the spec.
> > I'll be honest that I haven't had to learn OFT much yet myself, so I often skim over these PRs. Is there any tooling we can include in CI from OFT which would check that they are filled out or at least formatted correctly?
> 
> Good question. Let me see if I can find anything in the OpenFastTrace docs.

@PLeVasseur OpenFastTrace's `convert` command could be used to (try to) read the specification items from the documents and convert them to an XML format. This would implicitly perform all the syntax checks on the specitems found. However, I do not believe that this has much value (yet) because most of the data in specitem is optional and thus missing data would not lead to a failure. Additionally, any misspelled property names would also go undetected etc.

PLeVasseur:
> However, I do not believe that this has much value (yet) because most of the data in specitem is optional and thus missing data would not lead to a failure. Additionally, any misspelled property names would also go undetected etc.

Ah got it. Worth asking into, but seems the answer is it wouldn't add much value.

PR #274, opened 2025-03-28
Fix specitem identifier for response priority value

sophokles73:
The specification item that defines requirements for an RPC response
message's priority attribute had been using a misleading identifier.
This has been fixed.


AnotherDaniel (approved):
👍

Issue #275, opened 2025-04-03
Remove empty Spec content


stevenhartley:
This ticket is to cleanup the parts of the spec that are empty (ex. transports that are not implemented). 
goal is to have a baseline for beta.1 release.

stevenhartley:
Closing issue as changes were merged

PR #276, opened 2025-04-04
Remove empty transport spec documents

sophokles73:
Several of the anticipated transport bindings have not been defined
yet. Unless somebody takes an interest in writing them, it seems more
reasonable to remove the empty documents from the spec.

This addresses #275


stevenhartley (approved):
LGTM

PR #279, opened 2025-04-15
Add Gherkin Feature files

sophokles73:
The Feature files are supposed to serve as "test vectors" for
asserting compliance with (aspects of) the specification.

Addresses https://github.com/eclipse-uprotocol/up-tck/issues/116

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -66,3 +66,7 @@ The uProtocol specification documents contain a lot of requirements that are def
 | `TransportLayerImplPush` | Requirements that need to be covered by concrete implementations of xref:up-l1/README.adoc[uProtocol's Transport Layer API] supporting the the xref:up-l1/README.adoc#delivery-method[push message delivery] method
 
 |===
+
+=== Gherkin Feature Files
+
+The [features](features) folder contains Gherkin Features and Scenarios which can be used with test frameworks like [Cucumber](https://cucumber.io/) to assert compliance of implementations with the uProtocol specification. Not all parts of the specification are covered (yet), but we intend to add Scenario descriptions so that more and more of the requirements defined in the specification will be covered over time.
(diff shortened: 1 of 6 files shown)


sophokles73 (commented):
I wonder if it would be better to keep the feature files closer to the corresoponding adoc files where the requirements are specified. That way it would probably be easier to keep them consistent/in sync when making changes to one or the other ...

AnotherDaniel (approved):
I think once we have a basic 'productive' implementation for this overall setup, it will be valuable to add a document that explains the overall approach and highlights best-practice implementations in uproto components - for others to copy from, and for us to reference as part of the software quality process description materials.

PLeVasseur (approved):
Thanks @sophokles73 for taking the lead on the feature files! LGTM

btw I agree with this:
>  wonder if it would be better to keep the feature files closer to the corresoponding adoc files where the requirements are specified. That way it would probably be easier to keep them consistent/in sync when making changes to one or the other ...

sophokles73:
> Thanks @sophokles73 for taking the lead on the feature files! LGTM
> 
> btw I agree with this:
> 
> > wonder if it would be better to keep the feature files closer to the corresoponding adoc files where the requirements are specified. That way it would probably be easier to keep them consistent/in sync when making changes to one or the other ...

I will move the Gherkin files to the adoc files in a separate commit and then we can decide if we want to keep that or revert ...

sophokles73:
@PLeVasseur I have moved the feature files into the `basic` folder so that they no reside next to the corresponding spec files. I have also verified in my up-rust workspace that I can selectively refer to them from within the cucumber based tests.

If you agree, I can squash and merge this now ...

stevenhartley (approved):
LGTM

PLeVasseur (approved):
LGTM! Thank you @sophokles73

PR #283, opened 2025-06-04
Remove subscriber/subscription list pagination

AnotherDaniel:
uSubscription: as per discussion in #282, this PR removes protobuf attributes and requirements associated with the pagination of lists returned by getSubscribers/getSubscriptions.

I've not taken the detour of deprecating these attributes first - as we're not in production for now, I'm thinking we can directly remove&reserve these properties.


sophokles73 (approved):
LGTM

PR #284, opened 2025-06-04
Fix typo in requirement ID that breaks oft runs

AnotherDaniel:
This typo makes oft bark on requirements analysis runs, so fix it.


sophokles73 (approved):
LGTM

PLeVasseur (approved):
Less broken things = more good, thanks @AnotherDaniel!

PR #285, opened 2025-06-07
Add iceoryx2 UTransport specification

PLeVasseur:
closes https://github.com/eclipse-uprotocol/up-transport-iceoryx2-rust/issues/1

### DIFF
--- a/up-l1/iceoryx2.adoc
+++ b/up-l1/iceoryx2.adoc
@@ -1,0 +1,336 @@
+= iceoryx2
+:toc: preamble
+:sectnums:
+
+The key words "*MUST*", "*MUST NOT*", "*REQUIRED*", "*SHALL*", "*SHALL NOT*", "*SHOULD*", "*SHOULD NOT*", "*RECOMMENDED*", "*MAY*", and "*OPTIONAL*" in this document are to be interpreted as described in https://www.rfc-editor.org/info/bcp14[IETF BCP14 (RFC2119 & RFC8174)]
+
+----
+SPDX-FileCopyrightText: 2025 Contributors to the Eclipse Foundation
+
+See the NOTICE file(s) distributed with this work for additional
+information regarding copyright ownership.
+
+This program and the accompanying materials are made available under
+the terms of the Apache License Version 2.0 which is available at
+https://www.apache.org/licenses/LICENSE-2.0
+ 
+SPDX-FileType: DOCUMENTATION
+SPDX-License-Identifier: Apache-2.0
+----
+
+== Overview
+
+https://iceoryx.io[Eclipse iceoryx2&trade;] is an efficient, and ultra-low latency inter-process communication middleware. The library is designed to provide you with fast and reliable zero-copy and lock-free inter-process communication mechanisms.
+
+This document defines how uProtocol messages can be exchanged by means of iceoryx2 _Services_.
+
+== iceoryx2 Version
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-protocol-version~1",oft-needs="impl",oft-tags="TransportLayerImpl"]
+--
+Each transport implementing this specification **MUST** use version `0.6.1` of the iceoryx2 protocol in order to ensure interoperability of different language libraries.
+--
+
+== MessagingPattern
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-messaging-pattern~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+All types of uProtocol messages *MUST* be transferred using iceoryx2's PublishSubscribe MessagingPattern only. iceoryx2's RequestResponse API *MUST NOT* be used.
+--
+
+== UMessage Mapping
+
+A uProtocol message consists of _UAttributes_ and optional payload. The following sections define how these are mapped to/from iceoryx2 Services.
+
+=== UAttributes
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-attributes-mapping~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+The value of a iceoryx2 Service that is used to convey a uProtocol message *MUST* have a https://docs.rs/iceoryx2/0.6.1/iceoryx2/service/builder/publish_subscribe/struct.Builder.html#method.user_header[user header] as defined in <<uAttributes Mapping to iceoryx2 custom header>>.
+--
+
+==== uAttributes Mapping to iceoryx2 user header
+
+[source,rust]
+----
+#[repr(C)]
+#[derive(ZeroCopySend)]
+struct UProtocolHeader {
+    id: UUid,
+    type: MessageType,
+    source: UUri,
+    sink: UUri,
+    ttl: u32,
+    permission_level: u32,
+    commstatus: UStatus,
+    reqid: UUid,
+    token: FixedSizeByteString<255>,
+    traceparent: FixedSizeByteString<255>,
+    payload_format: UPayloadFormat,
+}
+----
+
+==== Uuid type for iceoryx2
+
+[source,rust]
+----
+#[repr(C)]
+#[derive(ZeroCopySend)]
+struct UUid {
+    msb: u64,
+    lsb: u64,
+}
+----
+
+==== MessageType enum for iceoryx2
+
+[source,rust]
+----
+#[repr(C)]
+#[derive(ZeroCopySend)]
+enum MessageType {
+    Unspecified = 0,
+    Publish = 1,
+    Request = 2,
+    Response = 3,
+    Notification = 4,
+}
+----
+
+==== UUri for iceoryx2
+
+[source,rust]
+----
+#[repr(C)]
+#[derive(ZeroCopySend)]
+struct UUri {
+    authority_name: FixedSizeByteString<255>,
+    ue_id: u32,
+    ue_major_version: u32,
+    resource_id: u32,
+}
+----
+
+==== UPriority enum for iceoryx2
+
+[source,rust]
+----
+#[repr(C)]
+#[derive(ZeroCopySend)]
+enum UPriority {
+    Unspecified = 0,
+    CS0 = 1,
+    CS1 = 2,
+    CS2 = 3,
+    CS3 = 4,
+    CS4 = 5,
+    CS5 = 6,
+    CS6 = 7,
+}
+----
+
+==== UStatus for iceoryx2
+
+[source,rust]
+----
+#[repr(C)]
+#[derive(ZeroCopySend)]
+struct UStatus {
+    code: UCode,
+    message: FixedSizeByteString<255>,
+    // leaving details empty intentionally as with
+    // as with iceoryx2 shared-memory, zero-copy
+    // we must have a fixed size in memory
+}
+----
+
+==== UCode enum for iceoryx2
+
+[source,rust]
+----
+#[repr(C)]
+#[derive(ZeroCopySend)]
+enum UCode {
+    Ok = 0,
+    Cancelled = 1,
+    Unknown = 2,
+    InvalidArgument = 3,
+    DeadlineExceeded = 4,
+    NotFound = 5,
+    AlreadyExists = 6,
+    PermissonDenied = 7,
+    Unauthenticated = 16,
+    ResourceExhausted = 8,
+    FailedPrecondition = 9,
+    Aborted = 10,
+    OutOfRange = 11,
+    Unimplemented = 12,
+    Internal = 13,
+    Unavailable = 14,
+    DataLoss = 15,
+}
+----
+
+==== UPayloadFormat enum for iceoryx2
+
+[source,rust]
+----
+#[repr(C)]
+#[derive(ZeroCopySend)]
+enum UPayloadFormat {
+    Unspecified = 0,
+    ProtoBufWrappedInAny = 1,
+    ProtoBuf = 2,
+    Json = 3,
+    Someip = 4,
+    SomeipTlv = 5,
+    Raw = 6,
+    Text = 7,
+    Shm = 8,
+}
+----
+
+
+==== Message Priority
+
+TODO(plevasseur@gmail.com): I don't see mention of a way to set priority natively for a Service in the iceoryx2 docs.
+
+=== Payload Mapping
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-mapping~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+An iceoryx2 message that is used to convey a uProtocol message *MUST* contain in its payload the unaltered value of the UMessage's _payload_ field.
+--
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-mapping~2",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+An iceoryx2 message that is used to convey a uProtocol message *MUST* be zero-copy and shared-memory compatible (i.e. be fixed in size, contain no pointers, and so on).
+--
+
+== iceoryx2 Service Name Structure
+
+Message producers publish messages to the iceory2 inter-process communication using _Service Names_. Other clients can indicate their interest in particular  or patterns in order to receive the messages that are being published using matching keys.
+
+The iceoryx2 _Service Name_ that is used to transfer a uProtocol message is derived from the message's `source` and `sink` attributes.
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-service-name~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+The ServiceName of an iceoryx2 message containing a _Publish_ UMessage **MUST** consist of the following segments:
+
+`up/[source.authority]/[source.ue_type]/[source.ue_instance]/[source.ue_version]/[source.resource]/{}/{}/{}/{}/{}`
+
+The ServiceName of an iceoryx2 message containing a _Notification_, _RPC Request_ or _RPC Response_ UMessage **MUST** consist of the following segments:
+
+`up/[source.authority]/[source.ue_type]/[source.ue_instance]/[source.ue_version]/[source.resource]/[sink.authority]/[sink.ue_type]/[sink.ue_instance]/[sink.ue_version]/[sink.resource]`
+
+Please refer to <<UUri Encoding Rules>> for details regarding the encoding of the `source` and `sink` UUris into the key expression's segments.
+--
+
+=== UUri Encoding Rules
+
+The table below contains the rules for encoding a UUri's fields into an iceoryx2 ServiceName's segments.
+
+TODO(plevasseur@gmail.com): I don't see mention of the ability to have a wildcard in a ServiceName in the iceoryx2 docs. If I'm wrong, will need to revise to add wildcards in below.
+
+[cols="2,2,6"]
+|===
+| ServiceName Segment
+| UUri Field
+| Encoding
+
+|`authority`
+|`authority_name`
+a| The segment *MUST* contain the (UTF8) string representation of the 
+
+1. name of the host/authority that the (local) uEntity is running on, if authority name is empty.
+2. authority name, otherwise.
+
+|`ue_type`
+|`ue_id`
+a| The segment *MUST* contain the (UTF8) string representation of the
+
+the upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the uEntity type identifier with all leading `0` characters omitted.
+
+|`ue_instance`
+|`ue_id`
+a| The segment *MUST* contain the (UTF8) string representation of the
+
+the upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the uEntity instance identifier with all leading `0` characters omitted.
+
+|`ue_version`
+|`ue_version_major`
+a| The segment *MUST* contain the (UTF8) string representation of the
+
+the upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the uEntity major version with all leading `0` characters omitted.
+
+|`resource`
+|`resource_id`
+a| The segment *MUST* contain the (UTF8) string representation of the
+
+the upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the resource identifier with all leading `0` characters omitted.
+
+|===
+
+=== Examples
+
+The examples below assume that the local entity's authority name is `device1`.
+
+.Publishing an event on a topic
+--
+[cols="2,8"]
+|===
+|*Source URI*
+|`up:/10AB/3/80CD`
+
+|*Sink URI*
+|-
+
+|*iceoryx2 ServiceName*
+|`up/device1/10AB/0/3/80CD/{}/{}/{}/{}/{}`
+|===
+--
+
+.Sending a Notification to another uEntity
+--
+[cols="2,8"]
+|===
+|*Source URI*
+|`up://device1/10AB/3/80CD`
+
+|*Sink URI*
+|`up://device2/300EF/4/0`
+
+|*iceoryx2 ServiceName*
+|`up/device1/10AB/0/3/80CD/device2/EF/3/4/0`
+|===
+--
+
+.Sending an RPC Request to a service provider
+--
+[cols="2,8"]
+|===
+|*Source URI*
+|`up:/403AB/3/0`
+
+|*Sink URI*
+|`up://device2/CD/4/B`
+
+|*iceoryx2 ServiceName*
+|`up/device1/3AB/4/3/0/device2/CD/0/4/B`
+|===
+--
+
+.Subscribe to a specific topic
+--
+[cols="2,8"]
+|===
+|*Source Filter*
+|`up://device2/10AB/3/80CD`
+
+|*Sink Filter*
+|-
+
+|*iceoryx2 ServiceName*
+|`up/device2/10AB/0/3/80CD/{}/{}/{}/{}/{}`
+|===
+--


up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,336 @@
...
+
+== UMessage Mapping
+
+A uProtocol message consists of _UAttributes_ and optional payload. The following sections define how these are mapped to/from iceoryx2 Services.
+
+=== UAttributes
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-attributes-mapping~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+The value of a iceoryx2 Service that is used to convey a uProtocol message *MUST* have a https://docs.rs/iceoryx2/0.6.1/iceoryx2/service/builder/publish_subscribe/struct.Builder.html#method.user_header[user header] as defined in <<uAttributes Mapping to iceoryx2 custom header>>.
```
sophokles73, line 49:
```suggestion
The value of a iceoryx2 Service that is used to convey a uProtocol message *MUST* have a https://docs.rs/iceoryx2/0.6.1/iceoryx2/service/builder/publish_subscribe/struct.Builder.html#method.user_header[user header] as defined in <<uAttributes Mapping to iceoryx2 user header>>.
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,336 @@
...
+--
+
+==== uAttributes Mapping to iceoryx2 user header
```
sophokles73, line 52: It seems undesirable to be required to do a manual mapping of Rust structs to other Rust structs that are basically the same thing with two additional attributes defined on them :-(
Based on that, I wonder if it wouldn't make more sense to take the same approach as we do in the Zenoh transport, that is, serializing the metadata to its protobuf and add that as (together with the uProtocol major version) into the header?

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,336 @@
...
+--
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-mapping~2",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+An iceoryx2 message that is used to convey a uProtocol message *MUST* be zero-copy and shared-memory compatible (i.e. be fixed in size, contain no pointers, and so on).
+--
+
+== iceoryx2 Service Name Structure
+
+Message producers publish messages to the iceory2 inter-process communication using _Service Names_. Other clients can indicate their interest in particular  or patterns in order to receive the messages that are being published using matching keys.
```
sophokles73, line 212:
```suggestion
Message producers publish messages to the iceory2 inter-process communication using _Service Names_. Other clients can indicate their interest in particular Services or patterns in order to receive the messages that are being published using matching names.
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,336 @@
...
+
+The iceoryx2 _Service Name_ that is used to transfer a uProtocol message is derived from the message's `source` and `sink` attributes.
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-service-name~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+The ServiceName of an iceoryx2 message containing a _Publish_ UMessage **MUST** consist of the following segments:
```
sophokles73, line 218: is it *ServiceName* or *Service Name*? We should use the term consistently across the spec ...

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,336 @@
...
+
+`up/[source.authority]/[source.ue_type]/[source.ue_instance]/[source.ue_version]/[source.resource]/{}/{}/{}/{}/{}`
```
sophokles73, line 220: based on a quick glance at the iceoryx2 docs, I am not sure if wildcards are supported in `ServiceName`s ...

sophokles73 (changes requested):

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,336 @@
...
+=== Payload Mapping
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-mapping~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+An iceoryx2 message that is used to convey a uProtocol message *MUST* contain in its payload the unaltered value of the UMessage's _payload_ field.
+--
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-mapping~2",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+An iceoryx2 message that is used to convey a uProtocol message *MUST* be zero-copy and shared-memory compatible (i.e. be fixed in size, contain no pointers, and so on).
```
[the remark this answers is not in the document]
sophokles73: Then I would suggest that we explicitly state here that the payload must implement the `ZeroCopySend` trait. We seem to be talking about Rust based implementations only anyway ...

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,336 @@
...
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-protocol-version~1",oft-needs="impl",oft-tags="TransportLayerImpl"]
+--
+Each transport implementing this specification **MUST** use version `0.6.1` of the iceoryx2 protocol in order to ensure interoperability of different language libraries.
+--
+
+== MessagingPattern
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-messaging-pattern~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+All types of uProtocol messages *MUST* be transferred using iceoryx2's PublishSubscribe MessagingPattern only. iceoryx2's RequestResponse API *MUST NOT* be used.
```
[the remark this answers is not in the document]
elfenpiff: @sophokles73 It would introduce a syscall on the sender side, which would cause latency, and when the receiver never acquires the notifications/samples the underlying buffer of the event mechanism might run out of memory, which the user has to handle - but this depends on the mechanism the user would like to use for events. iceoryx2 provides here semaphores (which do not have the problem but are not interoperable with file descriptors in the context of an event multiplexer like `select` or `epoll`) and unix datagram sockets that are interoperable with fds and event multiplexer but are not certifiable and have the buffer problem.

up-l1/iceoryx2.adoc
PLeVasseur: Thanks for the feedback @elfenpiff!

> Wildcards are not supported.

These are not supported in the ServiceName and never will be?

---------

Also, to make sure I am clear on how we should be using iceoryx2, I have some questions.

Here's what a uProtocol UUri looks like, I'll reference this:
image

(pulled from [this](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc) other spec page)

### First example

Let's say that I had some "service A" (I use this not in the iceoryx2 definition) which sends two messages representing two different pieces of information `Foo`s and `Bar`s RESOURCE_IDs.

Would I have a _single_ [`Service`](https://docs.rs/iceoryx2/latest/iceoryx2/service/trait.Service.html) for "service A" sending its two messages?

Or _two_ [`Service`](https://docs.rs/iceoryx2/latest/iceoryx2/service/trait.Service.html)s for "service A" sending its two messages?

### Second example

Now imagine that I have two services: "service A" and "service B".

"Service A" sends two messages representing two different pieces of information `Foo`s and `Bar`s RESOURCE_IDs.

And I have some "service B" which wants to subscribe to _all_ resources originating from "serviceA".

Is it possible to do this?

Would that then mean that "service A" should be two [`Service`](https://docs.rs/iceoryx2/latest/iceoryx2/service/trait.Service.html)s to allow for the attaching of attributes to each service indicating that the source UUri is the same except for the RESOURCE_ID?

And then in "service B" it would be possible to iterate over the services, picking out all for which the:
* AUTHORITY_NAME
* ENTITY_ID
* ENTITY_VERSION_MAJOR

matched and then subscribing to them?

Sorry this got quite long 😅 Hope that was clear-ish

PLeVasseur (commented):
Thanks for the feedback @sophokles73 & @elfenpiff!

I've left some follow-up questions 👍

up-l1/iceoryx2.adoc
PLeVasseur: > This is not correct. iceoryx2 supports dynamic payload, see this example here: https://github.com/eclipse-iceoryx/iceoryx2/tree/main/examples/rust/publish_subscribe_dynamic_data
> 
> Currently, the user can send a strongly typed slices, and with `loan_slice`, you can state how many elements you would like to send at runtime. If the `loan_slice` would exceed the maximum available memory iceoryx2 reallocates memory with a defined `AllocationStrategy`.

Oh! That's pretty cool. Thank you for sharing this example. (And sorry for missing it! 😅)

> It does not yet work with dynamic data which is part of a more complex data type like a struct. So this
> 
> ```rust
> struct MyPayload {
>   data1: RelocatableVec<u64>,
>   data2: RelocatableString
> }
> ```
> 
> Does not yet work but it is on the roadmap.
> 
> But this is a feature for desktop systems or proof-of-concept code. In mission-critical systems this feature would explicitly deactivated and the code would not compile when it is flagged as ASIL-D for instance.

Just to make sure I understood this --
* is it that using loaned slices would not compile for ASIL-D or
* that using the `RelocatableVec` or `RelocatableString` would not compile for ASIL-D

?

up-l1/iceoryx2.adoc
PLeVasseur: Ah I saw a couple more we _might_ be able to tackle via text:

> But if you want to be able to communication `X` and `T` over the same service, you could create a sum type (enum in Rust, variant in C++) and use it as payload.
> 
> ```rust
> pub enum MyPayload {
>   X(X),
>   Y(Y)
> }
> ```

This would enable one or the other type to be sent, but if we wanted to subscribe specifically to one resource out of a possible two, we'd have to follow your advice here, right?

> * Second example:
> 
> > Would that then mean that "service A" should be two [Service]>(https://docs.rs/iceoryx2/latest/iceoryx2/service/trait.Service.html)s to allow for the attaching of attributes to each service indicating that the source UUri is the same except for the RESOURCE_ID?
> > And then in "service B" it would be possible to iterate over the services, picking out all for which the:
> > AUTHORITY_NAME
> > ENTITY_ID
> > ENTITY_VERSION_MAJOR
> > matched and then subscribing to them?
> 
> This is exactly how you would realize it.

PLeVasseur: > But maybe it is enough to just support * and then you could write My*Service*Fuu. But even this could pose some interesting challenges especially for certification - but nothing we couldn't handle.

For example, the Zenoh UTransport spec sections [4.1. UUri Encoding Rules](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/zenoh.adoc#41-uuri-encoding-rules) and [4.2. Examples](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/zenoh.adoc#42-examples) shows a bit of what we'd aim to achieve.

Bits like this:

```
up://*/FFFFFFFF/FF/FFFF
```

or

```
up://*/FFFF05A1/2/FFFF
```

I suppose this may be a bit "simpler" alternative to the approach of using Attributes, but what do you think?

(again -- if these are too much to answer here we can wait till we meet up!)

up-l1/iceoryx2.adoc
elfenpiff: No, they would compile for ASIL-D but the `AllocationStrategy` cannot be defined in ASIL-D since everything is statically configured and pre-allocated. So if `loan_slice` would exceed the maximum available memory you would get an out-of-memory Err.

up-l1/iceoryx2.adoc
elfenpiff: > This would enable one or the other type to be sent, but if we wanted to subscribe specifically to one resource out of a possible two, we'd have to follow your advice here, right?

Yes

PLeVasseur:
Summarizing where I think the discussion landed / was headed after our meeting today with @elfenpiff @elBoberido before I go and make pretty significant modifications:

### uEntity  1 <--------> 1..* iceoryx2 Service

There will be one or more iceoryx2 Services to support the semantics of each uProtocol uEntity "under the hood" in the UTransport. The 1 to N mapping is required because iceoryx2 uses a different (correct? 😆) definition of a service that shall do only one thing.

In the case of uProtocol a uEntity, call it `uEntity A` may be offering up various services as an RpcServer such as:
* `up://[sink.authority]/[sink.ue_type]/[sink.ue_instance]/[sink.ue_version]/[sink.resource.a]` and
* `up://[sink.authority]/[sink.ue_type]/[sink.ue_instance]/[sink.ue_version]/[sink.resource.b]`

#### iceoryx2 implementation in `uEntity A`, hidden underneath the UTransport implementation

This can be implemented as follows:
* two iceoryx2 Services hosted out of `uEntity A`'s `up-transport-iceoryx2-rust`
  * one with ServiceName of "up/[sink.authority]/[sink.ue_type]/[sink.ue_instance]/[sink.ue_version]/[sink.resource.a]"
  * one with ServiceName of "up/[sink.authority]/[sink.ue_type]/[sink.ue_instance]/[sink.ue_version]/[sink.resource.b]"
  
  Service with ServiceName of "up/[sink.authority]/[sink.ue_type]/[sink.ue_instance]/[sink.ue_version]/[sink.resource.a]":
  * attaches attributes as follows
    * USchema with version of `1`
    * UAuthority with `sink.authority`
    * UeType with `sink.ue_type`
    * UeInstance with `sink.ue_instance`
    * UVersionMajor of `sink.ue_version`
    * UResourceId of `sink.resource.a`

  Service with ServiceName of "up/[sink.authority]/[sink.ue_type]/[sink.ue_instance]/[sink.ue_version]/[sink.resource.b]":
  * attaches attributes as follows
    * USchema with version of `1`
    * UAuthority with `sink.authority`
    * UeType with `sink.ue_type`
    * UeInstance with `sink.ue_instance`
    * UVersionMajor of `sink.ue_version`
    * UResourceId of `sink.resource.b`
 
 #### iceoryx2 implementation in `uEntity B`, hidden underneath the UTransport implementation
 
 If we're interested in only communicating to the certain method "up/[sink.authority]/[sink.ue_type]/[sink.ue_instance]/[sink.ue_version]/[sink.resource.b]"

to be continued... ran out of time!

Remaining
* flesh out a case of pub/sub with only wanting certain topics will then need to use iceoryx2 discovery mechanism by making use of the attributes to filter for only resource.a but not resource.b

sophokles73:
I agree in large parts. IMHO we should try to tackle the whole thing coming from the streamer use case as it is the most generic one and all other use cases will most likely only be specific scenarios of it.

A streamer will most likely always use/require wildcard filter UUris for registering the listeners for its routes. However, other uEntities will often use direct point-to-point message exchanges only and use specific filter UUris (not containing any wildcards) for registering listeners. It would be nice, if we could support at least the latter scenarios based on iceoryx2's standard publish/subscribe API, i.e. just using the ServiceBuilder with the corresponding Service Name on both the sending and receiving side.

For this to work, I propose to use the same mapping of source/sink UUris to Service Names as we use for (in-vehicle) MQTT5 topics.

A transport on the sending side uses iceoryx2's `publish_subscribe::Builder::open_or_create_with_attributes` function to either initially create the corresponding Service or open it, if it already exists. The attributes [defined above](https://github.com/eclipse-uprotocol/up-spec/pull/285#issuecomment-2981874679) seem appropriate for this purpose. The sender side transport then sends the message via the created `Publisher`.

A transport on the receiving side creates and uses a `service_discovery::Service` to get notified about new Services being created or removed (when a sender sends a message to it). The transport then creates an iceoryx2 `Subscriber` for each Service that matches any of its listeners' (wildcard) filter UUris. Note that this is a dynamic process and Services will be added and removed during the lifetime of the receiving transport.

The main concern with this approach is race conditions resulting from the asynchronous nature of discovering the Services dynamically. When a sender creates a new Service and sends its message, the receiving side might not yet have been notified about the newly created Service and thus the message might not be received as intended. However, this should not be an issue when publishing to topics or sending notifications due to the at-most-once delivery semantics. For RPC calls, the worst that can happen is that a request times out and the sender can decide to re-send its request message (at-least-once delivery).

The corresponding ServiceName patterns are:

| Scenario | ServiceName |
|:---------|:------------|
| publishing to a topic | `up/{source.authority}/{source.type_id}/{source.instance_id}/{source.major_version}/{source.resource_id}` |
| sending a notifcation, RPC request/response | `up/{source.authority}/{source.type_id}/{source.instance_id}/{source.major_version}/{source.resource_id}/{sink.authority}/{sink.type_id}/{sink.instance_id}/{sink.major_version}/{sink.resource_id}` |

PLeVasseur:
Hey @sophokles73 & @stevenhartley -- I've further fleshed out the details in [the comment I left above](https://github.com/eclipse-uprotocol/up-spec/pull/285#issuecomment-2981874679).

I do see some pains around RPC that I'd like your help in diagnosing.

Essentially you've got an 2 x N x M set of iceoryx2 Services in an RpcServer, which was starting to feel excessive.

Why?
* N => number of invokable methods on the RpcServer
* M => number of RpcClients
* 2 => we need to have one set of iceoryx2 Services for in-coming Requests and one set of iceoryx2 Services for out-going responses

I understand that we'd like to use the most fundamental messaging paradigm of the underlying protocol to "control" more of how uProtocol operates in the beginning. But perhaps worth exploring what this would look like if we used iceoryx2's [Request-Response](https://docs.rs/iceoryx2/0.6.1/iceoryx2/index.html#request-response) pattern for this case?

stevenhartley:
@sophokles73 & @PLeVasseur ,

In order to align the UTransport APIs usecase and mapping it to the ServiceName in Iceorxy, here is what I was thinking:

| UMessageType | ServiceName |
| ----------------- | -------------- |
| PUBLISH   | `up:/[SOURCE_UURI]` |
| REQUEST  | `up:[SINK_UUIR]` |
| RESPONSE  | `up:[REQUEST_SOURCE_UUIR]/[REQUEST_SINK_UURI]` |
| NOTIFICATION | `up:[SOURCE_UURI]/[SINK_UURI]` |

## Sending Messages

**API:** `send(umessage : UMessage)`

You would then use the source and sink UUris in the `UMessage` passed to `Utransport::send()` to Iceoryx "publish()" the message (to send the message type). 

## Receiving messages

**API:**  `registerListener(sourceFilter : UUri, sinkfilter : UUri, listener UListener)` 

The passed `sinkFilter` and `sourceFilter` are then also used to build the topic for the iceoryx "subscribe" API to receive the different message types. 

Hopefully this makes sense. Then from this we can start implementing the APIs and do a quick test of one uE building a UMessage of the 4 different types, and the other one receiving the uMessages.

stevenhartley:
> > The passed `sinkFilter` and `sourceFilter` are then also used to build the topic for the iceoryx _subscribe_ API to receive the different message types.
> 
> As I pointed out earlier, IMHO the filter UUris cannot be used directly with iceoryx2 _subscribe_ API which is based on ServiceNames.
> 
> Assume a streamer that should forward all messages matching a sink filter of `up://off_vehicle/FFFFFFFF/FF/FFFF`. This filter UUri cannot be used as a ServiceName because no uEntity actually sends a message to such a Service. Instead, the streamer would need to use the DiscoveryService to find matching Services being registered (by the sending uEntity), and then opening a _subscriber_ on the concrete ServiceName being used ...

I was only thinking of the uE to uE communication and not the streamer use case to start, the thought was that we would use service flags and discovery for device-2-device communication through the streamer (once we figured that out)

stevenhartley:
> Isn't RPC a special case that requires the step I outlined above for an RpcServer to obtain client Uri? For Notification and Publish message types the notifyee and the subscriber should be able to completely specify the Uri to subscribe to. Unless I'm missing something you could help explain.

to @sophokles73 point, the sender is setting the topics and the receiver needs to know what to listen to (iceoryx subscribe()" to). What I outlined above should work for all flows and use cases EXCEPT for the streamer usecase, that is where things breakdown. We need to dive further into that later but I don't think it is a showstoper.

stevenhartley:
> I think we're agreeing?

yes :-)

PLeVasseur:
Hey @sophokles73 & @stevenhartley -- could you take a look at the clarification and simplification I attempted in [0867bfb](https://github.com/eclipse-uprotocol/up-spec/pull/285/commits/0867bfb45ae06832f1941451e7131cb2546d68a2)?

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,264 @@
...
+[source,rust]
+----
+const MAX_FEASIBLE_UATTRIBUTES_SERIALIZED_LENGTH: usize = 1000; // choosing ~1000 u8s
+                                                                // somewhat arbitrarily
+                                                                // this should be confirmed
+
+#[repr(C)]
+#[derive(ZeroCopySend)]
+struct UProtocolHeader {
+    uprotocol_schema_version: u8,
```
sophokles73, line 63:
```suggestion
    uprotocol_major_version: u8,
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,264 @@
...
+    uattributes_serialized: FixedSizeVec<u8, MAX_FEASIBLE_UATTRIBUTES_SERIALIZED_LENGTH>
+}
+----
+
+.uAttributes Mapping to UProtocolHeader, iceoryx2 User Header
+[%autowidth]
+|===
+| field | value
+
+| `uprotocol_schema_version`
```
sophokles73, line 73:
```suggestion
| `uprotocol_major_version`
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,264 @@
...
+Currently not a clear path to map from uProtocol message priority to iceoryx2 QoS (Quality of Service). Skip for current revision.
+
+=== Payload Mapping
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-mapping~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+An iceoryx2 message that is used to convey a uProtocol message *MUST* contain in its payload the unaltered value of the UMessage's _payload_ field.
+--
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-mapping~2",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
```
sophokles73, line 94:
```suggestion
[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-zero-copy~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
```

this needs to get a different name, the number at the end of the ID is the *revision* of the specitem ...

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,264 @@
...
+--
+An iceoryx2 message that is used to convey a uProtocol message *MUST* be zero-copy and shared-memory compatible (i.e. be fixed in size, contain no pointers, and so on).
+--
+
+== iceoryx2 Service Name Structure
+
+Message producers publish messages to the iceory2 inter-process communication using _ServiceNames_. Other clients can indicate their interest in particular patterns in order to receive the messages that are being published using matching ServiceNames.
```
sophokles73, line 101:
```suggestion
Message producers publish messages to the iceory2 inter-process communication using _ServiceNames_. Other clients can then use the same _ServiceName_ to receive the messages that are being published.
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+The iceoryx2 _ServiceName_ that is used to transfer a uProtocol message is derived from the message's `source` and `sink` attributes.
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-service-name~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+The ServiceName of an iceoryx2 message containing a _Publish_ UMessage **MUST** consist of the following segments:
```
sophokles73, line 107:
```suggestion
The ServiceName for sending a _Publish_ UMessage **MUST** consist of the following segments:
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+`up/[source.authority]/[source.ue_type]/[source.ue_instance]/[source.ue_version]/[source.resource]`
+
+The ServiceName of an iceoryx2 message containing a _Notification_ or _RPC Response_ UMessage **MUST** consist of the following segments:
```
sophokles73, line 111:
```suggestion
The ServiceName for sending a _Notification_ or _RPC Response_ UMessage **MUST** consist of the following segments:
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+`up/[source.authority]/[source.ue_type]/[source.ue_instance]/[source.ue_version]/[source.resource]/[sink.authority]/[sink.ue_type]/[sink.ue_instance]/[sink.ue_version]/[sink.resource]`
+
+The ServiceName of an iceoryx2 message containing an _RPC Request_ UMessage **MUST** consist of the following segments:
```
sophokles73, line 115:
```suggestion
The ServiceName for sending an _RPC Request_ UMessage **MUST** consist of the following segments:
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,264 @@
...
+a| The segment *MUST* contain the (UTF8) string representation of the 
+
+1. name of the host/authority that the (local) uEntity is running on, if authority name is empty.
+2. authority name, otherwise.
+
+|`ue_type`
+|`ue_id`
+a| The segment *MUST* contain the (UTF8) string representation of the
+
+the upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the uEntity type identifier with all leading `0` characters omitted.
```
sophokles73, line 144:
```suggestion
a| The segment *MUST* contain the (UTF8) string representation of the
upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the uEntity type identifier with all leading `0` characters omitted.
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+|`ue_instance`
+|`ue_id`
+a| The segment *MUST* contain the (UTF8) string representation of the
+
+the upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the uEntity instance identifier with all leading `0` characters omitted.
```
sophokles73, line 150:
```suggestion
a| The segment *MUST* contain the (UTF8) string representation of the
upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the uEntity instance identifier with all leading `0` characters omitted.
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,264 @@
...
+|`ue_version_major`
+a| The segment *MUST* contain the (UTF8) string representation of the
+
+the upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the uEntity major version with all leading `0` characters omitted.
+
+|`resource`
+|`resource_id`
+a| The segment *MUST* contain the (UTF8) string representation of the
+
+the upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the resource identifier with all leading `0` characters omitted.
```
sophokles73, line 162:
```suggestion
a| The segment *MUST* contain the (UTF8) string representation of the
upper-case link:https://www.rfc-editor.org/rfc/rfc4648#section-8[base16 encoding] of the resource identifier with all leading `0` characters omitted.
```

up-l1/iceoryx2.adoc
PLeVasseur: Same comment as I left above.

up-l1/iceoryx2.adoc
PLeVasseur: Modified to hold serialized UAttributes for simplicity for now.

up-l1/iceoryx2.adoc
PLeVasseur: Tried to find and replace all to _ServiceName_

up-l1/iceoryx2.adoc
PLeVasseur: Let's track this in #291

up-l1/iceoryx2.adoc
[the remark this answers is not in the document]
PLeVasseur: Think we've aligned here.

up-l1/iceoryx2.adoc
sophokles73: There is no place in the uProtocol specification where we use the term _uProtocol Schema Version_, if I am not mistaken. Is the idea here that a receiver that has been built with uProtocol major version 2 is still able to process a message that has been created by sender based on uProtocol major version 1?

PLeVasseur: Ah, I'm misremembering the Zenoh UTransport spec. You're right!

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,268 @@
...
+}
+----
+
+.uAttributes Mapping to UProtocolHeader, iceoryx2 User Header
+[%autowidth]
+|===
+| field | value
+
+| `uprotocol_major_version`
+a| *MUST* be set to a single byte representing the UProtocol major version (`0x01` at the time of writing)
```
PLeVasseur, line 74:
```suggestion
a| *MUST* be set to a single byte representing the uProtocol major version (`0x01` at the time of writing)
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,268 @@
...
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-zero-copy~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+An iceoryx2 message that is used to convey a uProtocol message *MUST* be zero-copy and shared-memory compatible:
+
+* is self contained, no heap, no pointers to external sources, no handles to external resources (files etc)
+* has a uniform memory representation -> `#[reprpass:[(C)]]`
+* does not use pointers to manage their internal structure
+* and its members don't implement `Drop` explicitly
+* has a `'static` lifetime
```
sophokles73, line 102: these are all criteria that are defined based on Rust concepts, right? In the next section you then refer to the corresponding Trait _if implemented in Rust_. To me it seems that the only way to implement this transport is indeed using Rust, or am I mistaken? If so, then we could simply refer to the Trait and leave out the specific requirements. That way, we would be less prone to running into inconsistencies when iceoryx2 evolves ...

PLeVasseur: Hmm... you're right that this is mixed up a bit.

It is possible to have a C or C++ payload meet the criteria as well, as we can see in these [examples](https://github.com/eclipse-iceoryx/iceoryx2/tree/main/examples/c/publish_subscribe_cross_language/src).

Let me try to more clearly separate this out.

PLeVasseur: Could you take a look at [e220df9](https://github.com/eclipse-uprotocol/up-spec/pull/285/commits/e220df9116c090f128726e12d4042656abae4934) to see if this helps clear up the separation between general requirements on the payload and Rust-specific?

up-l1/iceoryx2.adoc
```diff
@@ -96,12 +96,20 @@ An iceoryx2 message that is used to convey a uProtocol message *MUST* contain in
...
-* has a uniform memory representation -> `#[reprpass:[(C)]]`
+* has a uniform memory representation
 * does not use pointers to manage their internal structure
-* and its members don't implement `Drop` explicitly
-* has a `'static` lifetime
+* has a lifetime of the entire program duration
+--
 
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-zero-copy-rust~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
```
sophokles73, line 105: We can leave it like that but I would suggest to remove the specitem declaration and just leave it as a hint. Otherwise I won't be able to cover that specitem if I implement the transport in C++, right?

PLeVasseur: Ah... leave it as a hint => no annotation needed for OFT at all then? That's doable, just confirming.

PLeVasseur: Updated in [8eb51ed](https://github.com/eclipse-uprotocol/up-spec/pull/285/commits/8eb51ed633b7cc6a9fc3cf95f6671eb31c7bdec1). Should this work?

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,273 @@
...
+
+==== Message Priority
+
+Currently not a clear path to map from uProtocol message priority to iceoryx2 QoS (Quality of Service). Skip for current revision.
+
+=== Payload Mapping
+
+[.specitem,oft-sid="dsn~up-transport-iceoryx2-payload-mapping~1",oft-needs="impl,utest",oft-tags="TransportLayerImpl"]
+--
+An iceoryx2 message that is used to convey a uProtocol message *MUST* contain in its payload the unaltered value of the UMessage's _payload_ field.
```
sophokles73, line 91:
```suggestion
An iceoryx2 `Sample` that is used to convey a uProtocol message *MUST* contain in its payload the unaltered value of the UMessage's _payload_ field.
```

up-l1/iceoryx2.adoc
```diff
@@ -0,0 +1,273 @@
...
+* has a lifetime of the entire program duration
+--
+
+The cleanest way to describe this is to say that the payload definition **MUST** have the link:https://docs.rs/iceoryx2/0.6.1/iceoryx2/prelude/trait.ZeroCopySend.html[`ZeroCopySend`] trait implemented for it if written in Rust.
+
+That encapsulates the following additional properties in Rust:
+
+* Rust `enum` and `struct` are marked  as `#[reprpass:[(C)]]`
+* members don't implement `Drop` explicitly
+* has a `'static` lifetime
```
sophokles73, line 110: I wonder if this is actually needed at all, given that this is already enforced by iceoryx2 itself ...

PLeVasseur: Okay, let's keep this wording for now.

PR #287, opened 2025-06-11
Cover UTransport requirements by default

sophokles73:
The Zenoh and the MQTT 5 transport spec have been amended to cover
some generic requirements from the Transport Layer specification
by default. This relieves implementations from creating corresponding
spec items which do contain anything specific to the implementation.


evshary (approved):
LGTM.
BTW, what does CSn (like CS6) here mean? Class Service?

PR #293, opened 2025-07-07
Fix covered OFT specitem ID

sophokles73:
The Gherkin scenarios for verifying UUID serialization to its
hex-and-dash string representation have been adapted to refer to the
specific OFT specitem from the UUID specification document.


AnotherDaniel (approved):
LGTM

PR #294, opened 2025-07-07
Adapt address spoofing prevention requirements

### LINKED ISSUE #292: Transport Layer spec's address spoofing requirement change
The [Transport Layer spec](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#31-send) currently requires an implementation to fail the sending of messages with a source URI that differs from the uEntity's associated local URI.
This is meant to prevent uEntities from spoofing their (logical) address, thus impersonating other components. The only component that should be allowed to do this, are uStreamers.

While trying to implement test cases for this behavior for the Zenoh transport, I ran into a problem: The `zenoh::Session::put` (publish) function for publishing messages into the Zenoh network _always_ succeeds, regardless of the specific ACLs defined for any of the nodes defined in the network. Based on the ACLS, nodes will _silently drop_ any messages that the ACLs filter out. This means that it is technically impossible for the Zenoh transport to fulfill this requirement.

What we need to prevent, of course, is the (unwanted) address spoofing to happen.
However, I wonder if it is actually imperative for a client that maliciously tries to spoof its address, to get this kind of explicit feedback (the error with code `UCode::PERMISSION_DENIED`).

FMPOV it would be sufficient to require such messages to **not** be processed and leave the way of ensuring this up to the implementer. In case of the MQTT5 transport it should be possible to return a corresponding error. For the Zenoh transport this would mean that such messages will be dropped silently and thus not reach their destination. Consequently, RPC calls would run into a timeout.

I would like to change the spec accordingly, to allow the Zenoh transport to fully cover all spec items.

WDYT @PLeVasseur @evshary @stevenhartley @AnotherDaniel @ValMobBIllich

sophokles73:
The Transport Layer specification's requirements regarding prevention of
address spoofing have been revised to focus on defining the undesired
outcome only. This should allow the Zenoh transport to also meet these
requirements.

Fixes #292

### DIFF
--- a/up-l1/README.adoc
+++ b/up-l1/README.adoc
@@ -147,11 +147,15 @@ UTransport implementations
 * *MUST* fail invocations of <<send>> with a `UCode.INVALID_ARGUMENT`, if the passed UMessage failed validation.
 --
 
-[.specitem,oft-sid="req~utransport-send-error-permission-denied~2",oft-needs="dsn,uman",oft-tags="TransportLayerImpl"]
+[.specitem,oft-sid="req~utransport-send-prevent-address-spoofing~1",oft-needs="dsn,uman",oft-tags="TransportLayerImpl"]
 --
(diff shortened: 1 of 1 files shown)


evshary (approved):
LGTM

PR #295, opened 2025-07-10
Fix copy-paste duplication of req id

AnotherDaniel:
In usubscription spec, there was a copy-paste induced duplication of a requirement id - this PR fixes that.


sophokles73 (approved):
LGTM

PR #297, opened 2025-07-15
Fix links to proto3 files

sophokles73:
Also replaced explicit links with AsciiDoc document attributes to
improve maintainability.


AnotherDaniel (approved):
Looks good - asciicode is way too complex for it's own good these days ;)

Issue #312, opened 2025-11-05
up-spec should not be a Java Maven project


sophokles73:
Many of the other uProtocol repositories include up-spec as a Git submodule in order to be able to trace requirements to code using OpenFastTrace.

The pom.xml file in the up-spec root folder is recognized by IDEs like VSCode and triggers the startup of Java tooling related plugins. This is neither necessary nor desirable because the uProtocol language libraries already provide the types created from the proto files.

Currently, the main purpose of the Java based build is to make sure that updated proto3 files are still valid and consistent with each other. This, however, could also be achieved by simply running `protoc` from within a CI job.

stevenhartley:
Hi @sophokles73 when we started there was only the up-java project and then we forked off to create the spec project so I kept the maven. Since vscode has all the plugins, it was easy to load and then run without installing anything on the dev system. since you  now setup git submodules, we don't need to publish this to maven central anymore.

sophokles73:
I have created https://github.com/eclipse-uprotocol/up-spec/pull/313

@stevenhartley would you mind taking a look?

PR #313, opened 2025-11-05
Do not use Maven for validation

sophokles73:
Removed all Maven-related files and references and changed "Verify PR" workflow to use plain protoc for validating updated proto files instead.

This addresses #312


stevenhartley (approved):
I used maven for historical reasons but they are not valid anymore. I'm OK to clean this up.

PR #318, opened 2026-01-02
Do not require priority for publish messages

### LINKED ISSUE #315: Publish messages should not be required to have priority level set explicitly
The [UAttributes specification](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc#4-publish-message-attributes) currently requires the *priority* as mandatory on Publish messages. However, I cannot see any compelling reason for doing so, or am I mistaken?

FMPOV the property should be defined as optional, which would then also match the current implementation of the Rust and Java UAttributesValidators.

@stevenhartley @PLeVasseur @AnotherDaniel WDYT?

sophokles73:
Updated UAttributes specification to make the 'priority' attribute
optional for publish messages. This change can help reduce the size of
encoded publish messages by allowing producers to omit the priority
setting when not necessary.

Fixes #315

### DIFF
--- a/basics/upriority.adoc
+++ b/basics/upriority.adoc
@@ -21,7 +21,7 @@ SPDX-License-Identifier: Apache-2.0
 
 uProtocol allows messages to be exchanged between uEntities on the same and/or different devices. The network bandwidth and/or CPU cycles available for performing this task may be limited, though. uProtocol therefore supports priority based message processing at the xref:../up-l1/README.adoc[Transport] and xref:../up-l2/README.adoc[Communication Layer] based on https://datatracker.ietf.org/doc/html/rfc4594[RFC4594 - Configuration Guidelines for DiffServ Service Classes].
 
-For this purpose, each xref:umessage.adoc[UMessage] is assigned to exactly one of the following classes:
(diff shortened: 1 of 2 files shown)


basics/upriority.adoc
```diff
@@ -21,21 +21,45 @@ SPDX-License-Identifier: Apache-2.0
 
 uProtocol allows messages to be exchanged between uEntities on the same and/or different devices. The network bandwidth and/or CPU cycles available for performing this task may be limited, though. uProtocol therefore supports priority based message processing at the xref:../up-l1/README.adoc[Transport] and xref:../up-l2/README.adoc[Communication Layer] based on https://datatracker.ietf.org/doc/html/rfc4594[RFC4594 - Configuration Guidelines for DiffServ Service Classes].
 
-For this purpose, each xref:umessage.adoc[UMessage] is assigned to one of the following service classes:
+For this purpose, each xref:umessage.adoc[UMessage] is assigned to exactly one of the following classes:
```
PLeVasseur: Slight wording change for clarity

```suggestion
For this purpose, each xref:umessage.adoc[UMessage] is assigned exactly one of the following classes:
```

PR #331, opened 2026-07-20
feat: open payload-encoding identity on UAttributes

PLeVasseur:
Adds three optional fields to UAttributes for payload encodings that have no UPayloadFormat equivalent:

  optional uint32 payload_encoding_registry_id = 13;
  optional string payload_encoding                      = 14;
  optional string payload_content_type               = 15;

Motivation: UPayloadFormat is a closed enum. Every new payload encoding (a CDR profile, an Arrow layout, a custom vehicle-domain codec) currently requires editing the enum in the core proto, which churns the wire contract and every implementation. These fields carry an open encoding identity instead: a registered numeric id, a literal id, or a media type, individually or together.

Contract (normative text in basics/uattributes.adoc):
- payload-free messages carry none of these;
- a payload-bearing message populates exactly one encoding mechanism, legacy payloadFormat or the open identity, never both;
- UPayloadFormat values 1..=8 are permanently reserved as the legacy-compatible numeric range and are never re-registered as open identities, so the two mechanisms can never alias;
- a receiver that does not understand an open identity rejects the payload rather than reinterpreting it as raw bytes.

The change is wire-compatible: fields are optional, field numbers 13-15 were unused, existing messages are unaffected, and the enum is untouched. A follow-up specification for the frame-native metadata model builds on these fields; they stand alone as an extensibility mechanism regardless.

### DIFF
--- a/up-core-api/uprotocol/v1/uattributes.proto
+++ b/up-core-api/uprotocol/v1/uattributes.proto
@@ -63,7 +63,24 @@ message UAttributes {
     optional string traceparent = 11;
 
     // The format for the data stored in the UMessage.
+    //
+    // Covers the legacy-compatible numeric range of the UFrame payload-encoding
+    // registry. Payload encodings outside that range are carried by the open
+    // payload-encoding identity fields below. A payload-bearing message MUST
+    // populate exactly one encoding mechanism.
     UPayloadFormat payload_format = 12;
+
+    // Optional numeric registry id from the UFrame payload-encoding registry
+    // for encodings that have no UPayloadFormat equivalent.
+    optional uint32 payload_encoding_registry_id = 13;
+
+    // Optional literal id from the UFrame payload-encoding registry for
+    // encodings that have no UPayloadFormat equivalent.
+    optional string payload_encoding = 14;
+
+    // Optional media type from the UFrame payload-encoding registry for
+    // encodings that have no UPayloadFormat equivalent.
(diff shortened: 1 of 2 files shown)


up-core-api/uprotocol/v1/uattributes.proto
```diff
@@ -63,7 +63,24 @@ message UAttributes {
...
+    // for encodings that have no UPayloadFormat equivalent.
+    optional uint32 payload_encoding_registry_id = 13;
+
+    // Optional literal id from the UFrame payload-encoding registry for
+    // encodings that have no UPayloadFormat equivalent.
+    optional string payload_encoding = 14;
+
+    // Optional media type from the UFrame payload-encoding registry for
+    // encodings that have no UPayloadFormat equivalent.
+    optional string payload_content_type = 15;
```
sophokles73, line 83: Why do we need these if they are synonymous with the ID from the _UFrame Registry_? Why would a uEntity know these values but could not cope with the ID instead?

sophokles73 (commented):
I do not really like the idea of introducing yet another mechanism for specifying payload format. If the existing mechanism is not suitable anymore, then we should replace it with something that is expressive enough to cover our use cases. 
Adding more optional fields makes it harder and harder to determine what the payload actually is. I would rather not make that logic more complicated than it already is.
FMPOV there is no need for backwards compatibility yet, because we are still producing alpha releases ...

Issue #328, opened 2026-06-22
Make some UMessage attributes mandatory


sophokles73:
In up-rust we have recently introduced custom definitions of the uProtocol types and no longer directly use the protoc generated types in our API. I am currently reviewing the existing API of UMessage/UAttributes and have stumbled across a few functions that I consider obsolete and/or confusing.

The [UAttributes spec](https://github.com/eclipse-uprotocol/up-spec/blob/v1.6.0-alpha.7/basics/uattributes.adoc) currently defines several attributes as optional instead of explicitly defining a default value. My understanding is, that this basically comes from the fact that we originally used proto3 files to define these types and protobuf tries to be as efficient as possible by means of omitting fields that have their default value, e.g. uint32 typed fields having value 0.

This is also reflected in our custom Rust types, for example the _time-to-live_ attribute is defined as `Option<u32>` instead of just `u32`. Client code retrieving the attribute is left alone with interpreting a value of `None`, though, because the spec is unclear about the meaning of a message that has no TTL defined.

My best guess is that by default, such a message should be considered to have a TTL of 0, which means that it never expires. An alternative interpretation might be that a processor can assume any reasonable TTL (e.g. 5 seconds) instead. However, this behavior is not defined explicitly.

To make the API more concise and to remove ambiguity, I suggest making the _ttl_ field mandatory and define 0 as its default value. Similarly, I suggest making _priority_ mandatory with a default value of `CS1`, and to make the _commstatus_ property of Response messages mandatory with a default value of `OK`.

Note that this has no direct impact on how the message is being encoded on the wire because the transports already are free to define the mapping of message attributes to their particular packet data units (PDU).

sophokles73:
@PLeVasseur @AnotherDaniel WDYT?

AnotherDaniel:
Agree with the sentiment; I am indifferent about making things mandatory, but the minimum thing to do is clearly define the defaults. And code tends to be more concise if you don't have to check for the presence of optional values all the time, so I guess your suggestion the make them mandatory is the way to go.

PLeVasseur:
I like this idea. Agreed that this seems to be an artifact of the choice of proto3 rather than an intentional choice that was made.

> To make the API more concise and to remove ambiguity, I suggest making the ttl field mandatory and define 0 as its default value.

Makes sense to me.

> Similarly, I suggest making priority mandatory with a default value of CS1, 

[...]

PR #334, opened 2026-08-13
[#328] Make `commstatus` mandatory for RPC Response messages

sophokles73:
* Updated the error model spec to make the `code` property mandatory for a UStatus.
* Updated the UAttributes spec to make the `commstatus` property mandatory for RPC Response messages.
* Adapted the proto3 files accordingly.

### DIFF
--- a/basics/error_model.adoc
+++ b/basics/error_model.adoc
@@ -103,7 +103,7 @@ A service consumer invoking this operation will then receive a `UMessage` with a
 
 == Service Provider Implementation
 
-[.specitem,oft-sid="dsn~service-error-model-success-response~1",oft-needs="impl,utest",oft-tags="ServiceProvider"]
+[.specitem,oft-sid="dsn~service-error-model-success-response~2",oft-needs="impl,utest",oft-tags="ServiceProvider"]
 --
 In case of _successful_ execution of a service operation, a service provider
 
--- a/basics/error_model.adoc
+++ b/basics/error_model.adoc
@@ -41,36 +41,36 @@ classDiagram
 
 class UStatus {
   message: String[0..1]
-  details: Object[*]
+  details: Object[0..*]
 }
 
 class UCode {
   <<Enumeration>>
-  OK
-  CANCELLED
-  UNKNOWN
-  INVALID_ARGUMENT
-  DEADLINE_EXCEEDED
-  NOT_FOUND
-  ALREADY_EXISTS
-  PERMISSION_DENIED
-  UNAUTHENTICATED
-  RESOURCE_EXHAUSTED
-  FAILED_PRECONDITION
-  ABORTED
-  OUT_OF_RANGE
-  UNIMPLEMENTED
-  INTERNAL
-  UNAVAILABLE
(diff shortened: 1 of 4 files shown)


basics/uattributes.adoc
```diff
@@ -420,8 +420,8 @@ The amount of time after which this response message should no longer be deliver
 
 |`commstatus`
 |xref:error_model.adoc#data-model-definition[UCode]
-|no
-a|The outcome of the RPC. No value or `UCode.OK` indicate successful invocation of the operation. All other values indicate that invocation was unsuccessful. The concrete value indicates the type of error that has occurred.
+|yes
+a|The outcome of the RPC. `UCode.Ok` indicates successful invocation of the operation. All other values indicate that invocation was unsuccessful. The concrete value indicates the type of error that has occurred. The message's payload may contain additional information about the error.
```
PLeVasseur, line 424: Do we have an OFT specitem here? Should we? I noticed elsewhere in `basics/error_model.adoc`, we changed the text and bumped the version.

basics/error_model.adoc
```diff
@@ -108,21 +108,21 @@ A service consumer invoking this operation will then receive a `UMessage` with a
...
-* *MAY* set the response message's `commstatus` attribute to value `UCode.OK` to indicate successful execution. If the `commstatus` attribute is not set, it is assumed to be `UCode.OK`.
+* *MUST* set the response message's `commstatus` attribute to value `UCode.Ok` to indicate successful execution.
 --
 
 [.specitem,oft-sid="dsn~service-error-model-error-response~1",oft-needs="impl,utest",oft-tags="ServiceProvider"]
 --
 In case of _erroneous_ execution, a service provider
 
-* *MUST* set the `commstatus` attribute of the response xref:umessage.adoc[`UMessage`] to the <<ustatus-data-model,UCode>> value other than `UCode.OK` which most closely matches the reason for failure.
+* *MUST* set the `commstatus` attribute of the response xref:umessage.adoc[`UMessage`] to the <<ustatus-data-model,UCode>> value other than `UCode.Ok` which most closely matches the reason for failure.
```
PLeVasseur, line 118: Is this worth bumping the version of the specitem? Maybe not?

PLeVasseur (approved):
Looks good. I'm poking around to ask questions more than asking changes with the comments I left.

basics/uattributes.adoc
sophokles73: Hmm, good question. I would say that the underlying data model has not changed with regard to commstatus cardinality, i.e. UAttributes.commstatus is still optional, because only response messages carry it.

However, for response messages, the cardinality has changed from `0..1` to `1`. IMHO we cover that requirement in the error_model spec. And I agree that, because of the semantic change, we should bump the specitem revision there ...

basics/error_model.adoc
sophokles73: No, there is no semantic change introduced here.

sophokles73:
@PLeVasseur I have pushed a change for the error_model spec

PR #336, opened 2026-08-22
feat!: define unified payload encoding identity

PLeVasseur:
This replaces `UPayloadFormat` with one optional numeric `payload_encoding_id`, giving payload encodings an extensible, language-neutral identity. It also aligns the CloudEvents, MQTT 5, and SOME/IP mappings, and corrects the related uDiscovery documentation.

### DIFF
--- a/basics/uattributes.adoc
+++ b/basics/uattributes.adoc
@@ -47,7 +47,7 @@ class UAttributes {
   requestId: UUID [0..1]
   token: String [0..1]
   traceparent: String [0..1]
-  payloadEncodingId: UInt32 [0..1]
+  payloadEncodingId: UInt16 [0..1]
 }
 
 class UMessageType {
@@ -173,26 +173,33 @@ The traceparent *MUST* either be empty or contain a value as defined by https://
 
 
 |`payloadEncodingId`
-|`UInt32`
+|`UInt16`
 |no
 |none
-a|The encoding of the message payload, represented by an entry identifier in the
-xref:#payload-encoding-registry[payload-encoding registry].
+a|The encoding of the message payload, represented by an identifier defined in
+the xref:#payload-encoding-registry[payload-encoding registry].
 
 [.specitem,oft-sid="dsn~up-attributes-payload-encoding-id~1",oft-needs="impl,utest",oft-tags="LanguageLibrary"]
 --
-The property
+The property and the UMessage's `payload` field *MUST* either both be present or
+both be absent. A present zero-length `payload` field is a present payload.
 
-* *MUST NOT* contain a value if the message has no payload;
-* *MUST NOT* contain the reserved identifier `0`; and
-* if the message has a payload, *MUST* contain the registry entry identifier of
-  the payload's encoding unless the applicable transport binding defines the
-  encoding out of band.
+If present, the property
 
-If a message has a payload and does not contain this property, the applicable
-transport binding *MUST* define the payload encoding. A receiver *MUST* reject a
-payload if its encoding cannot be determined or if the receiver does not
-recognize the declared identifier.
+* *MUST* contain a value in the range `0x0000` to `0xDFFF` or `0xF000` to
+  `0xFFFF`; and
+* *MUST NOT* contain a value greater than `0xFFFF` or in the reserved range
+  `0xE000` to `0xEFFF`.
+
+Identifier `0x0000` indicates that a binding, service, or schema contract
+defines the concrete payload encoding. A component that needs to interpret such
+a payload *MUST* have that contract and *MUST NOT* infer the encoding from the
+payload data.
+
+Structural validation *MUST NOT* reject an otherwise valid identifier in the
+registry-assignment or private-use range merely because the component does not
+recognize or support it. A component that needs to interpret a payload *MUST*
+reject the payload if its encoding cannot be determined or is not supported.
 --
 
 |===
@@ -464,34 +471,83 @@ a|The outcome of the RPC. `UCode.Ok` indicates successful invocation of the oper
 [#payload-encoding-registry]
 == Payload-Encoding Registry
 
-The payload-encoding registry assigns an unsigned 32-bit identifier to each
-payload encoding. A `payloadEncodingId` value is an entry identifier from this
-registry.
+The payload-encoding registry defines an unsigned 16-bit identifier space for
+payload encoding identity. The identifier space is partitioned as follows:
 
-[.specitem,oft-sid="dsn~payload-encoding-registry~1",oft-needs="impl,utest",oft-tags="LanguageLibrary"]
---
-* Identifier `0` is reserved and *MUST NOT* be used on the wire.
-* Identifiers `1`-`11` are assigned and *MUST NOT* be reassigned.
-* Identifiers `12` to `0x0FFF_FFFF` are available for registry assignments.
-* Identifiers `0x1000_0000` and above are private use: they are valid on the
-  wire, are never registered, and carry meaning only by deployment-level
-  agreement.
---
+* `0x0000` denotes an implicit encoding defined by an applicable binding,
+  service, or schema contract.
+* `0x0001` to `0xDFFF` are available for registry assignments.
+* `0xE000` to `0xEFFF` are reserved for future use and invalid on the wire.
+* `0xF000` to `0xFFFF` are private use. They are valid on the wire, are never
+  registered, and carry meaning only by deployment-level agreement.
+
+This is an independent uProtocol registry. Its identifiers do not reuse the
+numeric values of the IANA CoAP Content-Formats registry.
+
+A registry assignment is permanent and *MUST NOT* be reassigned. New
+assignments are made by changing the table below through the normal up-spec
+review process; changing the identifier-space partition is not required. Each
+assignment *MUST* provide a stable name, a precise description, a formal public
+specification, and a registered media type when one accurately identifies the
+encoding. Producers *MUST NOT* use an unassigned identifier.
+
+An identifier in the registry-assignment range remains structurally valid when
+a receiver's local copy of the registry does not contain it. This permits
+intermediaries and older implementations to preserve identifiers assigned by a
+newer specification without claiming support for decoding their payloads.
 
 .Payload-encoding registry
-[cols="1,3,3", options="header"]
+[cols="1,2,5,3,5", options="header"]
 |===
-|ID |Encoding |Media type
+|ID |Name |Description |Media type |Specification
 
-|1 |Protobuf, wrapped in `google.protobuf.Any` |`application/x-protobuf`
-|2 |Protobuf |`application/protobuf`
-|3 |JSON |`application/json`
-|4 |SOME/IP |`application/x-someip`
-|5 |SOME/IP TLV |`application/x-someip_tlv`
-|6 |Raw bytes |`application/octet-stream`
-|7 |UTF-8 text |`text/plain`
-|8 |Shared-memory reference |`application/x-shm`
-|9 |XCDR v2 (DDS serialization) |`application/vnd.uprotocol.xcdr-v2`
-|10 |Arrow IPC |`application/vnd.apache.arrow.stream`
-|11 |OMG IDL CDR |`application/vnd.uprotocol.omg-idl-cdr`
+|1
+|Protobuf Any
+|The Protobuf wire-format serialization of a `google.protobuf.Any`, including
+its type URL and serialized embedded message.
+|https://www.iana.org/assignments/media-types/application/protobuf[`application/protobuf`]
+|https://protobuf.dev/programming-guides/encoding/[Protobuf wire format] and
+https://protobuf.dev/reference/protobuf/google.protobuf/#any[`google.protobuf.Any`]
+
+|2
+|Protobuf
+|The Protobuf wire-format serialization of a message whose type is defined by
+the applicable service or schema contract.
+|https://www.iana.org/assignments/media-types/application/protobuf[`application/protobuf`]
+|https://protobuf.dev/programming-guides/encoding/[Protobuf wire format]
+
+|3
+|JSON
+|A JSON text encoded as UTF-8.
+|https://www.iana.org/assignments/media-types/application/json[`application/json`]
+|https://www.rfc-editor.org/rfc/rfc8259[RFC 8259]
+
+|4
+|SOME/IP
+|The SOME/IP serialization of parameters and data structures defined by the
+applicable service-interface contract.
+|None registered
+|https://some-ip.com/open-someip-specification/open_someip_specification_2025-12.pdf[Open SOME/IP Specification 2025-12, Section 5.4]
+
+|5
+|SOME/IP TLV
+|The SOME/IP serialization of parameters and data structures using the optional
+TLV encoding defined by the applicable service-interface contract.
+|None registered
+|https://www.autosar.org/fileadmin/standards/R25-11/FO/AUTOSAR_FO_PRS_SOMEIPProtocol.pdf[AUTOSAR SOME/IP Protocol Specification R25-11, Section 5.1.3.3]
+
+|6
+|Raw bytes
+|An arbitrary sequence of octets with semantics defined by the applicable
+service or schema contract.
+|https://www.iana.org/assignments/media-types/application/octet-stream[`application/octet-stream`]
+|https://www.rfc-editor.org/rfc/rfc2046#section-4.5.1[RFC 2046, Section 4.5.1]
+
+|7
+|UTF-8 text
+|A linear sequence of Unicode characters encoded as UTF-8, without content
+markup or format-specific interpretation.
+|https://www.iana.org/assignments/media-types/text/plain[`text/plain; charset=utf-8`]
+|https://www.rfc-editor.org/rfc/rfc3629[RFC 3629] and
+https://www.rfc-editor.org/rfc/rfc2046#section-4.1.3[RFC 2046, Section 4.1.3]
 |===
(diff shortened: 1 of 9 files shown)


basics/uattributes.adoc
```diff
@@ -47,6 +47,7 @@ class UAttributes {
   requestId: UUID [0..1]
   token: String [0..1]
   traceparent: String [0..1]
+  payloadEncodingId: UInt32 [0..1]
```
sophokles73: Do we really need 32 bits for this identifier? 16 bits would allow for >65000 encodings already. I also don't think that we need such a generous _private_ range, something like [0xF000, 0xFFFF] should do, shouldn't it? That's already >4000 IDs that you can use for internal stuff ...

basics/uattributes.adoc
```diff
@@ -176,13 +163,26 @@ A tracing identifier to use for correlating messages across the system. Intended
...
+a|The encoding of the message payload, represented by an entry identifier in the
+xref:#payload-encoding-registry[payload-encoding registry].
+
+[.specitem,oft-sid="dsn~up-attributes-payload-encoding-id~1",oft-needs="impl,utest",oft-tags="LanguageLibrary"]
 --
-* *MUST NOT* contain a value if the message has no payload.
+The property
+
+* *MUST NOT* contain a value if the message has no payload;
+* *MUST NOT* contain the reserved identifier `0`; and
```
sophokles73: I guess this is so that we can use it for protobuf's UNSPECIFIED value?

basics/uattributes.adoc
```diff
@@ -176,13 +163,26 @@ A tracing identifier to use for correlating messages across the system. Intended
...
+* if the message has a payload, *MUST* contain the registry entry identifier of
```
sophokles73: can we make this as simple as:
if it has a payload, it must have an encoding ID and vice versa

We could then define a dedicated ID for _out-of-band_ or _implicit_. WDYT?

FMPOV that would make it much easier for implementations to correctly handle the payload and validate message integrity

basics/uattributes.adoc
```diff
@@ -425,49 +425,37 @@ a|The outcome of the RPC. No value or `UCode.OK` indicate successful invocation
...
 
+[.specitem,oft-sid="dsn~payload-encoding-registry~1",oft-needs="impl,utest",oft-tags="LanguageLibrary"]
+--
+* Identifier `0` is reserved and *MUST NOT* be used on the wire.
+* Identifiers `1`-`11` are assigned and *MUST NOT* be reassigned.
+* Identifiers `12` to `0x0FFF_FFFF` are available for registry assignments.
+* Identifiers `0x1000_0000` and above are private use: they are valid on the
+  wire, are never registered, and carry meaning only by deployment-level
+  agreement.
+--
```
sophokles73: I do not quite see how a language library would _impl_ and _utest_ this ...

FMPOV we should also not explicitly define the 1-11 range as already taken. Instead, we should simply define [0x0001, 0xDFFF] as _registry assignments_. The table below then simply needs to be updated with newly registered IDs without the need to alter this spec item (and increase its revision) ...

basics/uattributes.adoc
```diff
@@ -425,49 +425,37 @@ a|The outcome of the RPC. No value or `UCode.OK` indicate successful invocation
...
-|`text/plain`
-|The network byte order UTF-8 encoding of a unicode string representing the payload data.
-
-|`UPAYLOAD_FORMAT_SHM`
-|`application/x-shm`
-|The network byte order address of a shared memory segment.
+.Payload-encoding registry
+[cols="1,3,3", options="header"]
+|===
+|ID |Encoding |Media type
```
sophokles73: FMPOV we should keep the description column. We might also want to mandate the existence of a _formal_ definition of the encoding that we can link to ...

sophokles73 (changes requested):

basics/uattributes.adoc
PLeVasseur: Probably don't need 32 bits, no. We can lower this to 16 bits.

sophokles73: We might even get away with simply using the already existing CoAP Content-Formats registry maintained by IANA: https://www.iana.org/assignments/core-parameters#content-formats
However, thinking about it a little more, the CoAP people probably won't like us to add content-formats that do not play any role in the CoAP domain, like SOMEIP or protobuf formats ...

PLeVasseur: This is now implemented in the current revision. `payloadEncodingId` is a conceptual `UInt16`; protobuf field 12 remains `optional uint32` because protobuf has no `uint16` scalar, and values above `0xFFFF` are invalid. Under the current independent-registry model, `0x0001`-`0xDFFF` is the assignment range, `0xE000`-`0xEFFF` is reserved and invalid on the wire, and `0xF000`-`0xFFFF` is private use.

I also took a deeper look at reusing the IANA CoAP Content-Formats registry. The strong benefit is that uProtocol would not need to allocate and permanently curate another numeric registry. A CoAP Content-Format already identifies a media type together with valid media-type parameters and an optional content coding, and existing values cover UTF-8 text (`0`), arbitrary bytes (`42`), and JSON (`50`).

The downside is that strict reuse does not cover every uProtocol case we may plan for. For example, Protobuf and Apache Arrow have registered media types but no assigned CoAP numbers; they might be registration candidates, but we would need to confirm with the CoRE designated experts that registrations motivated by uProtocol are appropriate.

SOME/IP, XCDR2, and GIOP/CDR would first need suitable media-type registrations, while deployment-specific native or shared-memory ABI identities are not really Internet media types. CoAP also assigns `0` to text, and its `65000`-`65535` range is for experiments only, not operational private use. Adopting it would therefore require absent format metadata to mean "defined by the service or schema contract" and would require native ABI/type identity to remain separate from payload representation identity.

A broader pattern is the RFC 9193 Content-Format-Spec, defined for SenML: carry a decimal CoAP number when one exists, or a content-format string such as `application/protobuf` or `application/vnd.apache.arrow.stream` otherwise. Adapting that model would avoid a uProtocol allocation registry and support registered media types, parameters, and content codings immediately, but it changes this proposal from a compact integer to a value that may carry a string and has broader API, binding, and native-frame consequences.

So CoAP reuse looks feasible, but it is not a drop-in substitution. Would you favor pausing this PR to investigate strict CoAP reuse or the RFC 9193-style hybrid, or do you think the scope and semantic mismatches are strong enough to justify an independent uProtocol registry and its ongoing curation cost?

basics/uattributes.adoc
PLeVasseur: That was the original reason for reserving `0`, but because this is an `optional` scalar, field presence already distinguishes no value from a present zero; we do not need an enum-style `UNSPECIFIED` value for protobuf.

In the current independent-registry revision, I therefore used `0` as the explicit implicit/out-of-band value. It is present only with a payload, the binding or service/schema contract supplies the concrete encoding, and a receiver without that contract must reject the payload. This depends on the registry decision discussed above: if we adopt CoAP values, `0` already means UTF-8 text, so contract-defined encoding would need to be represented by absence or another model rather than overloading CoAP identifier `0`.

basics/uattributes.adoc
PLeVasseur: Agreed. The current revision now has the exact presence invariant you proposed: a `UMessage` has a payload if and only if it has `payloadEncodingId`. Payload means presence of the optional payload field, including a present zero-length value. Under the current independent-registry model, ID `0` is the dedicated implicit/out-of-band value. The CloudEvents, MQTT 5, and SOME/IP mappings were updated to preserve or reconstruct that rule and reject mismatches.

This is tied to the CoAP decision above. Strict CoAP reuse cannot use `0` for implicit encoding and may instead need to allow a payload with an absent format when the service/schema contract defines it. If we choose that model, I would revise the presence rule and bindings together rather than leave an exception in only one layer.

basics/uattributes.adoc
PLeVasseur: Agreed. In the current revision, the registry allocation policy is no longer wrapped in a language-library `impl,utest` specification item. The testable wire-validity rules remain on `payloadEncodingId`, while the governance text defines `0x0001`-`0xDFFF` once as the assignment range. The table currently lists only assignments `1` through `7`; adding a future assignment changes the table without revising the range definition.

If we adopt the IANA CoAP registry, the local allocation policy and assignment table would be replaced rather than implemented by language libraries, but the separation between registry governance and testable message validity would still apply.

basics/uattributes.adoc
PLeVasseur: Agreed. The current revision restores the description column and adds a formal specification/reference column. Its policy requires each assignment to have a stable name, a precise description, a formal public specification, and a registered media type when one accurately identifies the encoding.

I also audited the entries instead of treating media-type labels as encoding definitions. The current table retains only IDs `1` through `7` with explicit descriptions and references. The proposed shared-memory, XCDR2, Arrow IPC, and GIOP/CDR entries were deferred rather than assigning numbers before their exact representations and authoritative definitions were settled. If we adopt CoAP, the IANA media-type and Content-Format records for assigned representations would become authoritative instead of a uProtocol-owned table; formats without an assignment would remain contract-defined.

PLeVasseur (commented):
Thanks for the review, @sophokles73.

I've left some thoughts; overall adopting CoAP looks trickier than I initially thought and seems to involve some degree of then coordinating with those maintaining that standard to see if we could add new entries.

I'm not entirely sure it's the wrong thing to do, but it certainly does add further delays and that gives me pause in pursuing that route.

AnotherDaniel:
The thing I'm wondering: if we make the payload encoding such a generic/prominent/extensible concept, does this maybe warrant a dedicated up-spec document that is the source of truth for the encodings uProtocol supports/declares identifiers for? 
Related question: currently the encoding list is defined in two (and a half) places, in the uattributes.adoc, plus two tables in cloudevents.adoc. That feels like maintenance accidents waiting to happen...

basics/uattributes.adoc
sophokles73: > or do you think the scope and semantic mismatches are strong enough to justify an independent uProtocol registry and its ongoing curation cost?

I share your concerns here and therefore believe, that we would be better off maintaining our own little registry. As @AnotherDaniel pointed out in a separate comment, I also believe that the registry would deserve its own document :-)

sophokles73:
> The thing I'm wondering: if we make the payload encoding such a generic/prominent/extensible concept, does this maybe warrant a dedicated up-spec document that is the source of truth for the encodings uProtocol supports/declares identifiers for? Related question: currently the encoding list is defined in two (and a half) places, in the uattributes.adoc, plus two tables in cloudevents.adoc. That feels like maintenance accidents waiting to happen...

I agree that we need a single source of truth of the payload encoding registry. I would also be in favor of creating a separate document for it that we can refer to.

The uProtocol Cloud Events spec's main concern is currently, which fields of the CloudEvent protobuf the different payload types should be mapped to. [The Cloud Events Protobuf Format spec](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/protobuf-format.md#3-data) mandates that particular types of payload _MUST_ be mapped to corresponding fields, i.e. text-like data must go into _text_data_, protobufs must go into _protobuf_data_ and binary data mut go into _binary_data_. So we would probably need to also include this information in the Payload Encoding Registry document, so that we have a clear indication of where to put the data.

basics/uattributes.adoc
sophokles73: Ok, based on the assumption that we do _not_ adopt the CoAP registry, defining the field in the proto file as _optional_ should be sufficient then.

basics/uattributes.adoc
```diff
@@ -185,14 +172,34 @@ The traceparent *MUST* either be empty or contain a value as defined by https://
...
+
+* *MUST* contain a value in the range `0x0000` to `0xDFFF` or `0xF000` to
+  `0xFFFF`; and
+* *MUST NOT* contain a value greater than `0xFFFF` or in the reserved range
+  `0xE000` to `0xEFFF`.
+
+Identifier `0x0000` indicates that a binding, service, or schema contract
+defines the concrete payload encoding. A component that needs to interpret such
+a payload *MUST* have that contract and *MUST NOT* infer the encoding from the
+payload data.
```
sophokles73, line 197: FMPOV we do not need to reiterate the definition of the number ranges here. The semantics of these ranges should be defined in the (separate) registry document and simply referred to here. Even a value in the _reserved_ range should actually be acceptable here: why else would we want to reserve a range, if it cannot be used?

basics/uattributes.adoc
```diff
@@ -461,55 +468,86 @@ a|The outcome of the RPC. `UCode.Ok` indicates successful invocation of the oper
...
-| `Unspecified`
-| N/A
-| The payload format is unknown. Clients will need to have gained knowledge about the concrete encoding and semantics of the payload by means of an out-of-band mechanism.
-
-
-|`ProtobufWrappedInAny`
-|`application/x-protobuf`
-a|The network byte order serialization of a `google.protobuf.Any` structure that contains the payload itself along with the schema URI indicating the (protobuf) type of the payload.
+[#payload-encoding-registry]
+== Payload-Encoding Registry
```
sophokles73, line 472: as pointed out in another comment, this should go into its own document FMPOV

basics/umessage.adoc
```diff
@@ -27,6 +27,10 @@ uProtocol entities communicate with each other by means of exchanging `UMessages
 
 A UMessage consists of xref:uattributes.adoc[UAttributes] holding the message's meta data and (optional) payload as shown in the class diagram below (using UML2 notation).
 
+Payload presence is determined by whether the optional `payload` field is
+explicitly set, not by the number of bytes it contains. An explicitly set
+zero-length byte sequence is therefore a present payload.
+
```
sophokles73, line 33: once this is merged, I will add a corresponding formal UML2 constraint ...

up-l1/cloudevents.adoc
```diff
@@ -164,7 +174,11 @@ a|A code indicating an error that has occurred during the delivery of either an
 
 The sections below define the mapping of uPayload to CloudEvent attributes.
 
-NOTE: The custom `pformat` property is used to indicate the payload type instead of the standard `datacontenttype` property defined by CloudEvents. Its value is set to the integer value assigned to the link:../up-core-api/uprotocol/uattributes.proto[UPayloadFormat]. This helps to reduce the size of the resulting data structure.
+The `pformat` property carries the numeric payload-encoding identifier.
+CloudEvents `datacontenttype` does not carry uProtocol payload-encoding
+identity. It *MAY* be set to the registered media type for a registry entry when
+that media type accurately describes the payload, but it *MUST NOT* replace
+`pformat`. Using the numeric identifier avoids requiring a media-type string.
```
sophokles73, line 181: What would be the purpose of still adding the media type? FMPOV this would simply allow for contradicting values being specified ...

up-l1/cloudevents.adoc
```diff
@@ -174,48 +188,58 @@ This mapping *MUST* be used by uProtocol Transport Libraries which use the Proto
 
 [%autowidth]
 |===
-|UPayloadFormat |CE `pformat` |CE `dataschema` |CE Property to map Payload Data to
+|Payload-encoding ID |CE `pformat` |CE `dataschema` |CE Property to map Payload Data to
```
sophokles73, line 191: as pointed out in another comment, I'd rather put the _categorization_ (`text`, `protobuf`, `binary`) into the registry document and then define a generic mapping here ...

up-l1/mqtt_5.adoc
```diff
@@ -134,13 +137,26 @@ a| A tracing identifier to use for correlating messages across the system
...
-* *MUST* be set to the link:../up-core-api/uprotocol/v1/uattributes.proto[UPayloadFormat enum] integer value's decimal string representation.
-
-Note that the enum's integer value is used instead of the _uprotocol.mime_type_ option's value in order to reduce the overall size of the PUBLISH packet.
+a| The identifier defined in the
+xref:../basics/uattributes.adoc#payload-encoding-registry[payload-encoding
+registry]
+
+* _Content Type_ *MUST* be present if and only if the PUBLISH packet represents
+  a present UMessage `payload` field. This includes a present zero-length
+  payload.
```
sophokles73, line 148:
```suggestion
* _Content Type_ *MUST* be present if and only if the PUBLISH packet represents
  a UMessage with a present `payload` field. This includes a present zero-length
  payload.
```

up-l1/mqtt_5.adoc
```diff
@@ -134,13 +137,26 @@ a| A tracing identifier to use for correlating messages across the system
...
+* When present, _Content Type_ *MUST* be set to the identifier's decimal string
+  representation, including `0` for an implicit encoding.
```
sophokles73, line 150: IMHO we do not need to explicitly mention the `0` value, because it is simply one specific encoding ID that is defined in the registry. The semantics of this value should be defined there ...

up-l1/mqtt_5.adoc
```diff
@@ -134,13 +137,26 @@ a| A tracing identifier to use for correlating messages across the system
...
+* When reconstructing a UMessage, the presence of _Content Type_ determines the
+  presence of its `payload` and `_payload_encoding_id_` fields. A receiver
+  *MUST* reject a _Content Type_ value that is not a decimal unsigned integer,
+  is greater than `65535`, or is in the reserved range `57344` to `61439`
+  (`0xE000` to `0xEFFF`).
```
sophokles73, line 155: see my comment above regarding the validity of the [0xE000; 0xEFFF] range

sophokles73 (changes requested):

PR #337, opened 2026-08-26
Use camel case names in data models

sophokles73:
The UML data models use camel case names for attributes and messages. This commit updates the data models defined throughout the specification to follow this naming convention.

This is to further distinguish the logical data model from the corresponding protobuf message definitions, which merely define a wire format for the data model that implementors must support.


AnotherDaniel (approved):
Thanks for picking up usub v4 as well with this.

PR #338, opened 2026-08-26
[#328] Make TTL and Priority mandatory for UAttributes

### LINKED ISSUE #328: Make some UMessage attributes mandatory

sophokles73:
#328

### DIFF
--- a/basics/uattributes.adoc
+++ b/basics/uattributes.adoc
@@ -41,7 +41,7 @@ class UAttributes {
   id: UUID
   source: UUri
   sink: UUri [0..1]
-  ttl : UInt32 [0..1]
+  ttl : UInt32 = 0
   commstatus: UCode [0..1]
   permissionLevel: UInt32 [0..1]
   requestId: UUID [0..1]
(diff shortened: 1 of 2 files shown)


AnotherDaniel (approved):
LGTM

PR #339, opened 2026-08-26
Require message UUris to not contain wildcards

sophokles73:
The up-rust crate already verifies that the source and sink UUris of a message do not contain any wildcards, but the corresponding requirement was not contained in the UProtocol specification. This commit adds the requirement to the specification.


AnotherDaniel (approved):
LGTM

PR #345, opened 2026-09-18
Fix requirement ID name typo

AnotherDaniel:
Fix duplicate requirement ID (originally caused by copy-paste inattention)


sophokles73 (approved):
LGTM

